PR #38, opened 2023-02-21
feat(ocm): Install and configure the ocm plugin

### LINKED ISSUE #7: Add OCM Plugin
Story
- As a user, I wish to be able to view information about the clusters that I have available to me using the Open Cluster Management plugin within Backstage

Description
- The Open Cluster Management plugin allows users the ability to see information about their available clusters. Using OCM, we can view information about clusters as well as access them straight from the Backstage app.

AC
- OCM plugin added to the Backstage Showcase app
- View cluster related information within the Backstage Showcase app
- Document the OCM specific process for deployment

Additional Notes
- Will require the Kubernetes plugin to be implemented within the Backstage Showcase app

SamoKopecky:
## What does this PR do / why we need it

Install the OCM plugin and all of the fronted components. Discovered clusters are converted to catalog entities with kind `resource`.

Fixes #7

## PR acceptance criteria

- [ ] Unit Tests
- [ ] E2E Tests
- [ ] Documentation

The secret `moc_infra_token` is located in vault in `moc/smaug/service-catalog/k8s-plugin-tokens`. The added external secret will create a secret in the `janus-idp` namespace and will get consumed.

### DIFF
--- a/app-config.yaml
+++ b/app-config.yaml
@@ -115,3 +115,9 @@ kubernetes:
           authProvider: 'serviceAccount'
           skipTLSVerify: true
           serviceAccountToken: ${K8S_CLUSTER_TOKEN}
+
+ocm:
+  hub:
+    name: ${OCM_HUB_NAME}
+    url: ${OCM_HUB_URL}
+    serviceAccountToken: ${moc_infra_token}
--- a/packages/backend/src/plugins/catalog.ts
+++ b/packages/backend/src/plugins/catalog.ts
@@ -1,5 +1,6 @@
 import { CatalogBuilder } from '@backstage/plugin-catalog-backend';
 import { ScaffolderEntitiesProcessor } from '@backstage/plugin-scaffolder-backend';
+import { ManagedClusterProvider } from '@janus-idp/backstage-plugin-ocm-backend';
 import { Router } from 'express';
 import { PluginEnvironment } from '../types';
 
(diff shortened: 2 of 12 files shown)


app-config.yaml
```diff
@@ -115,3 +115,9 @@ kubernetes:
           authProvider: 'serviceAccount'
           skipTLSVerify: true
           serviceAccountToken: ${K8S_CLUSTER_TOKEN}
+
+ocm:
+  hub:
+    name: infra
+    url: https://api.moc-infra.massopen.cloud:6443
```
christophe-f, line 122: Should make an environment variable

packages/backend/src/plugins/catalog.ts
```diff
@@ -1,14 +1,27 @@
...
   builder.addProcessor(new ScaffolderEntitiesProcessor());
   const { processingEngine, router } = await builder.build();
   await processingEngine.start();
+  await env.scheduler.scheduleTask({
+    id: 'run_ocm_refresh',
+    fn: async () => {
+      await ocm.run();
+    },
+    frequency: { minutes: 30 },
+    timeout: { minutes: 10 },
```
christophe-f, line 24: Can we add this in the app-config.yaml?

christophe-f (approved):

packages/backend/src/plugins/catalog.ts
SamoKopecky: I don't think we can put this into the `app-config.yaml` at this time, this would require some additional changes to the source code. However we can talk about this more and maybe implement it in the future.

christophe-f: no problem. I just wanted to keep track here.
Will merge later today. Need the environment up for a recording.

PR #120, opened 2023-03-15
added support for github org data provider

raffaelespazzoli:
## Description

added support for github org data provider.

The backstage example had some information hardcoded. @christophe-f can you help me parameterize them?

## Which issue(s) does this PR fix

- Fixes #115 

## PR acceptance criteria

_Put an `x` in the boxes that apply_

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary

PR #121, opened 2023-03-15
enabled default github identity resolver

### LINKED ISSUE #116: Enable github identity resolver
### What do you want to improve?

Enable github identity resolver, this will allow a user that logs in to be mapped to an identity imported via the github org processor (see #115 ). This is a necessary step to understand the ownership of a user that is logged it. Having an identity resolver is also the base of any serious RBAC implementation.

see here: https://backstage.io/docs/auth/identity-resolver/#built-in-resolvers
also an implementation with the default github idneityt resolver: https://github.com/raf-backstage-demo/backstage-app-demo/blob/main/packages/backend/src/plugins/auth.ts

### :warning: Is this a breaking change?

- [ ] Yes
- [ ] No
- [x] Don't know

raffaelespazzoli:
## Description

enabled default github identity resolver. This change worked for me. But it may need some conditional logic to coexist with other identity resolved. Basically one can assume that given how a user was authenticated, there is an identity resolved that applies better than others. And then perhaps have a default one (the one I commented out).
Not sure how to write that conditional logic. This is better tested only after #120 is merged and it needs that data to do the mapping.

- Fixes #116 

## PR acceptance criteria

- [ ] Documentation is updated if necessary

### DIFF
--- a/packages/backend/src/plugins/auth.ts
+++ b/packages/backend/src/plugins/auth.ts
@@ -37,16 +37,16 @@ export default async function createPlugin(
       //   https://backstage.io/docs/auth/identity-resolver
       github: providers.github.create({
         signIn: {
-          resolver(_, ctx) {
-            const userRef = 'user:default/guest'; // Must be a full entity reference
-            return ctx.issueToken({
-              claims: {
-                sub: userRef, // The user's own identity
-                ent: [userRef], // A list of identities that the user claims ownership through
-              },
-            });
-          },
-          // resolver: providers.github.resolvers.usernameMatchingUserEntityName(),
+          // resolver(_, ctx) {
+          //   const userRef = 'user:default/guest'; // Must be a full entity reference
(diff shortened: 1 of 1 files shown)


packages/backend/src/plugins/auth.ts
```diff
@@ -37,16 +37,16 @@ export default async function createPlugin(
...
-          // resolver: providers.github.resolvers.usernameMatchingUserEntityName(),
+          // resolver(_, ctx) {
+          //   const userRef = 'user:default/guest'; // Must be a full entity reference
+          //   return ctx.issueToken({
+          //     claims: {
+          //       sub: userRef, // The user's own identity
+          //       ent: [userRef], // A list of identities that the user claims ownership through
+          //     },
+          //   });
+          // },
```
christophe-f, line 48: Can delete the unused section

christophe-f:
LGTM

Issue #131, opened 2023-03-17
Use actions/cache in the GitHub Actions


christophe-f:
### What do you want to improve?

Reduce the GitHub action execution time

### What is the current behavior?

We are not using caching

### What is the new behavior?

Use caching to improve execution time on most of the steps in the different GitHub Actions with actions/cache save and restore

### :warning: Is this a breaking change?

_Put an `x` in the boxes that apply_

- [ ] Yes
- [x] No
- [ ] Don't know

christophe-f:
Reopening. The cache is enable in the CI.yaml and reduced the time by 2 minutes.

But it's not enabled on Docker yet. https://docs.docker.com/build/cache/backends/gha/

PR #140, opened 2023-03-21
Add caching to docker-build workflow and update deps

schultzp2020:
## Description

- Adds cache to docker-build workflow
- Updates deps

- Fixes https://github.com/janus-idp/backstage-showcase/issues/131

## PR acceptance criteria

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -45,6 +45,12 @@ jobs:
       packages: write
 
     steps:
+      - name: Checkout
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
 
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -37,19 +37,16 @@ env:
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
-  push-docker-build:
-    name: Push Docker Build Job
+  docker-build:
+    name: Docker Build Job
     runs-on: ubuntu-latest
-    if: github.event_name == 'push'
     permissions:
       contents: read
       packages: write
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -51,8 +51,5 @@
       "prettier --write"
     ]
   },
-  "packageManager": "yarn@3.4.1",
-  "dependencies": {
-    "@backstage/plugin-sonarqube": "^0.6.5"
```
christophe-f, line 56: The SonarQube plugin won't work if it's not installed.
I will check if it can be moved to the frontend dependencies but from the documentation, it's in this package.json file.

.github/workflows/docker-build.yaml
```diff
@@ -73,21 +85,17 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
-          push: true
+          push: ${{ github.event_name == 'push' }}
           provenance: false
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
```
christophe-f, line 93: If we are using type=local how is it going to be reused for the next build? type=gha is probably a better option

.github/workflows/docker-build.yaml
```diff
@@ -73,21 +85,17 @@ jobs:
...
-
-      - name: Docker
-        run: docker build .
+      - # Temp fix
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
+        name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```
christophe-f, line 101: I don't think this is needed if we are using the gha (github action) cache.

christophe-f (changes requested):

package.json
schultzp2020: The SonarQube plugin does not need to be installed in the root `package.json`. It is already installed under the `packages/app` `package.json`. For future reference, the only packages that need to be installed on the root level are workspace wide dependencies like `turbo`.

.github/workflows/docker-build.yaml
schultzp2020: We only want to push the image on a `push` event type; therefore, `type=gha` cannot be used here because it requires an external container registry to push to.

.github/workflows/docker-build.yaml
schultzp2020: Look at the above comment.

christophe-f (approved):

PR #157, opened 2023-03-28
feat: Add postgres database to deployment

### LINKED ISSUE #141: Add a Postgresql database
### Acceptance criteria

The showcase app is currently using the default in memory database. We should install a PG database and use it.

SamoKopecky:
## Description

Add a deployment for a postgres database which includes:
- a postgres `component` for adding the operator
- a `PostgresCluster` which is the actual database instance
- the database will automatically backup to a storage volume every week with a full backup and every 6 days with a differential backup

- Fixes #141

## PR acceptance criteria

- [ ] Documentation is updated if necessary

### DIFF
--- a/app-config.production.yaml
+++ b/app-config.production.yaml
@@ -11,25 +11,19 @@ backend:
   baseUrl: http://localhost:7007
   # The listener can also be expressed as a single <host>:<port> string. In this case we bind to
   # all interfaces, the most permissive setting. The right value depends on your specific deployment.
-  listen: ':7007'
+  listen: ":7007"
 
   # config options: https://node-postgres.com/api/client
   database:
     client: pg
     connection:
-      host: ${POSTGRES_HOST}
-      port: ${POSTGRES_PORT}
-      user: ${POSTGRES_USER}
-      password: ${POSTGRES_PASSWORD}
+      host: ${host}
+      port: ${port}
+      user: ${user}
+      password: ${password}
       # https://node-postgres.com/features/ssl
       # you can set the sslmode configuration option via the `PGSSLMODE` environment variable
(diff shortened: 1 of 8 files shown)


app-config.production.yaml
```diff
@@ -11,25 +11,19 @@ backend:
...
     client: pg
     connection:
-      host: ${POSTGRES_HOST}
-      port: ${POSTGRES_PORT}
-      user: ${POSTGRES_USER}
-      password: ${POSTGRES_PASSWORD}
+      host: ${host}
+      port: ${port}
+      user: ${user}
+      password: ${password}
```
christophe-f, line 23: Any reasons to rename these variables?

SamoKopecky: Yes the secret `janus-idp-pguser-janus-idp` that is created by the `PostgresCluster` contains values using these keys:
```yaml
kind: Secret
apiVersion: v1
metadata:
  name: janus-idp-pguser-janus-idp
  namespace: janus-idp
...
data:
  host: 
  password: 
  port: 
  user: 
  verifier:
type: Opaque

```

christophe-f: ok, so we can rename the secret instead. We will have to add this in the https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md and having only `host, port`... are too generic names.

SamoKopecky: As far as I know we can't rename the keys in the secret since its auto-generated by the postgres cluster deployment. However if you know of a way to do this I all open since I also think the keys are too generic.

I'll update the documentation accordingly thought.

christophe-f (approved):
LGTM

PR #158, opened 2023-03-28
Read external data from s3

### LINKED ISSUE #79: Create 1 OBC in janus-idp
- [ ] Create a folder `homepage` in the bucket and move the data for the homepage
- [ ] Create a folder `techradar` in the bucket and move the data for the techradar page

We could also create a GPT for this feature. https://github.com/janus-idp/software-templates/issues/88

PatAKnight:
## Description

Adds the AWS SDK to allow us to read in external data into the showcase app. The homepage and tech radar both have files store in an object bucket claim that we can use AWS to read from. There is also added logic that will default to the files stored in the source code so that the homepage and tech radar will still load data even without access to our OBC.

- Fixes #79 

## PR acceptance criteria


PatAKnight:
There was a recent update to the homepage data yesterday. The slack channel url was updated. The new changes will need to be copied over to the OBC whenever this gets merged so that it will read in the most recent updates.

Here is the [info](https://www.operate-first.cloud/apps/content/odf/docs/buckets-external-access.html) on accessing the bucket from Operate First

PatAKnight:
Closing this in favor for #312 as there were a lot of changes that happened to Backstage since this PR has went up.

PR #185, opened 2023-03-30
Showcase data

### LINKED ISSUE #178: The data is not loading by default in showcase
### What happened?

We had a big change with more and better data. But it's not loading in the deployed version.

We could re add this in the app-config.yaml:
```
- type: file
      target: ./catalog-entities/all.yaml
```

But the best solution would be to use the url type:
```
- type: url
      target: https://github.com/janus-idp/backstage-showcase/blob/main/catalog-entities/all.yaml
```

And clean up the Dockerfile as we won't need the line:
`COPY ./catalog-entities ./catalog-entities`

PatAKnight:
## Description

Adds a URL pointing to the `all.yaml` within our catalog entities directory for the catalog. This will fix the of the data not showing up within the catalog. 

The issue is happening because we currently have to account for whether the Showcase app is running in a container or locally. This requires us to declare the location twice whenever we point to a local directory. The first declaration is if the app is running locally `./catalog-entities/` and the second is whenever it is running in a container `../../catalog-entities/`

- Fixes #178

## PR acceptance criteria

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,7 +93,6 @@ RUN tar xzf bundle.tar.gz && rm bundle.tar.gz
 # Copy any other files that we need at runtime
 COPY ./app-config.yaml .
 COPY ./github-app-backstage-showcase-credentials.yaml .
-COPY ./catalog-entities ./catalog-entities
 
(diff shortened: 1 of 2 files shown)


christophe-f (approved):
LGTM

PR #234, opened 2023-05-03
Add Tekton plugin

debsmita1:
## Description

- Adds the Tekton plugin to the Showcase App. The CI tab of the App will display the latest pipeline run visualization if the tekton CI annotation (`janus-idp.io/tekton-enabled: 'true'`) is available.
- Updated the topology package version

## Which issue(s) does this PR fix

- Fixes https://github.com/janus-idp/backstage-showcase/issues/172

## PR acceptance criteria

Please make sure that the following steps are complete:

- GitHub Actions are completed and successful
- Unit Tests are updated and passing
- E2E Tests are updated and passing
- Documentation is updated if necessary
- Add a screenshot if the change is UX/UI related

## How to test changes / Special notes to the reviewer

- Add ${K8S_CLUSTER_NAME} , ${K8S_CLUSTER_URL} ,  ${K8S_CLUSTER_TOKEN} info
- Create a pipeline run in the above cluster and ensure that the resource has the "backstage.io/kubernetes-id= test-backstage" label

## Screenshots
Screenshot 2023-05-03 at 3 40 56 PM

Screenshot 2023-05-04 at 10 14 08 PM

PR #245, opened 2023-05-08
add the annotation to enable tekton in the CI tab of Backstage Showcase entity

debsmita1:
## Description

The [PR](https://github.com/janus-idp/backstage-showcase/pull/234) adds the tekton plugin to the showcase app, but the pipelinerun visualization will appear on the CI tab only if the Component entity has the annotation `janus-idp.io/tekton-enabled: 'true'`. Without this change, the users using the showcase app will have to add another catalog entity with the above-mentioned annotation to view the latest pipelinerun visualization in the CI tab of that entity.

## Screenshot
Component entity has `GithubActions` and `Tekton` CIs enabled

Screenshot 2023-05-08 at 11 33 33 PM

## PR acceptance criteria

### DIFF
--- a/catalog-entities/components/showcase.yaml
+++ b/catalog-entities/components/showcase.yaml
@@ -19,6 +19,7 @@ metadata:
     backstage.io/kubernetes-id: "janus-idp"
     github.com/project-slug: janus-idp/backstage-showcase
     quay.io/repository-slug: janus-idp/backstage-showcase
+    janus-idp.io/tekton-enabled: 'true'
     backstage.io/techdocs-ref: url:https://github.com/janus-idp/backstage-showcase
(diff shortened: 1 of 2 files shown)


schultzp2020:
@debsmita1 can you rebase?

schultzp2020:
@debsmita1 can you also add a changeset? The changesets pr was just merged.

christophe-f:
What's the goal of this PR? Users will probably uses only GH Actions or Tekton. I don't think we need to have flags for both. Otherwise we will have to handle or the different combos (Jenkins, Azure Devops, GitLab CI, etc).

Also the showcase is not using Tekton pipeline yet so having the annotation set to true will never show any data (and probably confuse users for now). Once we move to the new environment, we should use Tekton and enable the tab.

PR #248, opened 2023-05-09
Add post merge and post checkout hooks

schultzp2020:
## Description

Add post merge hook to run `yarn install` when any `package.json` is out of date.

N/A

## PR acceptance criteria

### DIFF
--- a/.husky/scripts/sync-node-modules
+++ b/.husky/scripts/sync-node-modules
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+changed_files="$(git diff-tree -r --name-only --no-commit-id HEAD@{1} HEAD)"
+
+if [[ $changed_files == *"package.json"* ]]; then
+  eval "yarn install"
+fi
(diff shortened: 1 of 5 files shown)


.husky/scripts/sync-node-modules
```diff
@@ -0,0 +1,5 @@
+changed_files="$(git diff-tree -r --name-only --no-commit-id HEAD@{1} HEAD)"
+
+if [[ $changed_files == *"package.json"* ]]; then
+  eval "yarn install"
```
RobotSail, line 4: Why do we need this?

RobotSail (commented):
While I can see how running `yarn install` post-merge & post-checkout would make sense from a convenience standpoint, I personally prefer to handle this myself for a few reasons such as:
- Moving between branches temporarily
- Wanting to clone something to make a quick change, not wanting to install all of the packages

Is there a particular reason that we need these? I'm not trying to block progress, I can just see this becoming an annoyance for me personally.

.husky/scripts/sync-node-modules
RobotSail: Actually, this is probably fine for the backstage showcase.

RobotSail (dismissed):
Since this is the showcase repository and not the main dev environment, I think it makes sense to add the `yarn install` hook on every checkout.

RobotSail (approved):

PR #264, opened 2023-05-19
feat: modify app-configs so default entrypoint can be used by helm

tumido:
## Description

Chasing https://github.com/janus-idp/book/issues/14 we want to be able to run this image "as is" which should enable users to replicate a default Showcase setup. That means: most custom features are disabled because of missing credentials, but the pod starts and runs fine.

Therefore I'm proposing:

- rename `app-config.basic.yaml` to `app-config.example.yaml`
- add `app-config.example.production.yaml` which enables postgres and routes frontend via backend (similar to `app-config.production.yaml`, but without certificates and using different variable names that are available through the upstream helm chart)
- change `CMD` to `ENTRYPOINT` so it doesn't get overwritten by dynamic `args` in the pod spec of the Helm chart
- change the entrypoint arguments to list all configs that should be enabld when running the example application
- update `manifests/base/deployment.yaml` so Janus Showcase doesn't propagate any of these changes into its configuration.

Part of https://github.com/janus-idp/book/issues/14

## PR acceptance criteria

### DIFF
--- a/showcase-docs/getting-started.md
+++ b/showcase-docs/getting-started.md
@@ -6,7 +6,7 @@ There are several different methods for running the Backstage Showcase app today
 
 The easiest and fastest method for getting started: Backstage Showcase app, running it locally only requires a few simple steps.
 
-1. Copy `app-config.basic.yaml` and rename it as `app-config.local.yaml`.
+1. Copy `app-config.example.yaml` and rename it as `app-config.local.yaml`.
 
 ## Running Locally with the Optional Plugins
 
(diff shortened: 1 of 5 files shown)


schultzp2020 (approved):
LGTM

showcase-docs/getting-started.md
```diff
@@ -6,7 +6,7 @@ There are several different methods for running the Backstage Showcase app today
 
 The easiest and fastest method for getting started: Backstage Showcase app, running it locally only requires a few simple steps.
 
-1. Copy `app-config.basic.yaml` and rename it as `app-config.local.yaml`.
+1. Copy `app-config.example.yaml` and rename it as `app-config.local.yaml`.
```
christophe-f, line 9: why not using `app-config.local.yaml` and avoid the rename step?

tumido: I think `example` is fine...  `local` is by default `.gitignored` so users can change it to their liking if they are developing locally... This way we separate the example deployment (the source for example config) and the autoloaded local defaults. I would keep it as is

PR #312, opened 2023-06-15
feat: Read external data from s3

PatAKnight:
## Description

Adds the AWS SDK to allow us to read in external data into the showcase app. The homepage and tech radar both have files store in an object bucket claim that we can use AWS to read from. There is also added logic that will default to the files stored in the source code so that the homepage and tech radar will still load data even without access to our OBC.

## Which issue(s) does this PR fix

- Fixes #79 

## PR acceptance criteria

Please make sure that the following steps are complete:

- GitHub Actions are completed and successful
- Unit Tests are updated and passing
- E2E Tests are updated and passing
- Documentation is updated if necessary
- Add a screenshot if the change is UX/UI related

PR #351, opened 2023-06-23
feat(app): fetch customization data from remote

invincibleJai:
## Description

1. data powering the home page for quick access link is in code itself (https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/public/homepage/data.json) which ideally should be configurable.
2. data powering the techRadar plugin is in code itself (https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/public/tech-radar/data-default.json) which ideally should be configurable.

As part of this PR made the data for the home page configurable so that can be passed in `app-config.yaml `as a proxy. API url will return the data in JSON in the expected format/signature and fallback to the local data if api call fails or not configured.

For Home page(Quick Access links, the base URL at target should serve and for tech radar then svc URL should be `/tech-radar`

- Add below to app-config.yaml

```
proxy:
  # customize backstage(janus) instance
  '/developer-hub':
    target: ${HOMEPAGE_DATA_URL}
    headers:
       <header_key>: <header_value>
    changeOrigin: true
    secure: true
```

- Make sure the API URL passed in the target above returns the JSON in response in the expected format as here https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/public/homepage/data.json

- Fixes: https://github.com/janus-idp/backstage-showcase/issues/356

## PR acceptance criteria

- Add below to app-config.yaml

```
proxy:
  # customize backstage(janus) instance
  '/developer-hub':
    target: ${HOMEPAGE_DATA_URL}
    headers:
       <header_key>: <header_value>
    changeOrigin: true
    secure: true
```

- Make sure the API URL passed in the target above returns the JSON in response in the expected format as here https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/public/homepage/data.json

- If you want to test with actual service then can use https://github.com/invincibleJai/test-showcase-app

### DIFF
--- a/packages/app/src/lib/CustomTechRadar.ts
+++ b/packages/app/src/lib/CustomTechRadar.ts
@@ -2,12 +2,23 @@ import {
   TechRadarApi,
   type TechRadarLoaderResponse,
 } from '@backstage/plugin-tech-radar';
+import { JanusBackstageCustomizeApi } from '../api';
 
 export class CustomTechRadar implements TechRadarApi {
+  private readonly janusBackstageCustomizeApi: JanusBackstageCustomizeApi;
+  constructor(options: {
+    janusBackstageCustomizeApi: JanusBackstageCustomizeApi;
+  }) {
+    this.janusBackstageCustomizeApi = options.janusBackstageCustomizeApi;
+  }
   async load(id: string | undefined): Promise<TechRadarLoaderResponse> {
-    const data = (await fetch(`/tech-radar/data-${id}.json`).then(res =>
-      res.json(),
-    )) as TechRadarLoaderResponse;
+    let data;
+    try {
+      data = await this.janusBackstageCustomizeApi.getTechRadarDataJson();
+    } catch (e) {
+      const res = await fetch(`/tech-radar/data-${id}.json`);
+      data = (await res.json()) as TechRadarLoaderResponse;
+    }
 
     return {
       ...data,
--- a/packages/app/config.d.ts
+++ b/packages/app/config.d.ts
@@ -1,0 +1,10 @@
+export interface Config {
+  /** Configurations for the backstage(janus) instance */
+  developerHub?: {
+    /**
+     * The url of json data for customization.
+     * @visibility frontend
+     */
+    proxyPath?: string;
+  };
+}
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -98,6 +98,8 @@
     ]
   },
   "files": [
-    "dist"
-  ]
+    "dist",
+    "config.d.ts"
+  ],
+  "configSchema": "config.d.ts"
 }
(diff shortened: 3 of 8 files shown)


invincibleJai:
This PR is in the draft state as more of validation/spike and the final approach needs to be discussed

cc @christophe-f @serenamarie125

packages/app/src/lib/CustomTechRadar.ts
```diff
@@ -2,12 +2,24 @@ import {
...
+  }
   async load(id: string | undefined): Promise<TechRadarLoaderResponse> {
-    const data = (await fetch(`/tech-radar/data-${id}.json`).then(res =>
-      res.json(),
-    )) as TechRadarLoaderResponse;
+    let data;
+    try {
+      data = await this.janusBackstageCustomizeApi.getTechRadarDataJson();
+    } catch (e) {
+      data = (await fetch(`/tech-radar/data-${id}.json`).then(res =>
```
gorkem, line 19: Could this be just done with awaits only?

gorkem (approved):
I added one comment. Not a blocker but It would be good if we can address it.

packages/app/src/lib/CustomTechRadar.ts
invincibleJai: we can do that but then needs to await i.e 

```
const res = await fetch(`/tech-radar/data-${id}.json`) // as this return promise
data = await res.json();
```

LMK, your thoughts

schultzp2020 (changes requested):
Can you add a mechanism to have a fallback URL specified? This [PR](https://github.com/janus-idp/backstage-showcase/pull/312/files#diff-40cc91b81c9b62ce97c0e953ae93e508a07b769f5f2e28049e15aba0f6855681R4) was going to address it.

PatAKnight:
Do we also want to include the [learning path cards](https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/components/learningPaths/LearningPathsPage.tsx#L36) to this as well?

invincibleJai:
> Can you add a mechanism to have a fallback URL specified? This [PR](https://github.com/janus-idp/backstage-showcase/pull/312/files#diff-40cc91b81c9b62ce97c0e953ae93e508a07b769f5f2e28049e15aba0f6855681R4) was going to address it.

Hi @schultzp2020, by fallback you mean looking for local data? if yes then it's handled 

https://github.com/janus-idp/backstage-showcase/pull/351/files#diff-122aceaecd71102bb8be9c35a7287ded247fa6a176acac53c04287d789442dc5R21

https://github.com/janus-idp/backstage-showcase/pull/351/files#diff-2b67190e700def57d923e34d12138cc317d40dcc314c76a59334f9b8b65cbe6fR19

schultzp2020 (changes requested):

schultzp2020 (approved):
LGTM

PR #375, opened 2023-07-13
add documentation

varodrig:
## Description

Adding documentation to help developers containerized their apps.

- Fixes #? No Fixes.
- Adding new documentation

## PR acceptance criteria


tumido (commented):
Hi @varodrig, can you please elaborate on why would we want this guide in here? It doesn't seem to target Backstage in any means and is very removed from the reality of this repository. It seems more like a copy-pasted generic guide than documentation tailored to the specific needs of this application.

Points like:

> 1.1 Can this application be deployed in containers? Should it be in a restricted environment? Cloud? On-Prem?

> - Front-end and backend should exist in a separate repository to take advantage of the [container technology](https://kubernetes.io/docs/concepts/containers/).

> - Multiple apps sharing the same code violate the container factor.

> An app's _config_ is everything likely to vary between [deploys](./build-release-run) (staging, production, developer environments). Configurations are independently managed for each deployment. This model scales up smoothly as the app expands into more deploys over its lifetime.

I think this is totally omitting the decisions already made by Janus or Backstage upstream community. Developers using this repository are not meant to make those decisions themselves again. The upstream community already has well-defined standards about how the configuration looks like, what is contained in an instance repository etc.

Was this guide written with Backstage in mind or was it not considered at all?

varodrig:
Hi Tom,

Thanks for reaching out.

The documentation was reviewed with @Christophe Fargette
***@***.***>, and he thought it could be a good idea to include
it here.

The use case of this documentation is for the migration path developers
need to do to ensure the application is ready to move into containers
before deploying it into OpenShift.

Not all applications are ready to run on a container platform. Ensuring the
application is designed to run in containers and guiding developers will
help developers move faster.

The documentation is also related to the existing ones on migration
paths/processes. It's all part of the same use case, giving developers a
guide to modernizing applications.

We use this approach with customers, and current consulting teams also use
it.

In the context of Janus, it doesn't mean we'll not abstract some of the
complexity to them. This meant to be a guideline to provide developers
understanding of the source code modernization process, based on the
TWELVE-FACTOR-APP.


I hope this helps.

PR #393, opened 2023-07-20
Feat(app): Add default permission policy

### LINKED ISSUE #259: Add Permission framework
### Acceptance criteria

What are you trying to do, need? What are the acceptance criteria?

We need to use the [Permission framework](https://backstage.io/docs/permissions/overview) included in Backstage.
This issue is to investigate what can be done with this feature and what are the limitations.

Here are some example scenario:
- How a user can view only their components from the software catalog?
- How only admin can access plugins? ie In the announcement plugin or branding plugin, only admins should be able to update the content
- How to use that data from the Keycloak plugin? Do we have everything needed for authorization?

PatAKnight:
## Description
Adds a basic default permission policy to Showcase. It adds the permission plugin and enables it through app-config and [Service to Service Auth](https://backstage.io/docs/auth/service-to-service-auth#setup). It also adds an allow all permission policy to the showcase app.

This is the initial step in our larger RBAC work. 

- Fixes #259 

## PR acceptance criteria


app-config.yaml
```diff
@@ -275,6 +275,10 @@ pagerduty:
 lighthouse:
   baseUrl: ${LIGHTHOUSE_BASEURL}
 
+# Plugin: Permission
+permission:
+  enabled: true
```
schultzp2020, line 280:
```suggestion
enabled:
  permissions: ${PERMISSIONS_ENABLED}
```

Please keep this consistent with other plugins inside the `enabled` block.

packages/backend/src/index.ts
```diff
@@ -202,6 +203,13 @@ async function main() {
     router: jenkins,
     isOptional: true,
   });
+  await addPlugin({
+    plugin: 'permission',
+    config,
+    apiRouter,
+    createEnv,
+    router: permission,
+  });
```
schultzp2020, line 212:
```suggestion
  await addPlugin({
    plugin: 'permission',
    config,
    apiRouter,
    createEnv,
    router: permission,
    isOptional: true,
  });
```
The plugin should have `isOptional: true`.

app-config.yaml
```diff
@@ -15,9 +15,9 @@ backend:
   # Used for enabling authentication, secret is shared by all backend plugins
   # See https://backstage.io/docs/auth/service-to-service-auth for
   # information on the format
-  # auth:
-  #   keys:
-  #     - secret: ${BACKEND_SECRET}
+  auth:
+    keys:
+      - secret: ${BACKEND_SECRET}
```
schultzp2020, line 20:
```suggestion
      - secret: ${BACKEND_AUTH_SECRET}
```

The env var should be more specific.

.changeset/happy-cups-fry.md
```diff
@@ -0,0 +1,18 @@
+---
+'backend': patch
```
schultzp2020, line 2:
```suggestion
'backend': minor
```

This is adding a new nonbreaking feature.

packages/backend/src/index.ts
```diff
@@ -202,6 +203,13 @@ async function main() {
     router: jenkins,
     isOptional: true,
   });
+  await addPlugin({
+    plugin: 'permission',
```
schultzp2020, line 207:
```suggestion
    plugin: 'permissions',
```

The name needs to be updated to correlate with `enabled.permissions`.

packages/backend/src/index.ts
```diff
@@ -37,6 +37,7 @@ import scaffolder from './plugins/scaffolder';
 import search from './plugins/search';
 import sonarqube from './plugins/sonarqube';
 import techdocs from './plugins/techdocs';
+import permission from './plugins/permission';
```
schultzp2020, line 40: Should the filename and import be updated to `permissions` since it will have more than one permission?

schultzp2020 (changes requested):
Don't forget to update the changeset description.

app-config.yaml
PatAKnight: It is set like this as it is required by the [Permission Plugin](https://backstage.io/docs/permissions/getting-started#2-enable-and-test-the-permissions-system) within Backstage. Now, do we still want to include the change below? 
```yaml
enabled:
  permissions: ${PERMISSIONS_ENABLED}
```
I am worried that it could cause some slight confusion as they are very similiar and users might think setting one is enough. It also increases the number of environment variables that need to be managed.

packages/backend/src/index.ts
PatAKnight: Right now, this adds a single default permission policy to the showcase app. So I think `permission` should suffice. Whenever the RBAC plugin is introduced, that will handle the creation on several permission policies and I believe that it might warrent a change then.

app-config.yaml
PatAKnight: Fixed

.changeset/happy-cups-fry.md
PatAKnight: Fixed

app-config.yaml
christophe-f: Yes some plugins and Backstage will have the enabled attribute in a specific section. It also more traditional. We will have to change our existing config.

packages/backend/src/index.ts
PatAKnight: Looks like renaming it to `permissions` might not be possible due to [line 55](https://github.com/janus-idp/backstage-showcase/blob/11ca8dd941c149311f4815380b10ce0d11c075e2/packages/backend/src/index.ts#L55) and [line 76](https://github.com/janus-idp/backstage-showcase/blob/11ca8dd941c149311f4815380b10ce0d11c075e2/packages/backend/src/index.ts#L76). We could do something like `permissionPolicy` instead

PR #405, opened 2023-07-24
fix(catalog): ci tab content providers order

### LINKED ISSUE #404: The CI tab doesn't show Tekton if GitLab is enabled.
### What happened?
The CI tab only shows `GitLab Merge Request Status` instead of the Tekton last run.  There should be a way to prioritize Tekton over GitLab if using Tekton

### What are the steps to reproduce this issue?

1. enable gitlab
2. enable tekton
3. create a component with a tekton pipeline
4. view CI tab and see message above

### Additional information

- Which devices are affected? All
- Which browsers are affected? All
- Which version are you using? RHDH 1.0

We only support the last 2 major versions.

Attach any related screenshots, if any.
image

tumido:
## Description

GitLab/GitHub doesn't provide any meaningful detection to distinguish between GitLab usage as VCS and Gitlab usage as CI, we need to find some workaround

Additionally, this PR fixes wrong GitLab CI content from: `EntityGitlabMergeRequestsTable` to `EntityGitlabPipelinesTable`.

@christophe-f @sonyccd please advise, do we want to:

1. Show data from all providers if they are applicable based on the entity annotations
   - This has a 👎 in potentially showing empty cards for GitHub/GitLab if used for source code (as VCS) but not as CI
   - This has a 👍 since we capture all the data. It is fairly common that users use 1 CI provider for linting, enforcing rules etc. (like GitHub Actions) and another CI for integration tests etc.
2. Reorder CI providers in a way that we would default to GitHub/GitLab only if no other specialized solution applies. This will always show a single CI tile.

We can do only one for the above. I think we should do 1, but I would appreciate your input as well.

- Fixes #404 

## PR acceptance criteria


jayfray12:
If you go with option one which I'm not opposed to is there a way to prioritize the order in which the different CIs are shown?  Like if I'm using Tekton and GitHub Actions like you mentioned could I show Tekton first then GitHub or vice versa?

christophe-f:
For the PR, we should definitely merge it as it fixes a bug.

For the logic, we can move the discussion in the issue. Entreprises won't use 2 CI in the pipelines. Only 1 tool. They could use Jenkins on older/existing component, and replacing newer apps with GitLab CI or Tekton, but a component will not use 2 CI tools.

In the case we detect 2 annotations for GitLab and Tekton, I would imagine that GitLab will be used for the code only and not CI. And Tekton would be the CI. 
We won't get that right at the first time, but we can get better at every iteration.

christophe-f (approved):
LGTM

PR #430, opened 2023-08-14
chore: fix documentation for GitHub Discovery and Org data

kadel:
## Description
The quickstart guide was written in a way that looked like GitHub Org Entity, and GitHub Discovery shares the same configuration. This is not true. There are extra configuration steps required to enable GitHub Discovery.

This is just a quick fix, to show how to configure each. Ideally, we should have a little bit more detailed documentation for this. 

## PR acceptance criteria


app-config.yaml
```diff
@@ -215,12 +215,18 @@ catalog:
         serviceAccountToken: ${moc_infra_token}
         owner: janus-authors
 
-    # Plugin: GitHub Discovery & Org Data
+    # Plugin: GitHub Org Data
     githubOrg:
       default:
         id: production
         orgUrl: ${GITHUB_ORG_URL}
```
christophe-f, line 222: Can this be a list of organizations?

kadel: `orgUrl` can't be a list.

theoretically based on https://github.com/janus-idp/backstage-showcase/blob/83003f013c0a3c0cf0e8de9b6bed2eccdd5594b2/packages/backend/src/plugins/catalog.ts#L75-L94
it should be possible to specify multiple orgs like this:
```yaml
    githubOrg:
      org1:
        id: production-org1
        orgUrl: org1
      org2:
        id: production-org2
        orgUrl: org2
```

I haven't tested it. It looks weird to me, and I don't know if it can deal with conflicting items.

There is `GithubMultiOrgEntityProvider`, which we could use instead of `GithubOrgEntityProvider`. Based on docs, this is how to ingest data from multiple orgs. This one will namespace groups according to the org they originate from to avoid potential name duplicates.


If we want to change to `GithubMultiOrgEntityProvider` I can look into it in a separate PR. I would like to keep this PR documentation only.

kadel: I've tested both  `GithubOrgEntityProvider` and `GithubMultiOrgEntityProvider`.
Currently, we use `GithubOrgEntityProvider` and it is possible to specify multiple groups as I described in previous comment.

Let's say that we have two orgs that have the same group called `owners`.
`GithubOrgEntityProvider` will import only one group called `owners`, and then it will complain about duplicated groups when trying to import the second one.

I've also tested this with `GithubMultiOrgEntityProvider` and this will import both groups as `org1/owners` and `org2/owners`.

christophe-f (approved):

PR #522, opened 2023-09-12
Add dynamic backend plugins support

### LINKED ISSUE #526: Add support for the dynamic backend plugins in the showcase application

davidfestal:
## Description

Add support for dynamic backend plugins.
Add event router support.

- Fixes #526

## PR acceptance criteria


docker/Dockerfile
```diff
@@ -86,6 +86,12 @@ COPY $EXTERNAL_SOURCE_NESTED/app-config*.yaml ./
 # hadolint ignore=DL3059
 RUN $YARN install --frozen-lockfile --production --network-timeout 600000
 
+# Stage 4.5 - build dynamic plugins
+FROM skeleton AS dynamic-plugins
```
nickboldt, line 90: do we realllllly need yet another named layer? this is just another thing I have to remove downstream. What's the value in having all these names layers? Could we just have a `build` and a `runner` layer?

docker/brew.Dockerfile
```diff
@@ -128,6 +128,10 @@ RUN $YARN install --frozen-lockfile --production --network-timeout 600000
 FROM registry.redhat.io/ubi9/nodejs-18-minimal:1 AS runner
 USER 0
 
+RUN microdnf update -y && \
+  microdnf install -y python3 python3-pip && \
+  microdnf clean all
```
nickboldt, line 133: Do you also need this, as per the  upstream Dockerfile??

```
  pip3 install mkdocs-techdocs-core==1.* && \
```

docker/Dockerfile
```diff
@@ -106,6 +112,9 @@ WORKDIR $CONTAINER_SOURCE/
 # Upstream only
 COPY --from=cleanup --chown=1001:1001 $CONTAINER_SOURCE/ ./
 
+COPY --chown=1001:1001 docker/install-dynamic-plugins.py ./
+COPY --from=dynamic-plugins --chown=1001:1001 $CONTAINER_SOURCE/ ./
```
nickboldt, line 116: should this be copying only the `$CONTAINER_SOURCE/dynamic-plugins/` folder?

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,84 @@
+import os
+import sys
+import yaml
+import tarfile
+import shutil
+import subprocess
```
nickboldt, line 6: are any of these deps installed via pip? if so we need a requirements.txt

Examples:

https://github.com/redhat-developer/devspaces-jiralint/blob/devspaces-3-rhel-8/requirements.txt
https://github.com/redhat-developer/devspaces-images/tree/devspaces-3-rhel-8/devspaces-udi/build/python

docker/Dockerfile
davidfestal: No idea :-)
I just tried to respect the way it was done, but have no idea why the structures between Dockerfile and brew.Dockerfile need to be different.

docker/Dockerfile
davidfestal: True.
I'll rework this part anyway.

docker/brew.Dockerfile
davidfestal: @christophe-f could answer I assume. It mainly depends on how techdocs is supported on downstream...
I don't have sufficient context and history to answer TBF.

davidfestal: at least it seems that installing `mkdocs-techdocs-core` also installs the `yaml` package which we need in the dynamic-plugins-install.py script.

docker/install-dynamic-plugins.py
davidfestal: Yes if I remove the installation of `mkdocs-techdocs-core` by pip, the `yaml` package which we need in the dynamic-plugins-install.py script is not available anymore.
So I assume we need a requirements.txt file.

docker/Dockerfile
nickboldt: brew/osbs's implementation can't use named layers that refer to previous named layers. It can do multiple

```
FROM registry/org/container:tag as label
```

but not 

```
FROM previousLabel as newLabel
```

So downstream I ripped out all the intermediate layers that @schultzp2020 added for [reasons] and we have a working build again. If you want to follow his style, it just means more things to remove downstream in brew.Dockerfile.

packages/backend/package.json
```diff
@@ -25,10 +25,10 @@
     "@backstage/plugin-auth-backend": "0.18.8",
     "@backstage/plugin-auth-node": "0.2.19",
     "@backstage/plugin-azure-devops-backend": "0.3.29",
-    "@backstage/plugin-catalog-backend": "1.12.4",
-    "@backstage/plugin-catalog-backend-module-github": "0.3.7",
-    "@backstage/plugin-catalog-backend-module-gitlab": "0.2.6",
-    "@backstage/plugin-catalog-backend-module-openapi": "0.1.18",
+    "@backstage/plugin-catalog-backend": "^1.13.0-next",
```
schultzp2020, line 28: @davidfestal are you preparing for the 1.18.0 release? Currently our policy is not to use prereleases. 

cc @christophe-f

davidfestal: In fact I need those, because the backend-plugin-manager is not compatible with previous releases.
I relied on the expectation that finally the showcase application, will depend on the 1.18.0 release.

docker/Dockerfile
nickboldt: looks OK

docker/install-dynamic-plugins.py
davidfestal: I added a `requirements.txt` file is this commit, but I have no idea if it is the place to put it.

packages/backend/package.json
davidfestal: Now that the 1.18.0 release has been issued, I updated and fixed the version of those dependencies to the 1.18.0 ones.
Is it OK for you @schultzp2020

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,116 @@
...
+    if completed.returncode != 0:
+        raise InstallException(f'Error while installing plugin { package } with \'npm pack\' : ' + completed.stderr.decode('utf-8'))
+
+    archive = os.path.join(dynamicPluginsRoot, completed.stdout.decode('utf-8').strip())
+    directory = archive.replace('.tgz', '')
+    shutil.rmtree(directory, ignore_errors=True, onerror=None)
+    os.mkdir(directory)
+
+    # extract the plugin
+    file = tarfile.open(archive, 'r:gz')
```
schultzp2020, line 80:
```suggestion
    file = tarfile.open(archive, 'r:gz') # NOSONAR
```

This should remove the security hotspot false positive

davidfestal: Done. Thanks !

docker/install-dynamic-plugins.py
nickboldt: next to the .py script it refers to is fine. Once this is merged I can start testing it downstream ... or now that 0.2 is done, I can start buildin RHDH from this fork/branch, if you want to get some input on how the python cachito stuff works.

docker/brew.Dockerfile
```diff
@@ -128,6 +132,11 @@ RUN $YARN install --frozen-lockfile --production --network-timeout 600000
 FROM registry.redhat.io/ubi9/nodejs-18-minimal:1 AS runner
 USER 0
 
+RUN microdnf update -y && \
+  microdnf install -y python3 python3-pip && \
+  pip3 install mkdocs-techdocs-core==1.* && \
+  microdnf clean all
+
 # Downstream only - do not install techdocs dependencies (not required)
```
nickboldt, line 140: I guess we can remove this comment since we ARE installing the techdocs reqs now.

nickboldt, line 140:
```suggestion
```

docker/brew.Dockerfile
```diff
@@ -128,6 +132,11 @@ RUN $YARN install --frozen-lockfile --production --network-timeout 600000
 FROM registry.redhat.io/ubi9/nodejs-18-minimal:1 AS runner
 USER 0
 
+RUN microdnf update -y && \
+  microdnf install -y python3 python3-pip && \
+  pip3 install mkdocs-techdocs-core==1.* && \
```
nickboldt, line 137: this won't work downstream -- I'll draft you something as a replacement based on https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/Dockerfile#L226-L229 and https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-udi/container.yaml#L33-L40

docker/Dockerfile
```diff
@@ -106,6 +109,14 @@ WORKDIR $CONTAINER_SOURCE/
 # Upstream only
 COPY --from=cleanup --chown=1001:1001 $CONTAINER_SOURCE/ ./
 
+# Copy python script used to gather dynamic plugins
```
nickboldt, line 112: please remove the comment on line 96 above since "install techdocs dependencies" is no longer an upstream only thing (it's done differently downstream, but it WILL be downstream -- so we can have a different section entirely).

docker/brew.Dockerfile
nickboldt, line 140:
```suggestion
# Downstream only - install techdocs dependencies using cachito sources
```

docker/brew.Dockerfile
davidfestal: Thanks for noticing this! Let's do that in a followup PR, sure.

docker/Dockerfile
davidfestal: Done.

docker/brew.Dockerfile
davidfestal: Done

nickboldt (approved):

nickboldt:
Test `CI / Docker Build (pull_request)` will never pass if your PR is from your personal fork. It needs to be from the janus-idp origin for that to pass.

nickboldt:
Note: working on downstream in https://issues.redhat.com/browse/RHIDP-425

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,118 @@
...
+    """Exception class from which every exception in this library will derive."""
+    pass
+
+# test that plugins is a list
+if not isinstance(plugins, list):
+    raise InstallException(f'{dynamicPluginsFile} content must be a list')
+
+globalConfig = {
+  'dynamicPlugins': {
+        'rootDirectory': 'dynamic-plugins-root'
```
schultzp2020, line 42:
```suggestion
        'rootDirectory': 'dynamic-plugins'
```

The dockerfile uses `dynamic-plugins`

showcase-docs/dynamic-plugins.md
```diff
@@ -0,0 +1,271 @@
...
+- include the additions described above,
+- export it as a dynamic plugin.
+
+An example of such a wrapper plugin can be found in the [dynamic backend plugin showcase repository](https://github.com/janus-idp/dynamic-backend-plugins-showcase/tree/main/plugins/scaffolder-backend-module-http-request-wrapped): it wraps the `@roadiehq/scaffolder-backend-module-http-request` package to make it compatible with the dynamic plugin support and, through the use of the `--embed-package` option in the [`export-dynamic` script](https://github.com/janus-idp/dynamic-backend-plugins-showcase/blob/d254c065764ab49289c2bcaad9fd996d49003f9d/plugins/scaffolder-backend-module-http-request-wrapped/package.json#L21), embeds the wrapped plugin code in the generated code and hoist its `@backstage` dependencies as peer dependencies in the resulting dynamic plugin.
+
+## Installing a dynamic plugin package in the showcase
+
+### Local configuration
+
+- Create a `dynamic-plugins-root` folder at the root of the showcase application repository.
```
schultzp2020, line 163:
```suggestion
- Create a `dynamic-plugins` folder at the root of the showcase application repository.
```

The dockerfile uses `dynamic-plugins`

showcase-docs/dynamic-plugins.md
```diff
@@ -0,0 +1,271 @@
...
+
+- In the `app-config.yaml`` file, add the following entry:
+
+  ```yaml
+  dynamicPlugins:
+    rootDirectory: dynamic-plugins-root
```
schultzp2020, line 169:
```suggestion
    rootDirectory: dynamic-plugins
```

The dockerfile uses dynamic-plugins

showcase-docs/dynamic-plugins.md
```diff
@@ -0,0 +1,271 @@
...
+  tar -xzf "$archive" && rm "$archive"
+  mv package $(echo $archive | sed -e 's:\.tgz$::')
+  ```
+
+  It will create a sub-folder named after the package name, and containing the dynamic plugin package.
+
+- Start the showcase application. During the initialization step it should have a log entry similar to the following:
+
+  ```
+  backstage info loaded dynamic backend plugin '@scope/some-plugin-dynamic' from 'file:///showacase-root/dynamic-plugins-root/scope-some-plugin-dynamic-0.0.1'
```
schultzp2020, line 186:
```suggestion
  backstage info loaded dynamic backend plugin '@scope/some-plugin-dynamic' from 'file:///showacase-root/dynamic-plugins/scope-some-plugin-dynamic-0.0.1'
```

The dockerfile uses dynamic-plugins

showcase-docs/dynamic-plugins.md
schultzp2020, line 163: Should we point them to `./install-dynamic-plugins.py` instead?

schultzp2020 (changes requested):

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,118 @@
...
+dynamicPluginsFile = os.path.join(dynamicPluginsRoot, 'dynamic-plugins.yaml')
+dynamicPluginsGlobalConfigFile = os.path.join(dynamicPluginsRoot, 'app-config.dynamic-plugins.yaml')
+
+# test if file dynamic-plugins.yaml exists
+if not os.path.isfile(dynamicPluginsFile):
+    print(f'No {dynamicPluginsFile} file found. Skipping dynamic plugins installation.')
+    with open(dynamicPluginsGlobalConfigFile, 'w') as file:
+        file.write('')
+        file.close()
+    exit(0)
```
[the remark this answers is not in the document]
davidfestal: We don'"t want that, because we don't want the init container to fail.
This is a valid case, exit fast, but valid.

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,118 @@
+import os
+import sys
+import yaml
+import tarfile
+import shutil
+import subprocess
+
+dynamicPluginsRoot = sys.argv[1]
```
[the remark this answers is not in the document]
davidfestal: what does it bring ?

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,118 @@
...
+
+with open(dynamicPluginsFile, 'r') as file:
+  plugins = yaml.safe_load(file)
+
+if plugins == '' or plugins is None:
+    print(f'{dynamicPluginsFile} file is empty. Skipping dynamic plugins installation.')
+    with open(dynamicPluginsGlobalConfigFile, 'w') as file:
+        file.write('')
+        file.close()
+    exit(0)
```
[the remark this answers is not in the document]
davidfestal: We don'"t want that, because we don't want the init container to fail.
This is a valid case, exit fast, but valid.

docker/install-dynamic-plugins.py
davidfestal: No, that's not the same.
`dynamic-plugins` is the source folder for dynamic plugins that will be built inside the docker image
`dynamic-plugins-root` is the destination folder in which all the dynamic plugins packages (not the sources) will be unpacked.

showcase-docs/dynamic-plugins.md
davidfestal: See answer above: https://github.com/janus-idp/backstage-showcase/pull/522#discussion_r1344704182

showcase-docs/dynamic-plugins.md
davidfestal: https://github.com/janus-idp/backstage-showcase/pull/522#discussion_r1344704182

showcase-docs/dynamic-plugins.md
davidfestal: see answer above: https://github.com/janus-idp/backstage-showcase/pull/522#discussion_r1344704182

showcase-docs/dynamic-plugins.md
davidfestal: I don't think so. There's not use for this. This seems over-complicated for the local case.

docker/install-dynamic-plugins.py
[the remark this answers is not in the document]
schultzp2020: The `argparse` module makes it easy to write user-friendly command-line interfaces.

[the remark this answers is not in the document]
davidfestal: This command is not expected to be run manually. It is part of the docker file definition. I’m not convinced of the added value of using argparse in this context

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,118 @@
...
+import sys
+import yaml
+import tarfile
+import shutil
+import subprocess
+
+dynamicPluginsRoot = sys.argv[1]
+maxEntrySize = os.environ.get('MAX_ENTRY_SIZE', 10000000)
+
+dynamicPluginsFile = os.path.join(dynamicPluginsRoot, 'dynamic-plugins.yaml')
```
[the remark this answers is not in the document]
davidfestal: Done in commit 7b7452f2800e4669b37cc8e00ebdacc78af5ec30

docker/install-dynamic-plugins.py
```diff
@@ -0,0 +1,121 @@
...
+        print('\t==> Merging plugin-specific configuration', flush=True)
+        config = plugin['pluginConfig']
+        if config is not None and isinstance(config, dict):
+                merge(config, globalConfig)
+
+        print('\t==> Successfully installed dynamic plugin', package, flush=True)
+
+    yaml.safe_dump(globalConfig, open(dynamicPluginsGlobalConfigFile, 'w'))
+
+main()
```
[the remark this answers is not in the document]
davidfestal: From the link you provided, in the [When Should You Avoid the Name-Main Idiom](https://realpython.com/if-name-main-python/#when-should-you-avoid-the-name-main-idiom) header, I read:

> There are also other scenarios when it’s best to avoid the name-main idiom in Python:
> 
> - A pure script: If you write a script that’s _meant_ to be run as a script, then you can put your code execution into the global namespace without nesting it in the name-main idiom. You _can_ use Python as a scripting language because it doesn’t enforce strong object-oriented patterns. You don’t have to stick to design patterns from other languages when you program in Python.
> 

*meant to be run as a script* is **precisely** what this file is.

I agreed to wrap the content into a `main()` function, because it is known to increase performance.
But I don't see why adding this **Name-Main Idiom** in this file, which would decrease readability IMO.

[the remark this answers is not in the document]
schultzp2020: > But I don't see why adding this Name-Main Idiom in this file, which would decrease readability IMO.

If others have dealt with python before they would also know about this idiom, even if it's annoying :(. The reason we may want to add the name-main idiom is for testing purposes. If we want to test this file we should add the name-main idiom. If not, I am fine avoiding the idiom.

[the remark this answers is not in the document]
davidfestal: As for testing trough this idiom, in the same section of the link you provided, I read:

> While this works for small files, it’s generally not considered good practice. It’s not advised to mix tests and code in the same file. Instead, write your tests in a separate file. Following this advice generally makes for a more organized code base. This approach also removes any overhead, such as the need to import `unittest` in your main script file.

In our case I'd rather avoid testing code land in the docker image.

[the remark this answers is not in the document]
davidfestal: > Do we want to add a CI test for this script?

Not in this PR., but this could be in a followup PR :-)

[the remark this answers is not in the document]
davidfestal: > If we want to test this file we should add the name-main idiom. If not, I am fine avoiding the idiom.

So is it OK for you to resolve this conversation then ?

schultzp2020 (approved):
LGTM

PR #599, opened 2023-10-10
feat: add `scalprum` backend plugin.

### LINKED ISSUE #583: Create a `DynamicFrontendPluginProvider` backend plugin, and integrate it with the frontend dynamic plugins POC
Create a `DynamicFrontendPluginProvider` backend plugin, and use it to integrate it with the frontend dynamic plugins work .

davidfestal:
## Description
Add a `scalprum` backend plugin which:
- depends on the `backend-plugin-manager`
- provides access to the scalprum defintinions,
as well as the plugin assets,
of the dynamic frontend plugins which are
installed in the dynamic plugins root folder.

- Fixes #583

## PR acceptance criteria


packages/backend/src/index.ts
```diff
@@ -168,6 +169,17 @@ async function main() {
...
+  // Scalprum frontend plugins provider
+  const scalprumEmv = useHotMemoize(module, () => createEnv('scalprum'));
+  apiRouter.use(
+    '/scalprum',
+    await scalprumRouter({
+      logger: scalprumEmv.logger,
+      pluginManager,
+      discovery: scalprumEmv.discovery,
+    }),
+  );
```
tumido, line 181: can't you use `addPlugin` factory?

davidfestal: No, because the `createRouter` method expected in `addPlugin` takes a `PluginEnvironment` as an argument.
But the `LegacyPluginEnvironment` provides a `PluginProvider` interface, and not the `PluginManager` class.
But in this case we need the `PluginManager (to get the frontend plugins, for example).
I assume that the PluginProvider should be completed on upstream, but in the meantime we can do this way.

davidfestal:
@tumido is my answer [here](https://github.com/janus-idp/backstage-showcase/pull/599#discussion_r1357417243) OK for you ?

@Hyperkid123 Are the data returned by the API what you expect ?

Being able to merge this soon would be great.

tumido (approved):
/lgtm
/approve

kadel (approved):

PR #600, opened 2023-10-11
fix : fix a bug in the upstream CommonJSModuleLoader

davidfestal:
## Description

Fix a bug in the upstream `CommonJSLoader`, which prevented laoding modules from embedded `node_modules` folders of private packages located in the plugin `node_modules` folder.

This is a fixed version of the [corresponding upstream code](https://github.com/backstage/backstage/blob/master/packages/backend-plugin-manager/src/loader/CommonJSModuleLoader.ts).

A PR on the upstream repository will follow, but until the change lands upstream and can be imported into the showcase after the next release, we will use the fixed loader added in the showcase application itself.

No issue

## PR acceptance criteria

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -49,6 +49,7 @@ import {
   LegacyPluginEnvironment as PluginEnvironment,
 } from '@backstage/backend-plugin-manager';
 import { DefaultEventBroker } from '@backstage/plugin-events-backend';
+import { CommonJSModuleLoader } from './loader/CommonJSModuleLoader';
 
 function makeCreateEnv(config: Config, pluginProvider: BackendPluginProvider) {
   const root = getRootLogger();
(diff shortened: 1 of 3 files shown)


packages/backend/src/index.ts
```diff
@@ -49,6 +49,7 @@ import {
   LegacyPluginEnvironment as PluginEnvironment,
 } from '@backstage/backend-plugin-manager';
 import { DefaultEventBroker } from '@backstage/plugin-events-backend';
+import { CommonJSModuleLoader } from './loader/CommonJSModuleLoader';
```
kadel, line 52: Please add a comment before this import explaining that this is a temporary fix.
 I'm afraid that we will forget why it is there :-/

kadel (commented):
I would just add one small comment, otherwise LGTM

packages/backend/src/index.ts
davidfestal: Added in commit 3976fb9d43f0bee97deb882fd7362827620aaf9e

kadel (approved):

PR #601, opened 2023-10-11
chore: Prepare for the switch of backend plugins to dynamic plugins

davidfestal:
## Description

This PR fixes the dynamic plugins build by adding the packages of the various dynamic plugins wrappers to the monorepo, which allows leveraging the whole yarn + turbo machinery, also for dynamic plugins. 

It adds the switch of a first, very simple plugin (roadie utils) to dynamic, in order to test and validate the whole machanism.

This PR is based on PR #600 which should be merged before.

No dedictaed issue. Part of a wider task.

## PR acceptance criteria

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -46,6 +46,7 @@ RUN chmod +x $YARN
 # Stage 2 - Install dependencies
 FROM skeleton AS deps
 
+COPY $EXTERNAL_SOURCE_NESTED/dynamic-plugins/ ./dynamic-plugins/
 COPY $EXTERNAL_SOURCE_NESTED/package.json $EXTERNAL_SOURCE_NESTED/yarn.lock ./
 COPY $EXTERNAL_SOURCE_NESTED/packages/app/package.json ./packages/app/package.json
 COPY $EXTERNAL_SOURCE_NESTED/packages/backend/package.json ./packages/backend/package.json
@@ -66,7 +67,18 @@ RUN rm app-config.yaml && mv app-config.example.yaml app-config.yaml
 RUN $YARN build --filter=backend
 
 # Build dynamic plugins
-RUN $YARN --cwd ./dynamic-plugins export-dynamic
+RUN $YARN export-dynamic
+RUN $YARN clean-dynamic-sources
+RUN mkdir -p dynamic-plugins-root && \
+    cd dynamic-plugins-root && \
+    rm -Rf * && \
+    for pkg in $CONTAINER_SOURCE/dynamic-plugins/*/dist-dynamic; do \
+      if [ -d $pkg ]; then \
+        archive=$(npm pack $pkg) && \
+        tar -xzf "$archive" && rm "$archive" && \
+        mv package $(echo $archive | sed -e 's:\.tgz$::'); \
+      fi; \
+    done
 
 # Stage 4 - Build the actual backend image and install production dependencies
 FROM skeleton AS cleanup
@@ -117,6 +129,10 @@ RUN chmod a+r ./install-dynamic-plugins.py
 # Copy embedded dynamic plugins
 COPY --from=build $CONTAINER_SOURCE/dynamic-plugins/ ./dynamic-plugins/
 RUN chmod -R a+r ./dynamic-plugins/
+
+# Copy default dynamic plugins root
+COPY --from=build $CONTAINER_SOURCE/dynamic-plugins-root/ ./dynamic-plugins-root/
+RUN chmod -R a+r ./dynamic-plugins-root/
 
 # The fix-permissions script is important when operating in environments that dynamically use a random UID at runtime, such as OpenShift.
 # The upstream backstage image does not account for this and it causes the container to fail at runtime.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -96,6 +96,7 @@ RUN tar xzf $TARBALL_PATH/skeleton.tar.gz; tar xzf $TARBALL_PATH/bundle.tar.gz; 
 # Copy app-config files needed in runtime
 # Upstream only
 COPY $EXTERNAL_SOURCE_NESTED/app-config*.yaml ./
+COPY $EXTERNAL_SOURCE_NESTED/dynamic-plugins.default.yaml ./
 
 # Install production dependencies
 # hadolint ignore=DL3059
--- a/docker/brew.Dockerfile
+++ b/docker/brew.Dockerfile
@@ -46,6 +46,7 @@ COPY $EXTERNAL_SOURCE_NESTED/.yarnrc.yml ./
 RUN chmod +x $YARN
 
 # Stage 2 - Install dependencies
+COPY $EXTERNAL_SOURCE_NESTED/dynamic-plugins/ ./dynamic-plugins/
 COPY $EXTERNAL_SOURCE_NESTED/package.json $EXTERNAL_SOURCE_NESTED/yarn.lock ./
 COPY $EXTERNAL_SOURCE_NESTED/packages/app/package.json ./packages/app/package.json
 COPY $EXTERNAL_SOURCE_NESTED/packages/backend/package.json ./packages/backend/package.json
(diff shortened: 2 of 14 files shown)


docker/Dockerfile
```diff
@@ -65,7 +66,18 @@ RUN rm app-config.yaml && mv app-config.example.yaml app-config.yaml
...
+RUN $YARN export-dynamic
+RUN $YARN clean-dynamic-sources
+RUN mkdir -p dynamic-plugins-root && \
+    cd dynamic-plugins-root && \
+    rm -Rf * && \
+    for pkg in $CONTAINER_SOURCE/dynamic-plugins/*/dist-dynamic; do \
+      if [ -d $pkg ]; then \
+        archive=$(npm pack $pkg) && \
+        tar -xzf "$archive" && rm "$archive" && \
+        mv package $(echo $archive | sed -e 's:\.tgz$::'); \
```
nickboldt, line 78:
```suggestion
        tar -xzf "$archive" --strip-components=1 -C "$${archive#.tgz}" && rm "$archive" && \
```

docker/Dockerfile
```diff
@@ -65,7 +66,18 @@ RUN rm app-config.yaml && mv app-config.example.yaml app-config.yaml
...
-RUN $YARN --cwd ./dynamic-plugins export-dynamic
+RUN $YARN export-dynamic
+RUN $YARN clean-dynamic-sources
+RUN mkdir -p dynamic-plugins-root && \
+    cd dynamic-plugins-root && \
+    rm -Rf * && \
+    for pkg in $CONTAINER_SOURCE/dynamic-plugins/*/dist-dynamic; do \
+      if [ -d $pkg ]; then \
+        archive=$(npm pack $pkg) && \
+        tar -xzf "$archive" && rm "$archive" && \
```
nickboldt, line 77: so, pack it into a tarball, then tar.gz unpack it, and move/rename into a named folder? 

I wonder if it would be cleaner this way...

```suggestion
        mkdir -p "${archive#.tgz}" &&\
```

docker/brew.Dockerfile
```diff
@@ -107,8 +108,18 @@ RUN git config --global --add safe.directory ./
...
-RUN $YARN --cwd ./dynamic-plugins export-dynamic
+RUN $YARN export-dynamic
+RUN $YARN clean-dynamic-sources
+RUN mkdir -p dynamic-plugins-root && \
+    cd dynamic-plugins-root && \
+    rm -Rf * && \
+    for pkg in $CONTAINER_SOURCE/dynamic-plugins/*/dist-dynamic; do \
+      if [ -d $pkg ]; then \
+        archive=$(npm pack $pkg) && \
+        tar -xzf "$archive" && rm "$archive" && \
```
nickboldt, line 119: see above; you can do this a bit simpler if desired

dynamic-plugins/.gitkeep
nickboldt, line 1: rather than a .gitkeep, would a .gitignore be useful here to exclude things we never want to commit accidentally?

package.json
```diff
@@ -51,7 +54,8 @@
   "prettier": "@spotify/prettier-config",
   "lint-staged": {
     "*": "yarn run prettier:fix",
-    "*.{js,jsx,ts,tsx,mjs,cjs}": "yarn run lint -- -- --fix"
+    "*.{jsx,ts,tsx,mjs,cjs}": "yarn run lint -- -- --fix",
+    "!(.eslintrc).js": "yarn run lint -- -- --fix"
```
nickboldt, line 58: this syntax for "all `.js` except `.eslintrc.js` looks weird to me, but if you've tested and it works, then +1.

nickboldt (approved):

dynamic-plugins/.gitkeep
davidfestal: `.gitignore` is already set at the root of the repo.

docker/Dockerfile
davidfestal: TBH, I'd rather go for the current way, so that I don't have to test it again :-)

docker/brew.Dockerfile
davidfestal: TBH, I'd rather go for the current way, so that I don't have to test it again :-)

packages/backend/src/index.ts
```diff
@@ -113,12 +113,28 @@ type AddPlugin = {
...
+const optionalDynamicPlugins: { [key: string]: OptionalPluginOptions } = {
+  techdocs: {},
+  argocd: {},
+  sonarqube: {},
+  kubernetes: {},
+  'azure-devops': { key: 'enabled.azureDevOps' },
+  jenkins: {},
+  ocm: {},
+  gitlab: {},
+};
```
schultzp2020, line 136:
```suggestion
const OPTIONAL_DYNAMIC_PLGUINS = {
  techdocs: {},
  argocd: {},
  sonarqube: {},
  kubernetes: {},
  'azure-devops': { key: 'enabled.azureDevOps' },
  jenkins: {},
  ocm: {},
  gitlab: {},
} as const satisfies { [key: string]: OptionalPluginOptions };
```

Please use MACRO_CASE and the `as const` assertion when specifying constants 🙂

docker/install-dynamic-plugins.py
```diff
@@ -95,8 +137,30 @@ def main():
                 file.extract(member, path=directory)
             elif member.isdir():
                 print('\t\tSkipping directory entry', member.name, flush=True)
+            elif member.islnk() or member.issym():
+                if not member.linkpath.startswith('package/'):
+                  raise InstallException('NPM package archive contains a link outside of the archive: ' + member.name + ' -> ' + member.linkpath)
+
+                member.name = member.name[8:]
+                member.linkpath = member.linkpath[8:]
```
schultzp2020, line 145: Instead of using string slicing lets use a regex instead. This way it's easier to track if we change the `package/` location. We should also update line 136 as well. 

```suggestion
import re

locationRegex = r"package\/(.*)"

...

                member.name = re.search(r"package\/(.*)", member.name).group(1)
                member.linkpath = re.search(r"package\/(.*)", member.linkpath).group(1)
```

docker/install-dynamic-plugins.py
```diff
@@ -95,8 +137,30 @@ def main():
                 file.extract(member, path=directory)
             elif member.isdir():
                 print('\t\tSkipping directory entry', member.name, flush=True)
+            elif member.islnk() or member.issym():
+                if not member.linkpath.startswith('package/'):
+                  raise InstallException('NPM package archive contains a link outside of the archive: ' + member.name + ' -> ' + member.linkpath)
+
+                member.name = member.name[8:]
```
schultzp2020, line 144: Minor nit, it's considered best practice to not reassign a variable that iterates. If possible, we should create a new variable instead. 

```suggestion
                name = member.name[8:]
```

package.json
schultzp2020: Why does `.eslintrc.js` need to be excluded? I'm fine excluding it, I'm just a little curious to what the error is.

schultzp2020 (changes requested):

package.json
davidfestal: Because without it, the `yarn lint-staged` command fails (in a number of case, according to what you stage for commit) with many errors like the following:

```
app:lint: 
app:lint: 
app:lint:   ⚠  http://eslint.org/docs/rules/
app:lint: 
app:lint:      File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override
app:lint: 
app:lint: 
app:lint:      dynamic-plugins/backstage-plugin-aap-backend-wrapped/.eslintrc.js:0:0
app:lint: 
app:lint: ✘ 1 problem (0 errors, 1 warning)
app:lint: 
app:lint: 
app:lint: Warnings:
app:lint:   1  http://eslint.org/docs/rules/null
app:lint: 
```

docker/install-dynamic-plugins.py
davidfestal: There's not reason to change the `package/` location. It's part of how archives built by `npm pack` are structured.
That's why refuse any archive member which doesn't start with `package`.

docker/install-dynamic-plugins.py
davidfestal: Here precisely want to change the name field of the `member` variable (remove the `package` suffix) , so that the member is extracted with the right name by the `file.extract(member, path=directory)` call.

schultzp2020: Oh, I must've missed that.

docker/install-dynamic-plugins.py
schultzp2020: Can we use a regex instead of array slicing to assign `name` and `linkpath`? I think it's easier to read on a passthrough.

davidfestal: Fixed in commit 47e9246829cdbaa30ae780ab7fb1fe7d8d75cfd1 as discussed with the `removeprefix` function instead of a regexp (even more readable)

schultzp2020 (approved):
LGTM

Issue #591, opened 2023-10-06
Configure Additional Auth Providers


PatAKnight:
### What do you want to improve?
Backstage has the ability to configure many different authentication providers for sign in. I believe it is important to capture some of those providers in showcase to allow for more options for signing in.. 

### What is the current behavior?
Currently, we only support Guest, [GitHub](https://backstage.io/docs/auth/github/provider), and [OAuth2Proxy](https://backstage.io/docs/auth/oauth2-proxy/provider) sign on providers

### What is the new behavior?
Include more authentication providers to allow for more options. We can also attempt to add the ability to enable / disable the providers based on the app-config. 

### Acceptance Criteria
- Authentication providers configured
- Ability to enable / disable authentication providers based on app-config
- Documentation on how to configure the different authentication providers

PatAKnight:
@christophe-f  What authentication providers do we want to make sure that we capture? I was thinking [GitLab](https://backstage.io/docs/auth/gitlab/provider) and [Bitbucket](https://backstage.io/docs/auth/bitbucket/provider) since we already have some features included in showcase for them.

PR #602, opened 2023-10-11
feat(showcase): add conditional rendering for login page

schultzp2020:
## Description

- References https://github.com/janus-idp/backstage-showcase/issues/591
- Unblocks https://github.com/janus-idp/backstage-showcase/issues/645
- Blocked by https://github.com/backstage/backstage/issues/20877

## PR acceptance criteria


packages/app/src/components/SignInPage/SignInPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+    },
+  ],
+  ['oauth2Proxy', 'oauth2Proxy'],
+  [
+    'onelogin',
+    {
+      id: 'onelogin-auth-provider',
+      title: 'OneLogin',
+      message: 'Sign in using OneLogin',
+      apiRef: oktaAuthApiRef,
```
Zaperex, line 125:
```suggestion
      apiRef: oneloginAuthApiRef,
```

Zaperex:
Tried testing it with [gitlab](https://backstage.io/docs/auth/gitlab/provider/) and got the following error:
image

It appears we also need to configure the [auth provider integrations](https://backstage.io/docs/reference/plugin-auth-backend.providers) in the `packages/backend/src/plugins/auth` file and enable sign-in, to be able to allow users to login via these auth providers. Similar to how we configured the [github auth provider integration](https://github.com/janus-idp/backstage-showcase/blob/be3048497b80291ca42a9a19f55c6096488dfffe/packages/backend/src/plugins/auth.ts#L43-L74)

packages/app/src/components/SignInPage/SignInPage.tsx
schultzp2020: Good catch

gorkem:
@Zaperex  Test this PR together with #588 which brings signIn implementation for some of the providers already.

packages/app/src/api/JanusBackstageCustomizeApiClient.ts
```diff
@@ -0,0 +1,69 @@
...
+    const proxyUrl = await this.getBaseUrl();
+    const data = await this.fetcher(`${proxyUrl}`);
+    return data;
+  }
+
+  async getTechRadarDataJson() {
+    const proxyUrl = await this.getBaseUrl();
+    const data = await this.fetcher(`${proxyUrl}/tech-radar`);
+    return data;
+  }
```
christophe-f, line 68: Do we still need this? @invincibleJai

packages/app/src/api/index.ts
```diff
@@ -1,69 +1,2 @@
...
-  }
-
-  async getTechRadarDataJson() {
-    const proxyUrl = await this.getBaseUrl();
-    const data = await this.fetcher(`${proxyUrl}/tech-radar`);
-    return data;
-  }
-}
+export * from './AuthApiRefs';
+export * from './JanusBackstageCustomizeApiClient';
```
christophe-f, line 2: We can rename without `Janus` in the file name

packages/app/src/lib/CustomTechRadar.ts
```diff
@@ -2,7 +2,7 @@ import {
   TechRadarApi,
   type TechRadarLoaderResponse,
 } from '@backstage/plugin-tech-radar';
-import { JanusBackstageCustomizeApi } from '../api';
+import { JanusBackstageCustomizeApi } from '../api/JanusBackstageCustomizeApiClient';
 
 export class CustomTechRadar implements TechRadarApi {
   private readonly janusBackstageCustomizeApi: JanusBackstageCustomizeApi;
```
christophe-f, line 8: No need to use `Janus` in the name

packages/app/src/api/JanusBackstageCustomizeApiClient.ts
schultzp2020: @christophe-f yes

- [`getHomeDataJson()`](https://github.com/janus-idp/backstage-showcase/blob/conditional-login-page/packages/app/src/hooks/useQuickAccess.ts#L16)
- [`getTechRadarDataJson()`](https://github.com/janus-idp/backstage-showcase/blob/conditional-login-page/packages/app/src/lib/CustomTechRadar.ts#L15)

packages/app/src/components/SignInPage/SignInPage.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Key:
+ * string - Provider name.
+ *
+ * Value:
+ * SignInProviderConfig - Local sign-in provider configuration.
+ * string - Proxy sign-in provider configuration.
+ */
+const PROVIDERS = new Map<string, SignInProviderConfig | string>([
```
Zaperex, line 33: Currently, it seems that only one provider can be configured and when multiple are configured, the provider higher up in this list of providers takes precedence over all the other providers. Since currently our "default" `app-config.yaml` files [include the github auth provider](https://github.com/janus-idp/backstage-showcase/blob/0d27cbbda96f6ae2bde21bb1540a3e326b84ac18/app-config.yaml#L200-L203) by default, the providers below github such as gitlab, google oidc, otka, saml and onelogin will never be configurable unless they completely remove the `app-config.yaml` and `app-config.example.yaml` that the image is loaded with.

Would we want to change up the behavior to allow all of the configured providers to render rather than only allowing one to appear?

Zaperex: But I guess that wouldn't fix the issue of the github auth provider being rendered even if the user doesn't configure it.

schultzp2020: > Would we want to change up the behavior to allow all of the configured providers to render rather than only allowing one to appear?

@Zaperex No, we do not want to change this behavior because it would be a security concern. The more sign in methods we provide at the same time, the more attack vectors there are for bad actors. The platform engineer will need to provide their own `app-config` anyways so this should be a non-issue.

schultzp2020:
Blocked by https://github.com/backstage/backstage/issues/20877

showcase-docs/getting-started.md
```diff
@@ -132,20 +132,30 @@ The easiest and fastest method for getting started: Backstage Showcase app, runn
...
+       - GitLab - `gitlab`
+       - Google - `google`
+       - Google IAP - `gcp-iap`
+       - OIDC - `oidc`
+       - Okta - `okta`
+       - OAuth 2 Custom Proxy - `oauth2Proxy`
+       - OneLogin - `onelogin`
+       - SAML - `saml`
+
+     - Add the corresponding authentication provider key as the value to `signInPage` in your `app-config`.
```
PatAKnight, line 152: Should we include something along the lines of 'To disabled the Guest login set `auth.environment` to `production`'?

PatAKnight (commented):
Minor question but otherwise looks good

showcase-docs/getting-started.md
kadel: this brings me to the question, what if someone wants to enable guest in production environment? Is that possible?

kadel (approved):
lgmt

I have a few questions regarding functionality for example, how one enable guest login if the environment is `production` or what if someone needs multiple auth providers in sign-in page.
But neither of these is blocking.

schultzp2020:
> lgmt
> 
> I have a few questions regarding functionality for example, how one enable guest login if the environment is `production` or what if someone needs multiple auth providers in sign-in page. But neither of these is blocking.

Yeah I was debating the same thing. My conclusion was that users should really only configure one sign-in page, if this is not the case we can easily change this. The one issue is that the proxy page can ONLY have ONE option. I have also yet to find a case where logging in is not required for a company.

PatAKnight (approved):
LGTM

PR #694, opened 2023-10-31
chore(config): cleanup app-config files

schultzp2020:
## Description

- Unblocks https://github.com/janus-idp/backstage-showcase/pull/602

## PR acceptance criteria


schultzp2020:
@PatAKnight can you verify that this does not impact you?

PatAKnight (approved):
LGTM

app-config.yaml
```diff
@@ -21,8 +21,9 @@ app:
         headerColor1: ${HEADER_DARK_COLOR_1}
         headerColor2: ${HEADER_DARK_COLOR_2}
         navigationIndicatorColor: ${NAV_INDICATOR_DARK_COLOR}
+
 organization:
-  name: My Org
+  name: ${ORGANIZATION_NAME}
```
kadel, line 26: It looks like this introduces new variable, we should document it in https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md

app-config.production.yaml
```diff
@@ -33,12 +34,3 @@ backend:
       ssl:
         ca: # if you have a CA file and want to verify it you can uncomment this section
           $file: /mnt/certs/ca.crt
-
-auth:
-  environment: production
```
kadel, line 38: Is it possible to keep just following? 
```yaml

auth:
  environment: production
```

kadel (changes requested):
There is a lot of changes  and it invalidates most of the things in https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md it should be updated to reflect this change

schultzp2020:
@kadel I don't think it's worth while to update the getting started docs because it will need a huge rewrite because of dynamic plugins. We were also debating on deleting the non example `app-config`s.

kadel:
> @kadel I don't think it's worth while to update the getting started docs because it will need a huge rewrite because of dynamic plugins.

We should try to keep it up to date where it makes sense. If we let updates pile up it something important could get lost.



> We were also debating on deleting the non example `app-config`s.

+1 to that, we need to change the approach to configs and reduce the number of overlayed files.

davidfestal:
I agree that such a change will be required in a near future.
But it seems to me that it might be premature, since the [new Helm chart PR](https://github.com/janus-idp/helm-backstage/pull/111) is not even merged, and the frontend dyamic plugins is still in the process of landing, but not fully merged in upstream.

Until the switch to dynamic plugins is completely done (frontend + backend + helm chart), I think it is not a good choice to break compatibility with how the `enable.xxx` properties and default config file values used to work.  

To me the plan should be the following:
- finish frontend dynamic plugins integration
- merge / publish the new Helm Chart
- then the new helm chart will provide app-config values which will override those in the app-config.xxx.yaml files included in the container image. So this will work, while keeping compatibility with the old Helm chart.
- When the new Helm chart is published and widely used, and we have agreed on and documented the fact that no plugin is loaded by default, then we would be able to cleanup the app-config.xxx.yaml files the way it is done here, without breaking any user expectation.

davidfestal (changes requested):
Seems premature to me (see comment [here](https://github.com/janus-idp/backstage-showcase/pull/694#issuecomment-1790644823)), though I'm not saying that part of this cleanup could not be merged now.

christophe-f:
This is a blocking PR. So we should keep it to the minimum to unblock the conditional login. There is a clean up needed but that can be done later as it seems that there is a lot of considerations to look at.

schultzp2020:
@davidfestal the non example `app-config`s are actually our configs. The only config the user worries about is the example config because they will supply their own config.

schultzp2020:
> Until the switch to dynamic plugins is completely done (frontend + backend + helm chart), I think it is not a good choice to break compatibility with how the enable.xxx properties and default config file values used to work.

The compatibility will not be broken because the `enabled` section of the config will be added by the respective dynamic plugin installation.

schultzp2020:
@davidfestal and I just had a quick call about this. We think the best way forward is to create a new `app-config` called `app-config.deprecated.yaml` that is a copy of the old `app-config.yaml`. This way users can still get the old behavior by adding an arg in the helm chart to point to the old `app-config` at `app-config.deprecated.yaml`. Thoughts? @christophe-f @kadel

tumido:
Does showcase have a version associated with it? We can always say the Helm chart supports images up to certain `appVersion`:

https://helm.sh/docs/topics/charts/#the-appversion-field

I think the `app-config.deprecated.yaml` is a good solution if we need to offer backward compatibility. The question is - do we need to offer that?

kadel:
> Does showcase have a version associated with it? We can always say the Helm chart supports images up to certain `appVersion`:
> 
> https://helm.sh/docs/topics/charts/#the-appversion-field
> 
> I think the `app-config.deprecated.yaml` is a good solution if we need to offer backward compatibility. The question is - do we need to offer that?

I wouldn't worry about backward compatibility until we are at 1.0

kadel (approved):

davidfestal (approved):
Approving, to let others decide.

PR #704, opened 2023-11-02
fix: fix dynamic plugin package integrity verification

davidfestal:
## Description

Fix a bug in the dynamic plugin package integrity check that prevents disabling this feature as expected.

No issue number. Issue fixed while working on some other task.

## PR acceptance criteria

### DIFF
--- a/docker/install-dynamic-plugins.py
+++ b/docker/install-dynamic-plugins.py
@@ -107,7 +107,6 @@ def verify_package_integrity(plugin: dict, archive: str, working_directory: str)
     openssl_base64_process = subprocess.Popen(["openssl", "base64", "-A"], stdin=openssl_dgst_process.stdout, stdout=subprocess.PIPE)
 
     output, _ = openssl_base64_process.communicate()
-    print(output.decode('utf-8').strip())
     if hash_digest != output.decode('utf-8').strip():
       raise InstallException(f'{package}: The hash of the downloaded package {output.decode("utf-8").strip()} does not match the provided integrity hash {hash_digest} provided in the configuration file')
 
@@ -150,6 +149,7 @@ def main():
 
     if skipIntegrityCheck:
         print(f"SKIP_INTEGRITY_CHECK has been set to {skipIntegrityCheck}, skipping integrity check of packages")
+
     if 'includes' in content:
         includes = content['includes']
     else:
@@ -220,7 +220,7 @@ def main():
             package = os.path.join(os.getcwd(), package[2:])
         else:
           # If package is not local, then integrity check is mandatory
-          if not skipIntegrityCheck and not plugin['integrity']:
+          if not skipIntegrityCheck and not 'integrity' in plugin:
             raise InstallException(f"No integrity hash provided for Package {package}")
 
         print('\t==> Grabbing package archive through `npm pack`', flush=True)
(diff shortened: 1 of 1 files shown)


docker/install-dynamic-plugins.py
```diff
@@ -230,7 +230,7 @@ def main():
 
         archive = os.path.join(dynamicPluginsRoot, completed.stdout.decode('utf-8').strip())
 
-        if not package_is_local:
+        if not package_is_local and not skipIntegrityCheck:
```
Zaperex, line 233: Isn't this additional check redundant since it's not possible to get past this point without `skipIntegrityCheck` not being `True` when the package isn't local?

docker/install-dynamic-plugins.py
```diff
@@ -220,7 +220,7 @@ def main():
             package = os.path.join(os.getcwd(), package[2:])
         else:
```
Zaperex, line 221:
```suggestion
```

docker/install-dynamic-plugins.py
```diff
@@ -220,7 +220,7 @@ def main():
...
           # If package is not local, then integrity check is mandatory
-          if not skipIntegrityCheck and not plugin['integrity']:
+          if not skipIntegrityCheck and not 'integrity' in plugin:
```
Zaperex, line 223:
```suggestion
        elif not skipIntegrityCheck and not 'integrity' in plugin:
```

davidfestal:
@Zaperex Does the fix make more sense this way ? (I refactored a bit after your proposal review)

docker/install-dynamic-plugins.py
```diff
@@ -216,12 +216,13 @@ def main():
         print('\n======= Installing dynamic plugin', package, flush=True)
 
         package_is_local = package.startswith('./')
+
+        # If package is not local, then integrity check is mandatory
+        if not package_is_local and not skipIntegrityCheck and not 'integrity' in plugin:
```
Zaperex, line 221:
```suggestion
        if not (package_is_local and skipIntegrityCheck and 'integrity' in plugin):
```
sure this method works as well I guess. I personally felt the `if` `elif` chain was more concise and clear.

docker/install-dynamic-plugins.py
```diff
@@ -230,7 +231,7 @@ def main():
 
         archive = os.path.join(dynamicPluginsRoot, completed.stdout.decode('utf-8').strip())
 
-        if not package_is_local:
+        if not (package_is_local or skipIntegrityCheck):
```
Zaperex, line 234: This would cause integrity checks to occur for local packages as well, which I don't think we want?

davidfestal: I don't think so.

If `package_is_local` is `true`, then `package_is_local or skipIntegrityCheck` is always `true` due to the `or`.
So that in this case `not(package_is_local or skipIntegrityCheck)` is `false`.

So as soon as the package is local *`or`* we skip integrity check, we would *not* enter the conditional line.

That's how I understand it and thought it.

Zaperex: ok nevermind then, I think that'll be fine then. I just still have the same argument as [above](https://github.com/janus-idp/backstage-showcase/pull/704#discussion_r1380481117) regarding the redundancy of the `skipIntegrityCheck` check though.

docker/install-dynamic-plugins.py
```diff
@@ -230,7 +231,7 @@ def main():
 
         archive = os.path.join(dynamicPluginsRoot, completed.stdout.decode('utf-8').strip())
 
-        if not package_is_local and not skipIntegrityCheck:
```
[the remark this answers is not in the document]
Zaperex: resolved

Zaperex (approved):

schultzp2020 (approved):
LGTM

PR #710, opened 2023-11-02
fix: disable cache for injected file.

davidfestal:
## Description

Backend merged config is injected by the `app-backend` upstream plugin inside one of the Javascript
static assets before service static assets.
So we should never allow this specific JS file to be cached on the client side,
so that any change in the configuration can be seen on the frontend
side after a backend restart.
That's particularly critical to get the latest UI configuration
of added dynamic frontend plugins.

- Fixes No issue number.

## PR acceptance criteria

### DIFF
--- a/packages/backend/src/plugins/app.ts
+++ b/packages/backend/src/plugins/app.ts
@@ -4,6 +4,7 @@ import type { PluginEnvironment } from '../types';
 import { resolvePackagePath } from '@backstage/backend-common';
 import { resolve as resolvePath } from 'path';
 import fs from 'fs-extra';
+import rateLimit from 'express-rate-limit';
 
 export default async function createPlugin(
   env: PluginEnvironment,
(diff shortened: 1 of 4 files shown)


tumido (approved):
/lgtm
/approve

packages/backend/src/plugins/app.ts
```diff
@@ -1,14 +1,65 @@
...
+  const jsFiles = files.filter(file => file.endsWith('.js'));
+  let injectedJSFile: string | undefined = undefined;
+
+  for (const jsFile of jsFiles) {
+    const path = resolvePath(staticDir, jsFile);
+
+    const content = await fs.readFile(path, 'utf8');
+    if (content.includes('__APP_INJECTED_')) {
+      injectedJSFile = jsFile;
+      break;
```
Hyperkid123, line 27: Is it 100% confirmed that there will only ever be one file with the mark? I don't think its that much extra work to assume there can be more than one file that gest mutated at runtime.

packages/backend/src/plugins/app.ts
```diff
@@ -1,14 +1,65 @@
...
   });
+
+  const enclosingRouter = Router();
+  if (injectedJSFile) {
+    env.logger.info(
+      `Setting up static router for injected Javascript file ${injectedJSFile}`,
+    );
+
+    enclosingRouter.get(
+      `/static/${injectedJSFile}`,
```
Hyperkid123, line 45: We can iterate over the array to set up multiple routes with this rule

packages/backend/src/plugins/app.ts
davidfestal: > Is it 100% confirmed that there will only ever be one file with the mark

It is how the `backend-app` plugins works. See the `return` lines in https://github.com/backstage/backstage/blob/master/plugins/app-backend/src/lib/config.ts#L55 and https://github.com/backstage/backstage/blob/master/plugins/app-backend/src/lib/config.ts#L64

packages/backend/src/plugins/app.ts
davidfestal: Is it really useful based on my answer [above](https://github.com/janus-idp/backstage-showcase/pull/710#discussion_r1383255152) ?

packages/backend/src/plugins/app.ts
Hyperkid123: OK. Well. Then we are good.

packages/backend/src/plugins/app.ts
Hyperkid123: Nope

Hyperkid123 (approved):

PR #711, opened 2023-11-03
fix(dynamic-plugins): handle plugin initialization failures

Hyperkid123:
## Description

Add extra error handling while initializing dynamic front-end plugins. If a plugin fails to initialize, it won't be registered and won't be rendered at the expected place.


tumido (approved):
/lgtm
/approve

tumido (approved):
/lgtm
/approve

PR #721, opened 2023-11-06
chore: don't run scripts on dynamic plugin install

davidfestal:
Don't run package scripts (`prepack`, `postpack` foe example) when using `npm pack` at runtime to install the dynamic plugins.

In fact that's also quite expected that no arbitrary script would be run while installing downloaded dynamic plugins.


tumido (approved):
/lgtm
/approve

schultzp2020 (approved):
LGTM

Issue #715, opened 2023-11-03
bad ux: Plugin-in needs to be enabled in three places


kadel:
### What do you want to improve?

Currently If I want to enable plugin I have to change it in 
app-config.yaml for example `enabled.githubOrg: true` and then I have to also enable it in `dynamic-plugins.default.yaml`, there it looks like we have also two options to enable/disable plugin
```
    disabled: false
    pluginConfig:
      enabled:
        githubOrg: true
````

**Comment by kadel** [comment]
it should be controlled from one place and look like this: 

```
    disabled: false
    pluginConfig:
       . . . . .
```

PR #740, opened 2023-11-08
Remove legacy `enabled.xxx` option

schultzp2020:
## Description

Removes legacy `enabled.xxx` option

- Fixes https://github.com/janus-idp/backstage-showcase/issues/715

## PR acceptance criteria


davidfestal (commented):
Didn't review the whole, but You should have some cleaning to do as well in the janus-idp wrappers afair

packages/backend/src/index.ts
```diff
@@ -84,86 +84,28 @@ function makeCreateEnv(config: Config, pluginProvider: BackendPluginProvider) {
...
-} & AddRouterBase;
-
-async function addRouter(args: AddRouter | AddRouterOptional): Promise<void> {
-  const { isOptional, name, service, root, router } = args;
-
-  const isRouterEnabled =
-    !isOptional || args.config.getOptionalBoolean(`enabled.${name}`) || false;
-
-  if (isRouterEnabled) {
-    console.log(`Adding router ${name} to backend...`);
```
tumido, line 164: Not a request, just a thought... I actually liked this log line. 😄

tumido (approved):
No complaints. LGTM

packages/backend/src/index.ts
```diff
@@ -270,34 +204,28 @@ async function main() {
...
   await addRouter({
-    name: 'healthcheck',
     service,
     root: '',
     router: await createStatusCheckRouter(appEnv),
   });
-
-  // Optional routers
   await addRouter({
-    name: 'metrics',
```
davidfestal, line 289: We might want to find another way to enable / disable it (since it's not a dynamic plugin)

app-config.yaml
```diff
@@ -176,9 +176,5 @@ dynatrace:
 permission:
   enabled: ${PERMISSION_ENABLED}
 
-enabled:
```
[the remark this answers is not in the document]
davidfestal: Well, for permission I don't known since it doesn't seem to even be optional in the current `index.ts` file.

app-config.deprecated.yaml
davidfestal, line 1: Is it required to completely remove this file, instead of removing only the `enabled.xx` properties ?

packages/backend/src/index.ts
schultzp2020: I can add it back

app-config.deprecated.yaml
schultzp2020: It's not necessary anymore since it's missing so many values.

packages/backend/src/index.ts
schultzp2020: I think we should always leave the `metrics` and `healthcheck` router on. Thoughts? @Zaperex @kadel @tumido?

app-config.yaml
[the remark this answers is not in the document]
schultzp2020: https://github.com/janus-idp/backstage-showcase/pull/740#discussion_r1388607081

packages/backend/src/index.ts
Zaperex: I think it'll make sense to leave them on. Also, when I tested it with the current showcase main branch, it seems the `metrics` and `healthcheck` routers are always on regardless of the `enabled.xxx` values for each. So I think it's safe to remove the documentation for them.

schultzp2020:
@tumido @davidfestal PTAL?

dynamic-plugins/wrappers/backstage-plugin-catalog-backend-module-github-dynamic/src/dynamic/index.ts
```diff
@@ -4,20 +4,16 @@ import { GithubEntityProvider } from '@backstage/plugin-catalog-backend-module-g
...
+        schedule: env.scheduler.createScheduledTaskRunner({
+          frequency: { minutes: 30 },
+          timeout: { minutes: 3 },
+          initialDelay: { minutes: 1 },
         }),
-      );
-    } else {
-      env.logger.info('Github plugin is disabled');
-    }
+        scheduler: env.scheduler,
```
Zaperex, line 15: I think the upstream github plugin has the same behavior where the [hard coded `schedule` field takes precedence over the `scheduler`](https://github.com/backstage/backstage/blob/438f7f87ac61095ae700e085a375c41cf16f5c58/plugins/catalog-backend-module-github/src/providers/GithubEntityProvider.ts#L112-L114). This would require a change in the upstream.

dynamic-plugins/wrappers/backstage-plugin-catalog-backend-module-gitlab-dynamic/src/dynamic/index.ts
```diff
@@ -4,19 +4,15 @@ import { GitlabDiscoveryEntityProvider } from '@backstage/plugin-catalog-backend
...
+        logger: env.logger,
+        schedule: env.scheduler.createScheduledTaskRunner({
+          frequency: { minutes: 30 },
+          timeout: { minutes: 3 },
         }),
-      );
-    } else {
-      env.logger.info('Gitlab plugin is disabled');
-    }
+        scheduler: env.scheduler,
```
Zaperex, line 14: Similarly to the github provider above, [gitlab](https://github.com/backstage/backstage/blob/438f7f87ac61095ae700e085a375c41cf16f5c58/plugins/catalog-backend-module-gitlab/src/providers/GitlabDiscoveryEntityProvider.ts#L82-L84) also has this issue upstream

packages/backend/src/index.ts
tumido: I would expect `/metrics` and `/healthcheck` endpoints to be always available. If we want the app to be a good citizen of a greater ecosystem, this should be a no-brainer. 🙂

tumido (changes requested):
I agree with @Zaperex on the `schedule` properties. If the plugins prefer `schedule` over `scheduler` it should not be included in our dynamic plugin wrappers.

packages/backend/src/index.ts
davidfestal: Perfectly OK to have them always on !

davidfestal:
> I agree with @Zaperex on the `schedule` properties. If the plugins prefer `schedule` over `scheduler` it should not be included in our dynamic plugin wrappers.

I agree as well

schultzp2020:
@tumido @davidfestal @Zaperex fixed

tumido (approved):
/lgtm
/approve

davidfestal (approved):

schultzp2020:
@tumido @davidfestal I will create a PR on the plugins to remove the hard coded schedule. e.g. https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/src/dynamic/index.ts#L12-L16

Issue #569, opened 2023-09-27
Make tab icon configurable


Zaperex:
### Provide a clear and concise description of the feature request
Allow users to use the `app.branding.iconLogo` app configurations to customize the icon for the browser tab
### Describe the problem that this feature resolves
Currently, the icon for the browser tab is hard coded and cannot be configured.
### Additional context or screenshots for the feature request
![image](https://github.com/janus-idp/backstage-showcase/assets/50030060/5022c5b3-f1a3-4f07-a921-8c928b5edb1b)

### Acceptance Criteria
Ability to customize the browser tab icon via the `app-config.yaml`

PR #742, opened 2023-11-08
fix: correctly use dynamic plugin config schemas

### LINKED ISSUE #724: Config schema aggregation is missing for dynamic plugins
### Describe the bug

Dynamic plugins do not aggregate config schemas, therefore are missing `visibility` annotation in from resulting config. This has 2 side effects:

- on backend secrets may not be redacted from logs
- on frontend we have no access to config values that is directly read from config

### Expected Behavior

Parity with static plugins

### What are the steps to reproduce this bug?

Use a dynamic plugin like `backstage.plugin-lighthouse` and see it fails to access the config data

### Versions of software used and environment

Task:
- [ ] Expose config from plugins: https://github.com/janus-idp/backstage-showcase/issues/724
- [ ] Aggregate config to the instance

davidfestal:
## Description

Correctly use the dynamic plugins configuration schemasin the Showcase application

- Fixes #724

## PR acceptance criteria

### DIFF
--- a/packages/backend/src/plugins/app.ts
+++ b/packages/backend/src/plugins/app.ts
@@ -5,14 +5,20 @@ import { resolvePackagePath } from '@backstage/backend-common';
 import { resolve as resolvePath } from 'path';
 import fs from 'fs-extra';
 import rateLimit from 'express-rate-limit';
+import { completeFrontendSchemas } from '../schemas';
 
 export default async function createPlugin(
   env: PluginEnvironment,
+  dynamicPluginsSchemas: { value: any; path: string }[],
 ): Promise<Router> {
   const appPackageName = 'app';
 
   const appDistDir = resolvePackagePath(appPackageName, 'dist');
   const staticDir = resolvePath(appDistDir, 'static');
+
+  if (await fs.pathExists(appDistDir)) {
+    await completeFrontendSchemas(dynamicPluginsSchemas, appDistDir);
+  }
 
   let injectedJSFile: string | undefined = undefined;
 
--- a/.changeset/dull-hotels-dream.md
+++ b/.changeset/dull-hotels-dream.md
@@ -1,0 +1,5 @@
+---
+'backend': patch
+---
+
+Correctly use the dynamic plugins configuration schemas in the Showcase application
--- a/dynamic-plugins/wrappers/backstage-plugin-azure-devops-backend-dynamic/package.json
+++ b/dynamic-plugins/wrappers/backstage-plugin-azure-devops-backend-dynamic/package.json
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "@backstage/cli": "0.23.1",
-    "@janus-idp/cli": "1.3.3"
+    "@janus-idp/cli": "1.4.0"
   },
   "files": [
     "dist",
--- a/dynamic-plugins/wrappers/backstage-plugin-azure-devops/package.json
+++ b/dynamic-plugins/wrappers/backstage-plugin-azure-devops/package.json
@@ -24,7 +24,7 @@
   },
   "devDependencies": {
     "@backstage/cli": "0.23.1",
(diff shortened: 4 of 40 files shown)


dynamic-plugins/imports/package.json
```diff
@@ -18,7 +18,7 @@
     "@janus-idp/backstage-plugin-acr": "1.2.4",
     "@janus-idp/backstage-plugin-jfrog-artifactory": "1.2.4",
     "@janus-idp/backstage-plugin-nexus-repository-manager": "1.4.4",
-    "@janus-idp/backstage-plugin-ocm": "3.3.4",
+    "@janus-idp/backstage-plugin-ocm": "3.5.0",
```
tumido, line 21: Please let this be updated by #569

This version update is irrelevant to the PR merit.

dynamic-plugins/quay.io plugin.txt
tumido, line 1: What is this file about? Should it be here?

davidfestal: Pushed by error. I'll remove it.

davidfestal: Done.

schultzp2020 (approved):
LGTM

tumido (changes requested):
And one more: please bump as well for consistent (deduped) resolution:

https://github.com/janus-idp/backstage-showcase/blob/183f894361d648785bfbc83f51a3d4537c7d5b73/packages/app/package.json#L62

tumido:
And one last request. Running `yarn export-dynamic` on this PR produces diff in `dynamic-plugins/wrappers/*/dist-dynamic/package.json` for me:

```diff
-    "test": "backstage-cli package test --coverage",
+    "test": "backstage-cli package test --passWithNoTests --coverage",
```

can you please fix that?

packages/backend/src/plugins/app.ts
```diff
@@ -5,15 +5,21 @@ import { resolvePackagePath } from '@backstage/backend-common';
...
   env: PluginEnvironment,
+  dynamicPluginsSchemas: { value: any; path: string }[],
 ): Promise<Router> {
   const appPackageName = 'app';
 
   const appDistDir = resolvePackagePath(appPackageName, 'dist');
   const staticDir = resolvePath(appDistDir, 'static');
 
+  if (await fs.pathExists(appDistDir)) {
+    await completeFrontendSchemas(dynamicPluginsSchemas, appDistDir);
```
tumido, line 20: Does this mean if the `dist` doesn't exist we don't get schemas?

`yarn start` in dev mode doesn't create `package/app/dist`, so the default experience now means we do not aggregate schema properly.

Once I create the `package/app/dist` I get those schemas generated appropriately but they still do not get applied. 

What else do I have to do to make this work with `yarn start` locally?

tumido (changes requested):

packages/backend/src/plugins/app.ts
davidfestal: I assume that in the local dev mode, some changes should be done in the janus cli start command, to make it so that the webpack Dev server can also get the frontend configs related to dynamic plugins. 

That seems to be a completely distinct area and use-case, and maybe we could keep that use-case for a distinct issue ?

davidfestal:
@tumido 

> And one more: please bump as well for consistent (deduped) resolution:
> 
> https://github.com/janus-idp/backstage-showcase/blob/183f894361d648785bfbc83f51a3d4537c7d5b73/packages/app/package.json#L62

done

> bump to 1.4.0 can you please update it in the docs as well?

done

packages/backend/src/plugins/app.ts
davidfestal: I created a new issue for the Dev Server case: https://github.com/janus-idp/backstage-showcase/issues/763

tumido (approved):
/lgtm
/approve

PR #766, opened 2023-11-13
fix: increase max entry size for dynamic plugins

davidfestal:
## Description of the change

Increase max entry size for dynamic plugins in order to avoid triggering a zip-bomb error during installation for plugins delivered as part of the showcase docker image.

The previous value (10000000) was not enough for 2 of the plugins delivered as part of the showcase docker image (tekton plugin and the topology plugin)
## Checklist

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Variables are documented in the `values.yaml` and added to the README.md. The [pre-commit](https://pre-commit.com/) utility can be used to generate the necessary content. Use `pre-commit run -a` to apply changes.
- [ ] JSON Schema template updated and re-generated the raw schema via `pre-commit` hook.
- [ ] List tests pass for Chart using the [Chart Testing](https://github.com/helm/chart-testing) tool and the `ct lint` command.

### DIFF
--- a/docker/install-dynamic-plugins.py
+++ b/docker/install-dynamic-plugins.py
@@ -29,7 +29,7 @@ import binascii
 # the dynamic plugins will be installed.
 #
 # Additionally, the MAX_ENTRY_SIZE environment variable can be defined to set
-# the maximum size of a file in the archive (default: 10MB).
+# the maximum size of a file in the archive (default: 20MB).
 #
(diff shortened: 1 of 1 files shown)


nickboldt (approved):

Zaperex (approved):
lgtm

PR #768, opened 2023-11-13
fix: use the 'tar' filter to remove warnings

davidfestal:
## Description

Since recently on both docker images, which have been updated, the following warning is displayed in the initContainer logs during dynamic plugins installation:
```
==> Extracting package archive /dynamic-plugins-root/janus-idp-backstage-plugin-topology-1.16.4.tgz
/usr/lib64/python3.9/tarfile.py:2239: RuntimeWarning: The default behavior of tarfile extraction has been changed to disallow common exploits (including CVE-2007-4559). By default, absolute/parent paths are disallowed and some mode bits are cleared. See https://access.redhat.com/articles/7004769 for more details.
```

This is because a new `filter` option option has been added in the python `tar` library, whose default value changes the default behavior of the library.

Explicitely setting this filter option to `tar` (which is the new default) should remove the warning from the logs.

No issue

## PR acceptance criteria

### DIFF
--- a/docker/install-dynamic-plugins.py
+++ b/docker/install-dynamic-plugins.py
@@ -268,7 +268,7 @@ def main():
                 if not realpath.startswith(directoryRealpath):
                   raise InstallException('NPM package archive contains a link outside of the archive: ' + member.name + ' -> ' + member.linkpath)
 
-                file.extract(member, path=directory)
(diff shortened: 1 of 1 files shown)


nickboldt (approved):
sure

nickboldt:
cherrypicked as b2b8824

PR #791, opened 2023-11-16
fix(dynamic-frontend): allow users to pass custom api implementations and switch tech radar to dynamic

### LINKED ISSUE #741: Plugin loading method inconsistency between frontend and backend for the techdocs
### Describe the bug

The `backstage-techdocs` frontend plugins/modules [have not been switched to dynamic](https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/package.json#L38), though the corresponding backend plugin has been (and is even disabled by default).
It's an inconsistency that will have to be fixed, probably by switching the `techdocs` frontend plugin/module to dynamic as well.

Side note: The `backstage-tech-radar` also is also still statically included in the Showcase? Is it on purpose ?

tumido:
## Description

Migrates Tech Radar to a dynamic plugin and exposes [Utility APIs override option](https://backstage.io/docs/api/utility-apis/#app-apis) from `createApp`

This one was tricky, since it relies on a custom Utility API implementation. We didn't allow dynamic Utility API overrides/implementations so this PR corrects that also (commit 1).

Commit 2 is the actual Tech Radar migration - I had to split the `CustomDataAPI` and remove the tech-radar content from it. Instead I've created a new `TechRadarApi` that is dynamically loaded from the plugin.

Together with #775 resolves #741

## PR acceptance criteria


schultzp2020 (approved):
LGTM

gashcrumb (approved):
:+1: worked for me

PR #820, opened 2023-11-28
fix(servicenow): upgrade servicenow plugin version

davidfestal:
This PR upgrades the ServiceNow action plugin to the version that fixes issue https://github.com/janus-idp/backstage-plugins/issues/976


schultzp2020 (approved):
LGTM

PR #885, opened 2024-01-17
feat(notifications): introduce Notifications plugins to the showcase

mareklibra:
## Description

Adding the Notifications plugins to the show-case app.

- https://github.com/janus-idp/backstage-plugins/tree/main/plugins/notifications
- https://github.com/janus-idp/backstage-plugins/tree/main/plugins/notifications-backend

Screenshot from 2024-01-17 17-20-24

## TODO:
- [x] Have https://github.com/janus-idp/backstage-plugins/pull/1077 merged
- [x] rename dependencies from private NPM `@marek.libra/*` to `@janus-idp/*` once CI does its `npm publish`

## PR acceptance criteria

### DIFF
--- a/app-config.example.yaml
+++ b/app-config.example.yaml
@@ -43,7 +43,7 @@ catalog:
     entityFilename: catalog-info.yaml
     pullRequestBranchName: backstage-integration
   rules:
-    - allow: [Component, System, Group, Resource, Location, Template, API]
+    - allow: [Component, System, Group, Resource, Location, Template, API, User]
   locations:
     # Note: integrations.github[].apps must be correctly configured to read GitHub locations
     - type: file
--- a/catalog-entities/all.yaml
+++ b/catalog-entities/all.yaml
@@ -10,4 +10,5 @@ spec:
     - ./groups.yaml
     - ./systems.yaml
     - ./resources.yaml
+    - ./users.yaml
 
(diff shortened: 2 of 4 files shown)


davidfestal (changes requested):
Do we really want to add this as a static plugin ?

cc @kadel

mareklibra:
> Do we really want to add this as a static plugin

There is upstream work on the same feature in progress, we would ideally share (or build upon) what we have in Janus.
If we convert the plugin to a dynamic one here, we will deviate from the upstream even more.

Once the upstream Notifications land, it will be part of the core Backstage. And Janus will eventually rebase on top of the newer Backstage version.

kadel:
> Do we really want to add this as a static plugin ?
> 
> cc @kadel

Not at this point. Especially not as a static plugin. We don't know what will happen in the upstream. 

@mareklibra Is there a reason why this couldn't be a dynamic plugin?

davidfestal:
> > Do we really want to add this as a static plugin ?
> > cc @kadel
> 
> Not at this point. Especially not as a static plugin. We don't know what will happen in the upstream.
> 
> @mareklibra Is there a reason why this couldn't be a dynamic plugin?

Making it dynamic doesn't prevent it to be used as a static plugin, so wouldn't prevent it to be contributed to upstream at all.

pkliczewski:
> Making it dynamic doesn't prevent it to be used as a static plugin, so wouldn't prevent it to be contributed to upstream at all.

@mareklibra if dynamic plugin could be used as static plugin I think it is safe to change it.

mareklibra:
Closing the PR. The plugins are turned into dynamic ones.
I will file new PRs if further changes are needed.

PR #894, opened 2024-01-22
chore(scalprum-backend): Frontend plugins log line

### LINKED ISSUE #893: Add logging statements to scalprum-backend to log dynamic frontend plugins as well.
### What do you want to improve?

It'd be good if the frontend plugnis loaded were also logged just like the backend plugins, so that a user doesn't have to look in two places.

### What is the current behavior?

Currently loaded backend plugins are logged to stdout, however frontend plugins are not, and the way to see that is by using the rest API once the app is initialized.

### What is the new behavior?

We could have the scalprum plugin log what it discovers as well, which could look something like the second block of logging output in this screenshot:

image

@davidfestal do you think this would be worthwhile in the short term?

gashcrumb:
## Description

This commit adds a log statement to the scalprum-backend plugin to log each successfully loaded frontend dynamic plugin, using a similar format to what's current used to log backend dynamic plugins.  Here's what it would look like on the console:

image

- Fixes #893

## PR acceptance criteria

### DIFF
--- a/.changeset/swift-avocados-jog.md
+++ b/.changeset/swift-avocados-jog.md
@@ -1,0 +1,5 @@
+---
+'@internal/plugin-scalprum-backend': patch
+---
+
+Add logging statements when loading frontend dynamic plugins similar to the logging statements when loading backend dynamic plugins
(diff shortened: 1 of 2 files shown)


Zaperex (approved):
/lgtm

Zaperex:
@gashcrumb can you add a changeset?

Issue #912, opened 2024-01-25
The showcase doesn't start locally when running `yarn start`


Zaperex:
### Describe the bug
When using `yarn start` in the root directory to start the backstage application, the frontend fails with the following error:
![image](https://github.com/janus-idp/backstage-showcase/assets/50030060/4b43a780-241f-4466-8b97-e88de270a97f)

When testing doing `yarn start` from commit https://github.com/janus-idp/backstage-showcase/commit/4a6266e16f2548071b740e9f8d3bc4b6fd08b234 (before the 1.22 upgrade)
The logs produce the following and backstage starts up properly:
```bash
backend:start: cache bypass, force executing f3913656802e2f5f
@internal/plugin-dynamic-plugins-info-backend:start: cache bypass, force executing fd38329a4c7ee316
@internal/plugin-scalprum-backend:start: cache bypass, force executing 87683b206ad30f9b
app:start: cache bypass, force executing 6c746a5d57ae5faf
backend:start: $ cross-env LEGACY_BACKEND_START=true backstage-cli package start
@internal/plugin-scalprum-backend:start: $ backstage-cli package start
@internal/plugin-dynamic-plugins-info-backend:start: $ backstage-cli package start
app:start: $ janus-cli package start
@internal/plugin-scalprum-backend:start: (node:211262) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
@internal/plugin-scalprum-backend:start: (Use `node --trace-warnings ...` to show where the warning was created)
@internal/plugin-dynamic-plugins-info-backend:start: (node:211263) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
@internal/plugin-dynamic-plugins-info-backend:start: (Use `node --trace-warnings ...` to show where the warning was created)
backend:start: Build succeeded
```
The logs start returning the following when switching to the backstage 1.22.1 commit https://github.com/janus-idp/backstage-showcase/commit/8f9811cc303b40fbb98ab9f064b04bac63674b22, but the showcase still starts properly.
```bash
app:start: cache bypass, force executing f6333405e07da288
@internal/plugin-dynamic-plugins-info-backend:start: cache bypass, force executing 94c6b8fd02e56c01
@internal/plugin-scalprum-backend:start: cache bypass, force executing 6d2932d82f85349e
backend:start: cache bypass, force executing 243cea59deea3070
[...]

PR #925, opened 2024-01-26
chore(deps): update scalprum dependecies versions

Hyperkid123:
## Description

This is an issue with yarn and how its dependency deduplication works. The problems are exacerbated in mono repositories.

Why was this an issue?

The scallprum test utils package has the following scalprum core version requirement:
```
"scalprum/core": "^0.6.*"
```
Ant version `0.*.*` is compatible with the packages. However by locking the `@scalprum/core` dependency to `0.7.0` will result in the following dependency tree in the package:

```
yarn list --pattern @scalprum/core
yarn list v1.22.19
├─ @scalprum/core@0.6.6
└─ app@1.0.1
   └─ @scalprum/core@0.7.0
```
I've released versions that strictly require the latest versions of Scalprum. However, this is not an ideal solution. We might have to add yarn resolutions to ensure only actually installed packages is being used. OR switch to npm.

More information about the deduplication issues for nested dependencies can be found here: https://github.com/yarnpkg/yarn/issues/3778


gashcrumb (approved):

Hyperkid123:
@gashcrumb any idea why is the OCP/PR Test failing? I don't have access to the IBM cloud.

gashcrumb:
Yep, lemme check, though I think at the moment it's expected that they'll fail.  I tested this locally and see no problem.

nickboldt (commented):
This appears to be the opposite of https://github.com/janus-idp/backstage-showcase/pull/915 https://github.com/janus-idp/backstage-showcase/pull/912

but with an upgrade to "@scalprum/react-test-utils": "0.1.3"

@schultzp2020 @Zaperex

Zaperex:
I just tested it locally, and it seems to work now. Not sure why the tests e2e-tests are failing at the moment as it seems like it wasn't able to access the UI?

schultzp2020 (approved):
I was told by QE that we can ignore the e2e tests for now. They don't have it fully setup. 

/lgtm

PR #927, opened 2024-01-26
feat(app): add build details in settings page

schultzp2020:
## Description

image

- Fixes https://issues.redhat.com/browse/RHIDP-487

## PR acceptance criteria


packages/app/src/rhdh.json
```diff
@@ -0,0 +1,4 @@
+{
+  "rhdhVersion": "1.0.0",
```
nickboldt, line 2:
```suggestion
  "rhdhVersion": "1.1.0",
```

nickboldt: could this be inherited from the root package.json? https://github.com/janus-idp/backstage-showcase/blob/main/package.json#L3

packages/app/src/rhdh.json
```diff
@@ -0,0 +1,4 @@
...
+  "backstageVersion": "1.22.1"
```
nickboldt, line 3: could this be inherited from https://github.com/janus-idp/backstage-showcase/blob/main/backstage.json#L2 so we don't need a new config file with the same information?

packages/app/src/rhdh.json
schultzp2020: No because it lives outside of the `packages/app/src` scope.

packages/app/src/rhdh.json
schultzp2020: https://github.com/janus-idp/backstage-showcase/pull/927#discussion_r1468085932

packages/app/src/rhdh.json
nickboldt: ok, could this file be generated from package.json then? like `yarn build` just copies one line from the root package.json into this file?

also if the file is called rhdh.json and the field is called rhdhversion, then it's not really friendly for reskinning by a downstream consumer

maybe we want product.json and just `version` ?

packages/app/src/rhdh.json
nickboldt: can we generate this at build time then? https://github.com/janus-idp/backstage-showcase/pull/927#discussion_r1468199818 ? 

do you plan to also collect a build timestamp or other metadata fields? or would this be something I overrride downstream so we can have "1.1.0-CI-timestamp" instead of just "1.1.0" ?

nickboldt (approved):

nickboldt:
weird. this doesn't appear downstream yet...

image

@schultzp2020  Any idea why?

nickboldt:
image

from registry.redhat.io/rhdh/rhdh-hub-rhel9@sha256:569dc8ca02f4c46d21b2c7bcb1cd54ccc2578b51486f3009814654689d54aa77 via mirror to quay.io/rhdh/rhdh-hub-rhel9:1.1-66

PR #939, opened 2024-01-30
chore(changesets): remove changesets

schultzp2020:
## Description

We should remove the changesets from the showcase because we have not updated the changelogs since dev preview. 

- Fixes https://issues.redhat.com/browse/RHIDP-1129

## PR acceptance criteria


nickboldt:
Not to be too pedantic, but you're not removing the changesets...

they're still in https://github.com/janus-idp/backstage-showcase/tree/main/.changeset

you're just no longer checking PRs for new ones, and removing the yarn commands to add new ones. 

FWIW this may be a different path to what's suggested for migrated plugins from backstage-plugins -> community-plugins as that team likes changesets (but we might convince them to do releases on every commit using semantic release instead or in addition to changeset-grouped releases).

kadel:
Wouldn't it be better if we start using this properly instead of removing it? Last release we had a hard time coming up with the changelog, if we start using changesets properly it will make releases easier.

kadel:
>Wouldn't it be better if we start using this properly instead of removing it? Last release we had a hard time coming up with the changelog, if we start using changesets properly it will make releases easier.

we discussed that and come to the conclusion that we will remove changes fro now

nickboldt (approved):

PR #957, opened 2024-02-08
fix(ui): add pf5 button style config

ciiay:
## Description

A theme config for a Material-UI Button (v4) to look as similar as possible to a PF5 Button.

## Which issue(s) does this PR fix

- Fixes [RHIDP-1163](https://issues.redhat.com/browse/RHIDP-1163): Add MUI theme config so that Buttons looks similar to PF5

## PR acceptance criteria

Please make sure that the following steps are complete:

- [x] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [x] Add a screenshot if the change is UX/UI related

## How to test changes / Special notes to the reviewer
**Light theme buttons**:
light_buttons

Contained_button
light_contained_button_text_button
Contained_button_hover
light_contained_button_hover
Contained_button_disabled
light_contained_button_disabled
Outlined_button
light_outlined_button
Outlined_button_hover
light_outlined_button_hover
Text_button
light_text_button
Text_button_hover
light_text_button_hover

**Dark theme buttons**:
Contained button and text button:

https://github.com/janus-idp/backstage-showcase/assets/26255262/308cff15-0036-4def-95b7-e82c4265c2fb

Contained_button_disabled
dark_contained_button_disabled
Outline_button
dark_outline_button
Outline_button_hover
dark_outline_button_hover

Dark theme palette reference:
https://www.patternfly.org/developer-resources/dark-theme-handbook/
https://github.com/patternfly/patternfly/blob/main/src/patternfly/sass-utilities/themes/dark/scss-variables.scss
https://github.com/patternfly/patternfly/blob/main/src/patternfly/sass-utilities/themes/dark/colors.scss

PR #960, opened 2024-02-09
Add MUI theme config to fix Sidebar and Tabs color

debsmita1:

- Fixes  https://issues.redhat.com/browse/RHIDP-1165 ,  https://issues.redhat.com/browse/RHIDP-1164

Selected Tab (dark theme)
Screenshot 2024-02-09 at 11 24 37 PM

Tab when hovered (dark theme)
Screenshot 2024-02-09 at 11 24 24 PM

Selected Tab (light theme)
Screenshot 2024-02-09 at 11 25 31 PM

Tab when hovered (light theme)
Screenshot 2024-02-09 at 11 25 38 PM

Nav when hovered (light theme)
Screenshot 2024-02-09 at 11 27 41 PM

Nav when hovered (dark theme)
Screenshot 2024-02-09 at 11 29 17 PM

### DIFF
--- a/packages/app/src/themes/componentOverrides.ts
+++ b/packages/app/src/themes/componentOverrides.ts
@@ -1,4 +1,5 @@
 import { UnifiedThemeOptions } from '@backstage/theme';
+import { ThemeColors } from '../types/types';
 
 const redhatFont = `@font-face {
   font-family: 'Red Hat Font';
@@ -10,8 +11,54 @@ const redhatFont = `@font-face {
     url(/fonts/RedHatText-Regular.ttf) format('truetype');
 }`;
 
-export const components: UnifiedThemeOptions['components'] = {
-  MuiCssBaseline: {
-    styleOverrides: redhatFont,
-  },
+export const components = (
+  themeColors: ThemeColors,
+): UnifiedThemeOptions['components'] => {
+  return {
+    BackstageHeaderTabs: {
+      styleOverrides: {
+        tabsWrapper: {
+          paddingLeft: '0px',
+        },
+        defaultTab: {
+          textTransform: 'none',
+          fontSize: '0.875rem',
+        },
+      },
+    },
+    MuiTabs: {
+      defaultProps: {
+        TabIndicatorProps: {
+          style: {
+            height: '3px',
+            background: themeColors.navigationIndicatorColor || '#0066CC',
+          },
+        },
+      },
+      styleOverrides: {
+        root: {
+          borderBottom: '1px solid #d2d2d2',
+        },
+      },
+    },
+    MuiTab: {
+      defaultProps: {
+        disableRipple: true,
+      },
+      styleOverrides: {
+        root: {
+          minWidth: 'initial !important',
+          '&:hover': {
+            boxShadow: '0 -3px #b8bbbe inset',
+          },
+        },
+        disabled: {
+          backgroundColor: '#6a6e73',
+        },
+      },
+    },
+    MuiCssBaseline: {
+      styleOverrides: redhatFont,
+    },
+  };
 };
--- a/.changeset/sweet-colts-exercise.md
+++ b/.changeset/sweet-colts-exercise.md
@@ -1,0 +1,5 @@
+---
+'app': minor
+---
+
+updated Nav and tab styles to match PatternFly5.
--- a/packages/app/src/components/Root/Root.tsx
+++ b/packages/app/src/components/Root/Root.tsx
@@ -29,9 +29,6 @@ import { IconComponent, useApp } from '@backstage/core-plugin-api';
 const useStyles = makeStyles()({
   sidebarItem: {
     textDecorationLine: 'none',
-    '&:hover': {
-      textDecorationLine: 'underline',
-    },
   },
 });
 
--- a/packages/app/src/themes/darkTheme.ts
+++ b/packages/app/src/themes/darkTheme.ts
@@ -15,15 +15,15 @@ export const customDarkTheme = (themeColors: ThemeColors) =>
       }),
       navigation: {
         background: '#0f1214',
-        indicator: themeColors.navigationIndicatorColor || '#009596',
+        indicator: themeColors.navigationIndicatorColor || '#0066CC',
         color: '#ffffff',
         selectedColor: '#ffffff',
         navItem: {
-          hoverBackground: '#030303',
+          hoverBackground: '#3c3f42',
         },
       },
     },
     defaultPageTheme: 'home',
     pageTheme: pageTheme(themeColors),
-    components,
+    components: components(themeColors),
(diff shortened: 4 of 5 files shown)


packages/app/src/themes/componentOverrides.ts
```diff
@@ -11,6 +11,43 @@ const redhatFont = `@font-face {
...
+  MuiTab: {
+    defaultProps: {
+      disableRipple: true,
+    },
+    styleOverrides: {
+      root: {
+        minWidth: 'auto',
+      },
+      disabled: {
+        backgroundColor: 'gray',
```
invincibleJai, line 38: should we use hex code?

debsmita1:
> Thank you @debsmita1!, I have few comments
> 
> 1. **Tabs** (I compared the screenshots with the design [here](https://www.patternfly.org/components/tabs))
>    I think we don't display the gray background color when hovering over the tab or when it's selected

The tabs component that you see being used in plugins is a Backstage’s HeaderTab component  and the hover color is the same as the background color of the app.
```
    tabRoot: {
      '&:hover': {
        backgroundColor: theme.palette.background.default,
        color: theme.palette.text.primary,
      },
    },
```

> 2. **Navigation** (I compared the screenshots with the design [here](https://www.patternfly.org/components/navigation))
>    We shouldn't display the underline below the text when hovering

Similarly, the nav component is Backstage's SidebarItem component and there's no prop available that allows removing the underline

> 3. In dark mode, the blue color appears brighter. We use the same main color in both scenarios, both for the tab line color and the blue line on the navigation

Fixed it, using '#0066CC' as mentioned in the slide

Screenshot 2024-02-13 at 12 46 08 AM

Screenshot 2024-02-13 at 12 04 02 PM

ShiranHi:
> The tabs component that you see being used in plugins is a Backstage’s HeaderTab component  and the hover color is the same as the background color of the app.
> 
> >
Can we do the same? so the hover color will be the same to the background?


> Similarly, the nav component is Backstage's SidebarItem component and there's no prop available that allows removing the underline
> 
> >
I am not sure why we have that underline on the navigation when hovering, on the [Backstage demo](https://demo.backstage.io/home) env they don't show any underline when hovering

gashcrumb:
This looks really good!  The only thing that's bugging me is the border line under the tabs, for both light and dark it there's a margin on the left-hand side leaving a space but on the right-hand side of the page there's no space.  I tried capturing a more narrow window to help highlight this:

image

debsmita1:
> This looks really good! The only thing that's bugging me is the border line under the tabs, for both light and dark it there's a margin on the left-hand side leaving a space but on the right-hand side of the page there's no space. I tried capturing a more narrow window to help highlight this:
> 
> image
> 
> Shows up better in dark mode and I think maybe the border color could be too bright in this variant also:
> 
> image

@gashcrumb  The HeaderTab component adds this padding on the left

```
const useStyles = makeStyles(
  theme => ({
    tabsWrapper: {
      gridArea: 'pageSubheader',
      backgroundColor: theme.palette.background.paper,
      paddingLeft: theme.spacing(3), <--------
      minWidth: 0,
    },
    ...
```

and the `createUnifiedTheme` util doesn't expose any option to add spacing

gashcrumb:
> @gashcrumb The HeaderTab component adds this padding on the left
> and the createUnifiedTheme util doesn't expose any option to update the spacing

I don't know if maybe trying to override it using something [like this](https://github.com/gashcrumb/patternfly-backstage-theme/blob/main/src/componentOverrides.ts#L50-L52) could work, but if you like this could be investigated later as a separate thing.

debsmita1:
> > @gashcrumb The HeaderTab component adds this padding on the left
> > and the createUnifiedTheme util doesn't expose any option to update the spacing
> 
> I don't know if maybe trying to override it using something [like this](https://github.com/gashcrumb/patternfly-backstage-theme/blob/main/src/componentOverrides.ts#L50-L52) could work, but if you like this could be investigated later as a separate thing.

This worked! Thank you @gashcrumb !!
Screenshot 2024-02-15 at 11 59 57 AM

ShiranHi:
looks good @debsmita1 ! 🙂

packages/app/src/themes/componentOverrides.ts
```diff
@@ -10,8 +11,50 @@ const redhatFont = `@font-face {
...
+        },
+      },
+    },
+    MuiTab: {
+      defaultProps: {
+        disableRipple: true,
+      },
+      styleOverrides: {
+        root: {
+          minWidth: 'auto',
```
jerolimov, line 46: Unluckily this minWidth style is ignored because there is another style based on another break.

I would recommend this for now:

```suggestion
          minWidth: 'initial !important',
```

Or do you know how to define breakpoints here? I don't.

packages/app/src/themes/componentOverrides.ts
```diff
@@ -10,8 +11,50 @@ const redhatFont = `@font-face {
...
+          '&:hover': {
+            boxShadow: '0 -3px #b8bbbe inset',
+          },
```
jerolimov, line 49: I would also recommend to disable the text transformation that tab content is always UPPERCASE:

```suggestion
          // @ts-ignore: Tab contains a span 'wrapper'
          wrapper: {
            textTransform: 'none',
          },
          '&:hover': {
            boxShadow: '0 -3px #b8bbbe inset',
          },
```

jerolimov (commented):
I have two small improvement ideas.

@invincibleJai I'm also fine to get this in now and improve this later. Feel free to add approval, it's def. the right direction and btw it will create a conflict with #957

/lgtm

invincibleJai:
Thanks @jerolimov 

@debsmita1 PTAL at open comments and would be nice if we could handle those in the same PR. LMK

packages/app/src/themes/componentOverrides.ts
debsmita1: @jerolimov  Tried changing this by overriding the BackstageHeaderTabs textTransform, but the text looks really small

Screenshot 2024-02-17 at 1 15 40 AM
 
 cc @ShiranHi

ShiranHi: > @jerolimov Tried changing this by overriding the BackstageHeaderTabs textTransform, but the text looks really small

Can we change the text size so it will be like PF?
https://www.patternfly.org/components/tabs

debsmita1: Done

Screenshot 2024-02-19 at 11 35 23 AM

jerolimov (approved):
Looks good :+1:

/lgtm
/approve

jerolimov (approved):
/lgtm

kadel:
/lgtm
it was just update with main branch

PR #971, opened 2024-02-15
chore(e2e-tests) Test for dynamic-plugins-info

gashcrumb:
This change adds a E2E test for the dynamic plugins info frontend plugin that validates the plugin mounts at the expected location in the app and checks basic functionality.

For [RHIDP-1005](https://issues.redhat.com/browse/RHIDP-1005)


gashcrumb:
@subhashkhileri this probably needs the dynamic-plugins-info frontend plugin to be enabled in the app for the test to pass.  Where is that set up?

subhashkhileri:
@gashcrumb 

> @subhashkhileri this probably needs the dynamic-plugins-info frontend plugin to be enabled in the app for the test to pass. Where is that set up?

Here is the plugins config:  https://github.com/janus-idp/backstage-showcase/blob/main/.ibm/pipelines/value_files/values_showcase.yaml#L14

gashcrumb:
Awesome thanks!  Updated that, let's see how it goes...

gashcrumb:
This is not running locally now for me this morning after a rebase, looking into that.

gashcrumb:
> This is not running locally now for me this morning after a rebase, looking into that.

Ah, `APP_BASE` is now `BASE_URL`, easy fix :smile:

gashcrumb:
Hmm:

image

PR #982, opened 2024-02-19
draft: Add Notifications as a static plugin

mareklibra:
## Description

Draft for debugging only.

Adds the Notifications plugins as static ones.


mareklibra:
There is no more need for this, closing

PR #1055, opened 2024-03-06
Updated how OAuth headers are processed

sabre1041:
## Description

Ability to define the HTTP header containing the username for the OAuth 2 Proxy Provider

Implements a fallback strategy

1. Environment variable defined by the value of `OAUTH_USER_HEADER`
2. `x-forwarded-preferred-username` HTTP Header (Primary use case: with Keycloak/RHSSO/RHBK)
3. `x-forwarded-user` (Primary use case: OpenShift)

- Fixes #?

## PR acceptance criteria

Make use of a variety of OAuth Providers:

1. Configure OpenShift OAuth based on instructions [here](https://janus-idp.io/blog/2023/02/20/using-openshift-authentication-to-secure-access-to-backstage)
2. Configure Keycloak based on instructions [here](https://janus-idp.io/blog/2023/01/17/enabling-keycloak-authentication-in-backstage)
3. Specify the `OAUTH_USER_HEADER` with a desired header to utilize. You can specify any of the values that are natively provided by the two options previously mentioned

Within the documentation, ignore any details related to customizing the backstage image as the necessary configurations are implemented here

### DIFF
--- a/docs/configuration-env-variables.md
+++ b/docs/configuration-env-variables.md
@@ -1,0 +1,3 @@
+# Environment Variables
+
+- `OAUTH_USER_HEADER` - HTTP header containing the username for the OAuth 2 Proxy Provider. If not set, it tries `x-forwarded-preferred-username` and `x-forwarded-user`
--- a/packages/backend/src/modules/authProvidersModule.ts
+++ b/packages/backend/src/modules/authProvidersModule.ts
@@ -177,7 +177,10 @@ function getAuthProviderFactory(providerId: string): AuthProviderFactory {
       return providers.oauth2Proxy.create({
         signIn: {
           async resolver({ result }, ctx) {
-            const name = result.getHeader('x-forwarded-preferred-username');
+            const name = process.env.OAUTH_USER_HEADER
+              ? result.getHeader(process.env.OAUTH_USER_HEADER)
+              : result.getHeader('x-forwarded-preferred-username') ||
+                result.getHeader('x-forwarded-user');
             if (!name) {
               throw new Error('Request did not contain a user');
             }


kadel:
@sabre1041 Could you please add documentation for this? I know that he current state of the documentation is horrible, I am concerned that if we don't document this somewhere, we may forget that this option exists.

You could, for example, start a new file named `configuration.md` in the `showcase-docs` directory, we can later expand it with other options

sabre1041:
Since there is no documentation.  What about instead starting a new issue that we can build out what we would want in the docs?

This would allow us to capture the material that should be included

kadel:
> Since there is no documentation. What about instead starting a new issue that we can build out what we would want in the docs?
> 
> This would allow us to capture the material that should be included

This is precisely what got us to the point where we are now (zero documentation for most of the 
 things).
We need to make a habit of documenting things from the start. Yes, it sucks to write docs, but otherwise we will never get out of the mess we are in now)

sabre1041:
> > Since there is no documentation. What about instead starting a new issue that we can build out what we would want in the docs?
> > This would allow us to capture the material that should be included
> 
> This is precisely what got us to the point where we are now (zero documentation for most of the things). We need to make a habit of documenting things from the start. Yes, it sucks to write docs, but otherwise we will never get out of the mess we are in now)

Can there be a baseline example to build off of?

sabre1041:
> > > Since there is no documentation. What about instead starting a new issue that we can build out what we would want in the docs?
> > > This would allow us to capture the material that should be included
> > 
> > 
> > This is precisely what got us to the point where we are now (zero documentation for most of the things). We need to make a habit of documenting things from the start. Yes, it sucks to write docs, but otherwise we will never get out of the mess we are in now)
> 
> Can there be a baseline example to build off of?

any update?

sabre1041:
> @sabre1041
> 
> It can be as simple as this
> 
> ```
> # showcase-docs/configuration-env-variables.md
> 
> 
> - `OAUTH_USER_HEADER` - HTTP header containing the username for the OAuth 2 Proxy Provider. If not set, it tries `x-forwarded-preferred-username` and `x-forwarded-user`
> ```

Done

kadel (approved):

kadel:
test are unstable QE is working on fix

/override "Tests on OCP/PR Test"

PR #1154, opened 2024-04-03
 Add a new tab to the EntityPage for the Quarkus console

### LINKED ISSUE #1147: Add new tab EntityPage for the QuarkusConsole page
### Provide a clear and concise description of the feature request

We cannot until now use our QuarkusConsole page

Screenshot 2024-03-28 at 15 58 10

as the `EntityPage` on RHDH/showcase only supports the following [mountPoints](https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/dynamic-plugins.md#using-mount-points) and no tab exists [here](https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/components/catalog/EntityPage/index.tsx#L296) for Quarkus

cmoulliard:
## Description

This PR:
- Add a new tab to the `./packages/app/src/components/catalog/EntityPage/index.tsx` file in order to allow to display the [Quarkus console](https://github.com/q-shift/backstage-plugins)
- Enhance the documentation page to better explain how to run locally the dynamic plugins with `yarn start`
- Add a Table of Contents to the dynamic-plugins.md file

Screenshot 2024-04-03 at 10 19 36

- Fixes #1147 

## PR acceptance criteria

### DIFF
--- a/packages/app/src/components/catalog/EntityPage/index.tsx
+++ b/packages/app/src/components/catalog/EntityPage/index.tsx
@@ -294,6 +294,12 @@ export const entityPage = (
     })}
 
     {tab({
+      path: '/quarkus',
+      title: 'Quarkus',
+      mountPoint: 'entity.page.quarkus',
+    })}
+
+    {tab({
       path: '/topology',
       title: 'Topology',
       mountPoint: 'entity.page.topology',
--- a/showcase-docs/dynamic-plugins.md
+++ b/showcase-docs/dynamic-plugins.md
@@ -301,14 +301,18 @@ In order to add dynamic plugin support to a third-party front plugin, without to
 
 - Create a `dynamic-plugins-root` folder at the root of the showcase application repository.
 
-- In the `app-config.yaml` file, add the following entry:
+- To use the `dynamic` plugins packaged within this project, generate and copy them to the `dynamic-plugins-root` using the following commands:
 
-  ```yaml
-  dynamicPlugins:
-    rootDirectory: dynamic-plugins-root
+  ```bash
+  yarn tsc
+  yarn build
+  yarn export-dynamic
+  yarn copy-dynamic-plugins ../dynamic-plugins-root
   ```
 
-- Copy the dynamic plugin package to the `dynamic-plugins-root` folder with the following commands:
+  **Note**: Review the list of the dynamic plugins copied and keep the ones you need or use all of them !
+
+- Copy your dynamic plugin package(s) to be tested/developed to the `dynamic-plugins-root` folder using the following commands:
 
   ```bash
   pkg=<local dist-dynamic sub-folder or external package name of the dynamic plugin package>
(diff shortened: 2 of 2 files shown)


davidfestal (changes requested):
2 remarks here:
- Do you think you could create a dedicated PR for the Docs enhancement ?
- With dynamic plugins, we should not require changing the source code of the frontend application in order to wire a dynamic plugin. That seems like an anti-pattern. **_Users of the showcase that will want to bring their own plugins in a dedicated tab will not be able to change the showcase image_** . @tumido @gashcrumb Any idea about how to contribute a new tab ? I assume we should have something, right ?

davidfestal:
> @tumido @gashcrumb Any idea about how to contribute a new tab ? I assume we should have something, right ?

And if not the it seems to me it would be quite urgent to provide an option.

cmoulliard:
> * That seems like an anti-pattern.

I fully agree with you but my PR has been created based on discussions with @gashcrumb as currenly this is the only way to add a tab

cmoulliard:
> * Do you think you could create a dedicated PR for the Docs enhancement ?

I can do that but then it will be needed to remove commits pushed on this PR ....

davidfestal:
> I can do that but then it will be needed to remove commits pushed on this PR ....

I assume this PR should be rebased yes, or possibly replaced b the new one.

BTW I don't really see the point in adding a TOC, since one is already constructed automatically by GH (as would be in other markdown viewers) based on the Markown structure:

Screenshot from 2024-04-03 11-22-04

davidfestal:
> > The image is available at: `quay.io/janus-idp/backstage-showcase:pr-1154`!
> 
> I did a test on our cluster using the image of the PR and that works too.

My main problem is that it seems to me that **_we should not need to create a new container image_** in order to add a tab specific to a dynamic plugin.

cmoulliard:
> BTW I don't really see the point in adding a TOC, since one is already constructed automatically by GH (as would be in other markdown viewers) based on the Markown structure:

I removed the TOC from this PR

tumido:
So... The code is completely correct and valid.

Secondly, you should really think twice before adding new tabs. The fact that you can't just add a new tab by the dynamic plugins mechanism is 100% intentional. Our aim was to always reduce the amount of tabs we provide. We don't want to end up in a situation where we have a new tab per plugin.

Each tab should represent a purpose that is generic and common to every entity of the same kind and type in the catalog. That's why we don't have a distinct GitHub Actions tab and Tekton tab -> we have a single CI tab instead.

The presence of a new tab should be a PM/UX question, this is not something we should decide only on code quality/review basis.

cmoulliard:
> Each tab should represent a purpose that is generic and common to every entity of the same kind and type in the catalog. That's why we don't have a distinct GitHub Actions tab and Tekton tab -> we have a single CI tab instead.
> 
> The presence of a new tab should be a PM/UX question, this is not something we should decide only on code quality/review basis.

I agree with you and that will discussed soon between the Quarkus's Product Manager and RHDH team.

**FYI**: The Quarkus Console plugin could become (if needed) a runtime plguin supporting: Quarkus, Spring Boot, EAP, etc java runtimes

davidfestal:
> The fact that you can't just add a new tab by the dynamic plugins mechanism is 100% intentional

I think that even this assertion should be validated by PM / UX.
It might be that the availability of such an option (to place a dynamic plugin in a new tab) would still be required, even though not the default or the recommended way.

cc @christophe-f

christophe-f:
Like Tom mentioned, I agree that we should not ship too many tabs by default. 
With that said, a user should be able to create, remove, rename tabs as they wish and not be limited. We should definitely provide this feature.

We've been focusing a lot on the Component kind where most of the tabs are already present, but if someone would like to add more tab to the user, groups or custom Kind, currently they are not allowed.

davidfestal:
Issue https://github.com/janus-idp/backstage-showcase/issues/1160 was opened for this purpose,to avoid fixing this with a plugin-specific fix.

invincibleJai:
> Issue #1160 was opened for this purpose,to avoid fixing this with a plugin-specific fix.

@cmoulliard I see https://github.com/janus-idp/backstage-showcase/pull/1173 is merged and the associated issue is closed

so do we still need this PR or it can be closed?

cc @gashcrumb

gashcrumb:
Yep, this approach shouldn't be necessary anymore, so let's close this.

PR #1173, opened 2024-04-11
feat(app): Support extensible entity tabs

### LINKED ISSUE #1160: Allow adding a new Entity tab through dynamic plugin configuration
### Provide a clear and concise description of the feature request

Currently it is only possible to wire, through configuraion, some frontend dynamic plugin to an available mountpoint, and the plugin content will appear in the corresponding entity tab.
But it is not possible, through configuration, to add a new tab that would contain the dynamic plugin content.

### Describe the problem that this feature resolves

It should be possible to do add the content of a dynamic plugin in a new tab, though it is obviously not recommended to add too many tabs (or one tab per plugin).

### Additional context or screenshots for the feature request

Related PRs that should in fact be tackle with the general mechanism proposed here:
- https://github.com/janus-idp/backstage-showcase/pull/1154
- https://github.com/janus-idp/backstage-showcase/pull/902

gashcrumb:
## Description

This commit adds the ability to customize and extend the default set of tabs available for catalog entity items.  The default set of tabs is configured as part of the static frontend app configuration, overrides can be applied using the regular backstage config file via the `dynamicPlugins.entityTabs` configuration setting.  This commit also fixes a couple rendering warnings that can occur when configuring the same plugin for the same view multiple times.

- Fixes #1160?

## PR acceptance criteria

image

With this change it a new tab could be added like:

```yaml
dynamicPlugins:
	entityTabs:
		/quarkus:                                                # catalog sub-route
			title: Quarkus                                     # tab title
			mountPoint: entity.page.quarkus  #tab mount point
	frontend:
		some-plugin-id:
			mountPoints:
				- mountPoint: entity.page.quarkus
```

An existing tab can be renamed too, for example:
```yaml
dynamicPlugins:
	entityTabs:
		/:                                                           # "Overview" sub route
			title: General                                  # new tab name for "Overview"
```

Visibility of a tab is determined by the availability of items for the configured mount point and the type of catalog entity being displayed.


gashcrumb:
I would argue this Sonarcloud result :-)

gashcrumb:
Something to investigate:

image

gashcrumb:
That issue should be sorted now.  We'll want to have SonarCloud ignore the `baseFrontendConfig.ts` file for duplicate code structures.

gashcrumb:
This could do with some more testing on openshift, let's keep this as a draft for now.

davidfestal (commented):
I' not a big fan of having the `entityTabs` field directly at the `dynamicPlugins` level, because it is still a  *frontend* thing.
I probably would also prefer letting the contribution to a new tab be placed *inside* a frontend plugin config, like this:

```
dynamicPlugins:
  frontend:
    some-plugin-id:
      entityTabs:
        /quarkus:                                                # catalog sub-route
          title: Quarkus                                       # tab title
          mountPoint: entity.page.quarkus    #tab mount point
      mountPoints:
        - mountPoint: entity.page.quarkus
```

gashcrumb:
> I' not a big fan of having the `entityTabs` field directly at the `dynamicPlugins` level, because it is still a _frontend_ thing.

It's not my favorite solution either to be honest but given that `frontend` is a dictionary I didn't want to add it there. 

> I probably would also prefer letting the contribution to a new tab be placed inside a frontend plugin config, like this:

Great suggestion!  I think I can make this work, I might revise the name choice but I think this is a better direction.  Thanks!

davidfestal:
> > @josephca @subhashkhileri as far as I looked into the e2e test failure related to the `dynamic-plugins-info` test case, especially the recorded video, it seems to me that this is related to some authentication failure at start of this test when trying to authenticate with the GitHub auth provider. That doesn't seem related to the content of this PR at all. I restarted the test but it failed again. How should we cope with such situations that sort-of block the trustful merge of awaited PRs ? Is this failure in authentication a know fragile area in the overall e2e test flow that would be known to potentially fail any test case ?
> 
> @davidfestal This authentication failure is known issue. there are two PR's from @gustavolira that should fix all the unstable testcase's. #1186 #1200

Thanks @subhashkhileri for the answer. I assume that if we don't want to merge without a successful e2e test, we have to wait for the merge of those 2 PRs ? Or is there another way ?

subhashkhileri:
> > > @josephca @subhashkhileri as far as I looked into the e2e test failure related to the `dynamic-plugins-info` test case, especially the recorded video, it seems to me that this is related to some authentication failure at start of this test when trying to authenticate with the GitHub auth provider. That doesn't seem related to the content of this PR at all. I restarted the test but it failed again. How should we cope with such situations that sort-of block the trustful merge of awaited PRs ? Is this failure in authentication a know fragile area in the overall e2e test flow that would be known to potentially fail any test case ?
> > 
> > 
> > @davidfestal This authentication failure is known issue. there are two PR's from @gustavolira that should fix all the unstable testcase's. #1186 #1200
> 
> Thanks @subhashkhileri for the answer. I assume that if we don't want to merge without a successful e2e test, we have to wait for the merge of those 2 PRs ? Or is there another way ?

That's right, we have to wait for it. These two PR are fixing and re-enabling some existing test cases. Even if we try other way these test has to be fixed first.

davidfestal:
> > > > @josephca @subhashkhileri as far as I looked into the e2e test failure related to the `dynamic-plugins-info` test case, especially the recorded video, it seems to me that this is related to some authentication failure at start of this test when trying to authenticate with the GitHub auth provider. That doesn't seem related to the content of this PR at all. I restarted the test but it failed again. How should we cope with such situations that sort-of block the trustful merge of awaited PRs ? Is this failure in authentication a know fragile area in the overall e2e test flow that would be known to potentially fail any test case ?
> > > 
> > > 
> > > @davidfestal This authentication failure is known issue. there are two PR's from @gustavolira that should fix all the unstable testcase's. #1186 #1200
> > 
> > 
> > Thanks @subhashkhileri for the answer. I assume that if we don't want to merge without a successful e2e test, we have to wait for the merge of those 2 PRs ? Or is there another way ?
> 
> That's right, we have to wait for it. These two PR are fixing and re-enabling some existing test cases. Even if we try other way these test has to be fixed first.

Any ETA ?

subhashkhileri:
> > > > > @josephca @subhashkhileri as far as I looked into the e2e test failure related to the `dynamic-plugins-info` test case, especially the recorded video, it seems to me that this is related to some authentication failure at start of this test when trying to authenticate with the GitHub auth provider. That doesn't seem related to the content of this PR at all. I restarted the test but it failed again. How should we cope with such situations that sort-of block the trustful merge of awaited PRs ? Is this failure in authentication a know fragile area in the overall e2e test flow that would be known to potentially fail any test case ?
> > > > 
> > > > 
> > > > @davidfestal This authentication failure is known issue. there are two PR's from @gustavolira that should fix all the unstable testcase's. #1186 #1200
> > > 
> > > 
> > > Thanks @subhashkhileri for the answer. I assume that if we don't want to merge without a successful e2e test, we have to wait for the merge of those 2 PRs ? Or is there another way ?
> > 
> > 
> > That's right, we have to wait for it. These two PR are fixing and re-enabling some existing test cases. Even if we try other way these test has to be fixed first.
> 
> Any ETA ?

@gustavolira can provide more details on that.

packages/app/src/components/catalog/EntityPage/EntityPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+      mountPoint: 'entity.page.definition',
+    },
+    '/system': {
+      title: 'Diagram',
+      mountPoint: 'entity.page.diagram',
+    },
+  };
+  return (
+    <EntityLayout>
+      {Object.entries({ ...defaultTabs, ...entityTabOverrides }).map(
```
rohitkrai03, line 131: @gashcrumb Is there a way to configure the order of the tab? For example if I wanted to add a tab just after `Topology` one or just before `CI` one?

rohitkrai03: We could probably add something like `insertBefore` and `insertAfter` in the extension to let the users configure ordering of tabs. We did something similar in OpenShift Console Dynamic Plugin extension points.

gashcrumb: Yep, we certainly could do that, either now or in a subsequent improvement.  It's sort of possible to influence the order of the tabs with this change by nature of the way the dictionaries are combined, but having a more explicit means to do this would ultimately be nicer.

gashcrumb: Thinking on this further, with this change you can specify whatever order you like by overriding the default list in the config.  If in the future we decide to reduce [this list](https://github.com/janus-idp/backstage-showcase/pull/1173/files#diff-dc23b2c0e4792453496e74f891819cf0cf5172dd3d6b426bca6b9f85aef4d7e5R64-R128) and move the tab config into the plugin config (for example, move `/topology` to an `entityTabs` setting in [here](https://github.com/janus-idp/backstage-showcase/blob/main/dynamic-plugins.default.yaml#L229)) then we would definitely need these additional directives to influence the tab order.

rohitkrai03: For setting up a custom order right now users would need to override the complete list of default tabs, right? Otherwise I am not sure if `{ ...defaultTabs, ...entityTabOverrides }` would result in the desired order every time.

gashcrumb: The order is the order of enumerated properties of the combined object which indeed is not necessarily straightforward to reason through.  But keep in mind that the original goal of this effort is mainly to allow users to specify new tabs in addition to what is provided out of the box so that a user's plugin could have it's own entity page.  Being able to customize the existing tabs is also nice but not a primary goal of this feature.

davidfestal (approved):

davidfestal:
@gashcrumb About the SonarCould duplication report related to this block: https://github.com/janus-idp/backstage-showcase/pull/1173/files#diff-dc23b2c0e4792453496e74f891819cf0cf5172dd3d6b426bca6b9f85aef4d7e5R67-R128, would it make sense to extract this `defaultTabs` constant to its own file (or a file dedicated to such default config constants).
This would allow adding a SonarCloud exception for this file only with the following content in a [`.sonarcloud.properties` file](https://docs.sonarsource.com/sonarcloud/advanced-setup/automatic-analysis/#additional-analysis-configuration):
```
sonar.cpd.exclusions=packages/app/src/components/catalog/EntityPage/<your new file>
```

I'm not saying it should be done *now* in this PR, but at least it might be something we want to think about to avoid bypassing the SonarCloud analysis in such cases and having to force-merge.

We can also add the corresponding setting directl in the SonarCloud project settings, but I don't have the right permissions.
cc @kadel Any idea who can do that ?

gashcrumb:
> @gashcrumb About the SonarCould duplication report related to this block: [#1173 (files)](https://github.com/janus-idp/backstage-showcase/pull/1173/files#diff-dc23b2c0e4792453496e74f891819cf0cf5172dd3d6b426bca6b9f85aef4d7e5R67-R128), would it make sense to extract this `defaultTabs` constant to its own file (or a file dedicated to such default config constants). This would allow adding a SonarCloud exception for this file only with the following content in a [`.sonarcloud.properties` file](https://docs.sonarsource.com/sonarcloud/advanced-setup/automatic-analysis/#additional-analysis-configuration):
> 
> ```
> sonar.cpd.exclusions=packages/app/src/components/catalog/EntityPage/<your new file>
> ```
> 
> I'm not saying it should be done _now_ in this PR, but at least it might be something we want to think about to avoid bypassing the SonarCloud analysis in such cases and having to force-merge.
> 
> We can also add the corresponding setting directl in the SonarCloud project settings, but I don't have the right permissions, which would be much simpler obviously.
> 
> cc @kadel Any idea who can do that ?

I can do that in this PR.  I think probably this is the way to go too, as then we've some pointer of these duplicate lines somewhere.

Also likely #1202 will also need a similar approach to handle the mapping that needs to be done to match up the API response to the config.

PR #1200, opened 2024-04-22
Update openshift ci configuration to create showcase and showcase-rba…

gustavolira:
…c installation.

## Description

- Fixes #?

## PR acceptance criteria


.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -235,20 +244,20 @@ main() {
...
   uninstall_helmchart
 
   cd $DIR
-  apply_yaml_files $DIR
+  apply_yaml_files $DIR "$NAME_SPACE"
   add_helm_repos
 
   echo "Tag name : ${TAG_NAME}"
-  helm upgrade -i ${RELEASE_NAME} -n ${NAME_SPACE} ${HELM_REPO_NAME}/${HELM_IMAGE_NAME} --version ${CHART_VERSION} -f $DIR/value_files/${HELM_CHART_VALUE_FILE_NAME} --set global.clusterRouterBase=${K8S_CLUSTER_ROUTER_BASE} --set upstream.backstage.image.tag=${TAG_NAME}
+  helm upgrade -i "${NAME_SPACE}" -n ${NAME_SPACE} ${HELM_REPO_NAME}/${HELM_IMAGE_NAME} --version ${CHART_VERSION} -f $DIR/value_files/${HELM_CHART_VALUE_FILE_NAME} --set global.clusterRouterBase=${K8S_CLUSTER_ROUTER_BASE} --set upstream.backstage.image.tag=${TAG_NAME}
```
subhashkhileri, line 258: Why has been `${RELEASE_NAME}` removed and replaced with `"${NAME_SPACE}"`?
The current command appears incorrect; we need to include the release name.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -259,7 +268,36 @@ main() {
...
+  install_helm
+  uninstall_helmchart
+
+  cd "${DIR}"
+  apply_yaml_files "${DIR}" ${NAME_SPACE_RBAC}
+  add_helm_repos
+
+  echo "Tag name : ${TAG_NAME}"
+
+  helm upgrade -i ${NAME_SPACE_RBAC} -n ${NAME_SPACE_RBAC} rhdh-chart/backstage --version ${CHART_VERSION} -f $DIR/value_files/${HELM_CHART_VALUE_FILE_NAME} --set global.clusterRouterBase=${K8S_CLUSTER_ROUTER_BASE} --set upstream.backstage.image.tag=${TAG_NAME}
```
subhashkhileri, line 286:
```suggestion
  helm upgrade -i ${RELEASE_NAME} -n ${NAME_SPACE_RBAC} ${HELM_REPO_NAME}/${HELM_IMAGE_NAME} --version ${CHART_VERSION} -f $DIR/value_files/${HELM_CHART_VALUE_FILE_NAME} --set global.clusterRouterBase=${K8S_CLUSTER_ROUTER_BASE} --set upstream.backstage.image.tag=${TAG_NAME}
```

Please update the Helm upgrade as suggested. The current command appears incorrect; we need to include the release name and use the predefined environments for rhdh-chart/backstage as suggested above.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -259,7 +268,36 @@ main() {
...
+  if [ $backstage_status -ne 0 ]; then
+    echo "Backstage is not running. Exiting..."
+    exit 1
+  fi
+
+  run_tests ${NAME_SPACE_RBAC}
 }
 
 main
+main_rbac
```
subhashkhileri, line 303: Deploying RHDH first and waiting for its running status takes time. Since this approach requires another deployment for RBAC, it doubles the waiting time.

A suggested improvement is to create a separate Helm deployment function for regular tests, then immediately initiate a deployment for RBAC without waiting/checking for the previous deployments running state. we can check the running status in `main` and `main_rbac`, as they do now.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -152,47 +159,49 @@ run_tests() {
...
-      echo "######## BASE URL ########"
-      echo "$BASE_URL"
-      return 0
-    else
-      echo "Attempt $i of $max_attempts: Backstage not yet available (HTTP Status: $http_status)" | tee -a "/tmp/${LOGFILE}"
-      sleep $wait_seconds
-    fi
-  done
+  local release_name=$1
+    local namespace=$1
```
subhashkhileri, line 179: code in this check_backstage_running function does not appear to be properly formatted. there are extra white spaces.

subhashkhileri (changes requested):

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -142,6 +148,7 @@ apply_yaml_files() {
 }
 
 run_tests() {
+  local project=$1
```
josephca, line 151: There is an extra white space. Please fix indentation

.ibm/pipelines/openshift-ci-tests.sh
josephca: Even if  ${RELEASE_NAME} and ${NAME_SPACE} have the same value, it would be clearer if we use separate property names. A code reviewer would suspect it as a typo.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -4,6 +4,7 @@ set -e
 
 LOGFILE="test-log"
```
josephca, line 5: I think LOGFILE was used for pr-commenter. If it's no longer being used, we can safely remove the variable and coded that use 'LOGFILE'. @subhashkhileri - can you confirm?

subhashkhileri: @josephca This is required. It's log file where html log report is stored.
like : [Link](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/janus-idp_backstage-showcase/1170/pull-ci-janus-idp-backstage-showcase-main-e2e-tests/1781348181875888128/artifacts/e2e-tests/janus-idp-backstage-showcase/artifacts/)

.ibm/pipelines/openshift-ci-tests.sh
gustavolira: @subhashkhileri release name can be anything, we can just choose any name that we want

gustavolira: @subhashkhileri updated

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -152,32 +158,34 @@ run_tests() {
...
 
   pkill Xvfb
 
-  mkdir ${ARTIFACT_DIR}/test-results
-  cp -a /tmp/backstage-showcase/e2e-tests/test-results/* ${ARTIFACT_DIR}/test-results
+  mkdir -p ${ARTIFACT_DIR}/$project/test-results
+  cp -a /tmp/backstage-showcase/e2e-tests/test-results/* ${ARTIFACT_DIR}/$project/test-results
 
   ansi2html <"/tmp/${LOGFILE}" >"/tmp/${LOGFILE}.html"
   cp -a "/tmp/${LOGFILE}.html" ${ARTIFACT_DIR}
```
subhashkhileri, line 172: showcase log file will be overwritten by showcase-rbac test runs. better to move it inside the $project like `${ARTIFACT_DIR}/${project}`

subhashkhileri (changes requested):

.ibm/pipelines/openshift-ci-tests.sh
gustavolira: sorry for my miss understanding, but why will be overwritten if I'm already using `${ARTIFACT_DIR}/$projec` ?

subhashkhileri: As per my understanding you are storing the `test-results` as per the project but we also have [test-log.html](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/janus-idp_backstage-showcase/1200/pull-ci-janus-idp-backstage-showcase-main-e2e-tests/1783100244364365824/artifacts/e2e-tests/janus-idp-backstage-showcase/artifacts/test-log.html). which is currently outside the `$project` and both `showcase` and `showcase-rbac` will have there own `test-log.html` which needs to be moved inside the `$project`. like: 

```suggestion
  cp -a "/tmp/${LOGFILE}.html" ${ARTIFACT_DIR}/${project}
```

e2e-tests/playwright/utils/Common.ts
```diff
@@ -54,13 +54,50 @@ export class Common {
 
   async loginAsGithubUser() {
     await this.logintoGithub();
-    await this.page.goto('/');
+    await this.page.goto('/', { waitUntil: 'networkidle' });
```
subhashkhileri, line 57: The network might not always work or can take more time. I debugged and found that problem is there is a `MuiCircularProgress-root` loading bar that appears before the sign-in page loads. We have `await common.waitForLoad();` which waits until the loading bar is not visible, and you can pass a timeout as needed, with the default being 120 seconds.

e2e-tests/playwright/utils/Common.ts
```diff
@@ -54,13 +54,50 @@ export class Common {
...
     await this.page.waitForLoadState('load');
     await this.uiHelper.clickButton('Sign In');
     await this.checkAndReauthorizeGithubApp();
     await this.uiHelper.waitForSideBarVisible();
   }
 
+  async loginAsGithubUserWithRetry(retries: number = 3) {
```
subhashkhileri, line 64: @gustavolira why do we need to retry for login? It should work on the first attempt.

e2e-tests/playwright/utils/UIhelper.ts
```diff
@@ -104,7 +104,25 @@ export class UIhelper {
   }
 
   async waitForSideBarVisible() {
-    await this.page.waitForSelector('nav a');
+    await this.waitForElementWithRetry('nav a');
```
subhashkhileri, line 107: Doesn't `waitForSelector()` already retry until it reaches its default timeout? Maybe increasing the timeout, like `await this.page.waitForSelector('nav a', { timeout: 120000 });`, could be helpful.

e2e-tests/playwright.config.ts
```diff
@@ -9,7 +9,7 @@ import { defineConfig, devices } from '@playwright/test';
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
-  timeout: 60000,
+  timeout: 120000,
```
subhashkhileri, line 12: I think a 60-second timeout is a reasonable limit. If we need more time, we could increase it to 70 or 80 seconds, but 120 seconds for all actions seems too long. It might be better to adjust the timeout on individual actions, depending on the specific action.

e2e-tests/playwright/utils/Common.ts
gustavolira: > @gustavolira why do we need to retry for login? It should work on the first attempt.

@subhashkhileri  it's because this is the most flaky test, if it works at the first attempt the retry loop will be stopped and the code will move forward

.ibm/pipelines/env_variables.sh
```diff
@@ -21,20 +24,24 @@ GITHUB_APP_WEBHOOK_URL=aHR0cHM6Ly9zbWVlLmlvL0NrRUNLYVgwNzhyZVhobEpEVzA=
 GITHUB_APP_WEBHOOK_SECRET=$(cat /tmp/secrets/GITHUB_APP_WEBHOOK_SECRET)
 GITHUB_URL=aHR0cHM6Ly9naXRodWIuY29t
 GITHUB_ORG=amFudXMtcWU=
+GH_USER_ID=$(cat /tmp/secrets/GH_USER_ID)
+GH_USER_PASS=$(cat /tmp/secrets/GH_USER_PASS)
+GH_2FA_SECRET=$(cat /tmp/secrets/GH_2FA_SECRET)
```
subhashkhileri, line 30:
```suggestion
GH_RHDH_QE_USER_TOKEN=$(cat /tmp/secrets/GH_RHDH_QE_USER_TOKEN)

```

@gustavolira Please add this to resolve current github test failure.

josephca (approved):
/lgtm
/approve

PR #1202, opened 2024-04-23
feat(admin): identify external plugins

ciiay:
## Description

As a user of RHDH, I want to know which third party plugins are installed so that I know that there could be unsupported plugins running in my RHDH installation.

- Fixes [RHIDP-1854](https://issues.redhat.com/browse/RHIDP-1854): Identify externally installed plugins in the dynamic-plugins-info table

## PR acceptance criteria

Screenshot(updated 5/1)
rhidp_1854_update_51_1

_Note_: for internal packages that are not enabled, Version and Role values are left empty in table.
rhidp_1854_update_51_2

**Passed e2e tests**
```
➜  e2e-tests git:(rhidp-1854-identify-externally-installed-plugins) ✗ yarn showcase dynamic-plugins-info.spec
yarn run v1.22.19
$ playwright test --project=showcase dynamic-plugins-info.spec

Running 1 test using 1 worker

  ✓  1 … e2e/plugins/dynamic-plugins-info/dynamic-plugins-info.spec.ts:19:3 › dynamic-plugins-info UI tests › it should show a table, and the table should contain techdocs plugins (5.2s)
Dialog message: Failed to sign in as a guest using the auth backend. Do you want to fallback to the legacy guest token?

  1 passed (9.5s)
✨  Done in 10.39s.
```


davidfestal:
I thought we had discussed that this information could, for 1.2, be hard-coded on the client side ?
cc @durandom 

I'm of course not against making something more automated, but I'm not  a fan of the code depending on the content of the `dynamic-plugins.default.yaml`.

Another point to consider is the following; the fact that a plugin is mentioned in the `dynamic-plugins.default.yaml` file does *not* mean that it is supported, it only means that the plugin is embedded inside the RHDH container image and pre-configured. In other words it tells **whether the plugin is external or not**.

cc @nickboldt

ciiay:
Hi @davidfestal , thanks for the review. I also felt that package value in `dynamic-plugins.default.yaml` isn't very reliable for deciding if a plugin is external or rhdh embedded while working on it. Do we have a list of rhdh embedded plugin I can use as hard coded value for now?

davidfestal:
cc @gashcrumb as the related EPIC owner.

ciiay:
Hi @gashcrumb The mapping component seems not passing the SonarCloud duplication check. Did I misunderstand the mapping from our discussion?

gashcrumb:
> Hi @gashcrumb The mapping component seems not passing the SonarCloud duplication check. Did I misunderstand the mapping from our discussion?

Just took a quick glance (will look closer tomorrow) and I think this is along the lines of what we've discussed.  We may have to add an exception in SonarCloud for this (I need to learn how to do this for my own PR too), probably also worth adding a comment if there isn't one already explaining what the mapping is for.

plugins/dynamic-plugins-info/src/components/InternalPluginsMap.tsx
```diff
@@ -0,0 +1,126 @@
...
+  'backstage-plugin-github-actions':
+    './dynamic-plugins/dist/backstage-plugin-github-actions',
+  'backstage-plugin-github-issues':
+    './dynamic-plugins/dist/backstage-plugin-github-issues',
+  'roadiehq-backstage-plugin-github-insights':
+    './dynamic-plugins/dist/roadiehq-backstage-plugin-github-insights',
+  'roadiehq-backstage-plugin-github-pull-requests':
+    './dynamic-plugins/dist/roadiehq-backstage-plugin-github-pull-requests',
+  'roadiehq-backstage-plugin-security-insights':
+    'roadiehq-backstage-plugin-security-insights',
```
davidfestal, line 20: Is it expected that this one is not in `./dynamic-plugins/dist` ?

plugins/dynamic-plugins-info/src/components/DynamicPluginsTable/DynamicPluginsTable.tsx
```diff
@@ -24,11 +29,25 @@ export const DynamicPluginsTable = () => {
...
+      width: '15%',
+    },
+    {
+      title: 'Enabled',
+      field: 'enabled',
+      render: ({ enabled }) => <>{enabled ? 'Yes' : 'No'}</>,
+      width: '10%',
+    },
+    {
+      title: 'Internal',
```
davidfestal, line 41: I'd rather put `Preinstalled` if we want to stick to the documentation wording.

plugins/dynamic-plugins-info/src/components/InternalPluginsMap.tsx
ciiay: @davidfestal Thanks for the review. Good catch on this typo. Will fix it.

gashcrumb:
@ciiay Probably [this line](https://github.com/janus-idp/backstage-showcase/blob/main/e2e-tests/playwright/e2e/plugins/dynamic-plugins-info/dynamic-plugins-info.spec.ts#L25) will need updating for the e2e tests.  This particular e2e test should be runnable locally without too much hassle, let me know if I can help you set that up.

gashcrumb:
Awesome work BTW, swapped this into my existing setup:

image

davidfestal:
Not sure I'm a fan of the `Total Plugins` wording, even though I don't know what could replace it.

Is there an ability to filter the enabled / disabled ones ? Maybe a Button, something visual that would make it clear what is the list we're seeing ?

gashcrumb:
> Not sure I'm a fan of the `Total Plugins` wording, even though I don't know what could replace it.

Maybe just `Plugins`?

> Is there an ability to filter the enabled / disabled ones ? Maybe a Button, something visual that would make it clear what is the list we're seeing ?

The sorting can be used to get an idea but I agree it does feel like a quick filter of some variety would be nicer.  I think the filter widget that's used on other catalog pages is probably too much for this purpose.  @ShiranHi I wonder if there's other quick filter options we could use here?

ciiay:
> Not sure I'm a fan of the `Total Plugins` wording, even though I don't know what could replace it.
In the Support pop up, it uses `All of the installed plugins`. How about `Installed plugins` or simply `Available plugins`?

cc @gashcrumb

gashcrumb:
I think we should probably update the support popup to say "All of the plugins" and I still think maybe just simplifying the page title to just "Plugins" would be best here, as now the table shows all of this.  I was thinking this more generic title fits if we also have simple filter control perhaps in the toolbar area for quickly filtering out disabled plugins as @davidfestal suggests.

gashcrumb:
There's this [filtered table](https://backstage.io/storybook/?path=/story/data-display-table--filter-table) option we could try out.

ShiranHi:
> > Is there an ability to filter the enabled / disabled ones ? Maybe a Button, something visual that would make it clear what is the list we're seeing ?
> 
> The sorting can be used to get an idea but I agree it does feel like a quick filter of some variety would be nicer. I think the filter widget that's used on other catalog pages is probably too much for this purpose. @ShiranHi I wonder if there's other quick filter options we could use here?

I think we don't need to add a filter bar to this page. The columns are sortable, and users can use the search bar if they need to find something specific.
 
> I think we should probably update the support popup to say "All of the plugins" and I still think maybe just simplifying the page title to just "Plugins" would be best here, as now the table shows all of this. I was thinking this more generic title fits if we also have simple filter control perhaps in the toolbar area for quickly filtering out disabled plugins as @davidfestal suggests.

We decided against displaying any links in the UI. This is because the page is hardcoded, meaning any changes would require users to wait for a new release. Additionally, this approach wouldn't function in air-gapped environments. So I don't think we need to add this support button. 

I have 3 questions:
1. I noticed in the table that there's a "Role" column. Could we simply specify whether it's "Backend" or "Frontend"? The current text there seems repetitive. Can we change the column name to be "Type"?
2. What does it mean if a plugin is pre-installed but not enabled?
3. Should we include the source, indicating whether this is a Red Hat plugin or a community plugin?

gashcrumb:
BTW we did try the filtered table, however it introduces some layout issues and the actual filter panel doesn't seem to work, which will require additional investigation.  I propose we hold off introducing any additional filtering into this PR, we can always do this as an enhancement later.

ShiranHi:
> What about the question of "Installed Plugins" vs "Total Plugins" vs "Plugins" for the table title?

Right, I missed that. Only "Plugins" should work imo

gashcrumb (approved):
:+1:

gashcrumb:
Huh, didn't realize that would add both labels, a 2nd look by someone else would be good too :smile:

gashcrumb:
Or would have been, I guess it was on the merge queue already even though I removed the label manually.  @kadel wonder why both labels were applied at once, was it the :+1: in my comment when I used the approve button?

ciiay:
@gashcrumb I can raise a separate PR to address any further comments for this PR.

@davidfestal @ShiranHi Let me know if this PR needs more work. Thank you.

PR #1216, opened 2024-04-29
Update playwright config and enable screenshots and video recording for serial tests

subhashkhileri:
### Update Test Configuration and tests so that :

1. Report Flaky Tests as Failures: Modify the test configuration so that tests identified as flaky are automatically marked as failures.
2. Enable Screenshots on Failure for all the test case's
3. Refactor beforeAll Hook for Serial Test Cases for reusability.
4. Enable video recording for serial test cases, which is not done by default even though it's enabled in the Playwright config.
5. Increase the video recording resolution.
6. Remove Unused Scripts: Delete the outdated scripts located at .ibm/pipelines/openshift-tests.sh and .ibm/pipelines/openshift-tests.sh to clean up the codebase and reduce maintenance overhead.

https://issues.redhat.com/browse/RHIDP-2214

## PR acceptance criteria


.ibm/pipelines/functions.sh
nickboldt, line 1: did you intend to delete these .sh files? they're not mentioned in the issue/PR description.

subhashkhileri: I will update the PR description

e2e-tests/playwright.config.ts
```diff
@@ -11,12 +11,8 @@ import { defineConfig, devices } from '@playwright/test';
 export default defineConfig({
   timeout: 80000,
   testDir: './playwright',
-  /* Run tests in files in parallel */
-  fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
-  /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
```
gustavolira, line 19: It's better to receive a flaky test than an error test and have the build fail.
Always failing the build instead of receiving a flaky test will burden pull request approvals and create bottlenecks

josephca: > t's better to receive a flaky test than an error test and have the build fail.
> Always failing the build instead of receiving a flaky test will burden pull request approvals and create bottlenecks

Agreed unless there is a reason why we should report a flaky test as a failure. :)

subhashkhileri: I don't think it can create bottlenecks because it might pass when e2e pipeline is re-triggered because its just flaky test that happen to fail on 1st attempt.
I was considering marking tests as failures because it forces us to address the issues directly, rather than increasing more flaky tests. 
also It increases execution time and does not give confidence in the run when it reports some tests flaky. The majority of the time, tests are flaky due to not having proper waiting mechanisms for elements.
Also consider a scenario where actual UI or element is flaky in that case the test might pass on next retry.

But definitely it might increase  number of re-testing on PR.

I can revert back keep the retries. wdyt? @josephca @gustavolira

subhashkhileri: Okay, So I have reverted the change.

josephca (approved):

PR #1225, opened 2024-05-03
feat(backend): Support running behind a corporate proxy [RHIDP-2217]

rm3l:
## Description

This PR allows running the Showcase app behind a corporate proxy, which was the root cause of a login failure using the GitHub auth behind a proxy.

It does so per the upstream instructions provided on this page: https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/help-im-behind-a-corporate-proxy.md
This way, it can honor the conventional `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables (and their lowercase equivalents).

The `undici` and `global-agent` packages added here allow covering settings for the native `fetch` and `node-fetch` respectively. Note that even if `node-fetch` is recommended for HTTP data fetching by Backstage packages [[ADR013](https://backstage.io/docs/architecture-decisions/adrs-adr013/)], we identified a couple of plugins making use of other libraries like Axios (like the Keycloak backend plugin - see [RHIDP-2217](https://issues.redhat.com/browse/RHIDP-2217?focusedId=24665743&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24665743) for the full list). Axios reportedly already honors the proxy environment variables above [[doc](https://axios-http.com/docs/req_config)].
I tested the Keycloak backend plugin as an example to ensure this.

- Fixes [RHIDP-2217](https://issues.redhat.com//browse/RHIDP-2217)

## PR acceptance criteria

### Local Development
You can run locally with `yarn`:

1. First start a proxy server - example with a [Squid](https://www.squid-cache.org/) image:
```
podman container run --rm --name squid-container -e TZ=UTC -p 3128:3128 -it docker.io/ubuntu/squid:latest 
```

2. In a separate terminal, [start the application](https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#getting-started-running-backstage-showcase) with the proxy environment variables.
```
GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE='' \
  HTTP_PROXY=http://localhost:3128 \
  HTTPS_PROXY=http://localhost:3128 \
  NO_PROXY='localhost,example.org' \
  yarn start
```

3. Inspect the proxy logs and you should see that requests are sent through the proxy, e.g.:
```
1715067448.970    185 10.42.0.19 TCP_TUNNEL/200 5171 CONNECT raw.githubusercontent.com:443 - HIER_DIRECT/185.199.111.133 -
1715067450.517    256 10.42.0.19 TCP_TUNNEL/200 5129 CONNECT raw.githubusercontent.com:443 - HIER_DIRECT/185.199.111.133 -
```

### Full test
For a more complete test, the most challenging part of testing the changes here is to set up an environment where an application is forbidden access to the public Internet except through a given proxy.
We can simulate such an environment in a Kubernetes namespace with the help of [Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to control ingress/egress traffic.
For example:

1. Make sure the network plugin in your Kubernetes cluster supports network policies. I created a cluster with [k3d](https://k3d.io/v5.6.3/) (`k3d cluster create`), and it supports Network Policies out of the box.

5. Create a separate `proxy` namespace, and deploy a [Squid](https://www.squid-cache.org/)-based proxy application there:
```sh
kubectl create namespace proxy

cat <<EOF | kubectl -n proxy apply -f -
apiVersion: v1
kind: Service
metadata:
  name: squid-service
  namespace: proxy
  labels:
    app: squid
spec:
  ports:
  - port: 3128
  selector:
    app: squid

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: squid-deployment
  namespace: proxy
spec:
  replicas: 1
  selector:
    matchLabels:
      app: "squid"
  template:
    metadata:
      labels:
        app: squid
    spec:
      containers:
      - name: squid
        image: docker.io/ubuntu/squid:latest
        ports:
        - containerPort: 3128
          name: squid
          protocol: TCP
EOF
```

6. Create the namespace where the Showcase application will be running
```
kubectl create namespace my-ns
```

7. Add the network policies in the namespace above. The first one denies all egress traffic except to the DNS resolver and the Squid proxy. The second one allows ingress and egress traffic in the same namespace, because the Showcase app pod needs to contact the local Database pod.
```sh
cat <<EOF | kubectl -n my-ns apply -f -
---
# Deny all egress traffic in this namespace => proxy settings can be used to overcome this.
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: default-deny-egress-with-exceptions
spec:
  podSelector: {}
  policyTypes:
  - Egress
  egress:
  # allow DNS resolution (we need this allowed, otherwise we won't be able to resolve the DNS name of the Squid proxy service)
  - to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: kube-system
      podSelector:
        matchLabels:
          k8s-app: kube-dns
    ports:
      - port: 53
        protocol: UDP
      - port: 53
        protocol: TCP
  # allow traffic to Squid proxy
  - to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: proxy
    ports:
    - port: 3128
      protocol: TCP

---
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: allow-same-namespace
spec:
  podSelector: {}
  ingress:
  - from:
    - podSelector: {}
  egress:
  - to:
    - podSelector: {}
EOF
```

8. Deploy the showcase application using the image from this PR (`quay.io/janus-idp/backstage-showcase:pr-1225`). I tested this using the [RHDH Operator](https://github.com/janus-idp/operator), but it should work with the [Helm Chart](https://github.com/redhat-developer/rhdh-chart) as well. For example, to only test the GitHub login issue:
  a. Follow the instructions to create a GitHub app
  b. Make sure to use the image from this PR (`spec.application.image` field if using the Operator).
  c. GitHub login should fail unless the `HTTP(S)_PROXY` environment variables are set, like so:
```yaml
# Make sure to create the appropriate ConfigMaps and Secrets
apiVersion: rhdh.redhat.com/v1alpha1
kind: Backstage
metadata:
  name: my-rhdh
spec:
  application:
    image: quay.io/janus-idp/backstage-showcase:pr-1225
    appConfig:
      configMaps:
        - name: app-config-rhdh
    dynamicPluginsConfigMapName: dynamic-plugins-rhdh
    extraEnvs:
      envs:
        - name: HTTP_PROXY
          value: 'http://squid-service.proxy.svc.cluster.local:3128'
        - name: HTTPS_PROXY
          value: 'http://squid-service.proxy.svc.cluster.local:3128'
        - name: NO_PROXY
          value: 'localhost'
        - name: ROARR_LOG
          # Logs from global-agent (to inspect proxy settings)
          value: 'true'
      secrets:
        - name: secrets-rhdh

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: app-config-rhdh
data:
  "app-config-rhdh-auth.yaml": |
    auth:
      environment: development
      session:
        secret: my-super-secret-secret
      providers:
        github:
          development:
            clientId: '${GH_CLIENT_ID}'
            clientSecret: '${GH_CLIENT_SECRET}'

  "app-config-rhdh-catalog.yaml": |
    catalog:
      import:
        entityFilename: catalog-info.yaml
        pullRequestBranchName: backstage-integration
      rules:
        - allow: [Component, System, Group, Resource, Location, Template, API]
      locations:
        # Note: integrations.github[].apps must be correctly configured to read GitHub locations
        - type: url
          target: https://github.com/janus-idp/backstage-showcase/blob/main/catalog-entities/all.yaml

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: dynamic-plugins-rhdh
data:
  dynamic-plugins.yaml: |
    includes:
      - dynamic-plugins.default.yaml
    plugins:
      - package: './dynamic-plugins/dist/backstage-plugin-catalog-backend-module-github-dynamic'
        disabled: false
        pluginConfig:
          catalog:
            providers:
              github:
                myorg:
                  organization: '${GH_ORG}'
                  schedule:
                    # supports cron, ISO duration, "human duration" (used below)
                    frequency: { minutes: 30}
                    # supports ISO duration, "human duration (used below)
                    timeout: { minutes: 3}
                    initialDelay: { seconds: 15}

---
apiVersion: v1
kind: Secret
metadata:
  name: secrets-rhdh
stringData:
  # generated with the command below (from https://backstage.io/docs/auth/service-to-service-auth/#setup):
  # node -p 'require("crypto").randomBytes(24).toString("base64")'
  BACKEND_SECRET: "R2FxRVNrcmwzYzhhN3l0V1VRcnQ3L1pLT09WaVhDNUEK" # notsecret
  GH_ORG: "my-gh-org"
  GH_CLIENT_ID: "changeme"
  GH_CLIENT_SECRET: "changeme"
```

### DIFF
--- a/packages/backend/src/corporate-proxy.ts
+++ b/packages/backend/src/corporate-proxy.ts
@@ -1,0 +1,103 @@
+// Copyright 2024 The Janus IDP Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import { bootstrap } from 'global-agent';
+import { Agent, ProxyAgent, Dispatcher, setGlobalDispatcher } from 'undici';
+
+/**
+ * Adds support for corporate proxy to both 'node-fetch' (using 'global-agent') and native 'fetch' (using 'undici') packages.
+ *
+ * Ref: https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/help-im-behind-a-corporate-proxy.md
+ * Ref: https://gist.github.com/zicklag/1bb50db6c5138de347c224fda14286da (to support 'no_proxy')
+ */
+export function configureCorporateProxyAgent() {
+  // Bootstrap global-agent, which addresses node-fetch proxy-ing.
+  // global-agent purposely uses namespaced env vars to prevent conflicting behavior with other libraries,
+  // but user can set GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE to an empty value for global-agent to use
+  // the conventional HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables.
+  // More details in https://github.com/gajus/global-agent#what-is-the-reason-global-agentbootstrap-does-not-use-http_proxy
+  bootstrap();
+
+  // Configure the undici package, which affects the native 'fetch'. It leverages the same env vars used by global-agent,
+  // or the more conventional HTTP(S)_PROXY ones.
+  const proxyEnv =
+    process.env.GLOBAL_AGENT_HTTP_PROXY ??
+    process.env.GLOBAL_AGENT_HTTPS_PROXY ??
+    process.env.HTTP_PROXY ??
+    process.env.http_proxy ??
+    process.env.HTTPS_PROXY ??
+    process.env.https_proxy;
+
+  if (proxyEnv) {
+    const proxyUrl = new URL(proxyEnv);
+
+    // Create an access token if the proxy requires authentication
+    let token: string | undefined = undefined;
+    if (proxyUrl.username && proxyUrl.password) {
+      const b64 = Buffer.from(
+        `${proxyUrl.username}:${proxyUrl.password}`,
+      ).toString('base64');
+      token = `Basic ${b64}`;
+    }
+
+    // Create a default agent that will be used for no_proxy origins
+    const defaultAgent = new Agent();
+
+    // Create an interceptor that will use the appropriate agent based on the origin and the no_proxy
+    // environment variable.
+    // Collect the list of domains that we should not use a proxy for.
+    // The only wildcard available is a single * character, which matches all hosts, and effectively disables the proxy.
+    const noProxyEnv =
+      process.env.GLOBAL_AGENT_NO_PROXY ??
+      process.env.NO_PROXY ??
+      process.env.no_proxy;
+    const noProxyList = noProxyEnv?.split(',') || [];
+
+    const isNoProxy = (origin?: string): boolean => {
+      for (const exclusion of noProxyList) {
+        if (exclusion === '*') {
+          // Effectively disables proxying
+          return true;
+        }
+        // Matched as either a domain which contains the hostname, or the hostname itself.
+        if (origin === exclusion || origin?.endsWith(`.${exclusion}`)) {
+          return true;
+        }
+      }
+      return false;
+    };
+
+    const noProxyInterceptor = (
+      dispatch: Dispatcher['dispatch'],
+    ): Dispatcher['dispatch'] => {
+      return (opts, handler) => {
+        return isNoProxy(opts.origin?.toString())
+          ? defaultAgent.dispatch(opts, handler)
+          : dispatch(opts, handler);
+      };
+    };
+
+    // Create a proxy agent that will send all requests through the configured proxy, unless the
+    // noProxyInterceptor bypasses it.
+    const proxyAgent = new ProxyAgent({
+      uri: proxyUrl.protocol + proxyUrl.host,
+      token,
+      interceptors: {
+        Client: [noProxyInterceptor],
+      },
+    });
+
+    // Make sure our configured proxy agent is used for all `fetch()` requests globally.
+    setGlobalDispatcher(proxyAgent);
+  }
+}
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -17,6 +17,27 @@ import {
   pluginIDProviderService,
   rbacDynamicPluginsProvider,
 } from './modules/rbacDynamicPluginsModule';
+import 'global-agent/bootstrap';
+import { setGlobalDispatcher, ProxyAgent } from 'undici';
+
+const proxyEnv =
+  process.env.GLOBAL_AGENT_HTTP_PROXY ||
+  process.env.GLOBAL_AGENT_HTTPS_PROXY;
+
+if (proxyEnv) {
+  const proxyUrl = new URL(proxyEnv);
+  setGlobalDispatcher(
+    new ProxyAgent({
+      uri: proxyUrl.protocol + proxyUrl.host,
+      token:
+        proxyUrl.username && proxyUrl.password
+          ? `Basic ${Buffer.from(
+              `${proxyUrl.username}:${proxyUrl.password}`,
+            ).toString('base64')}`
+          : undefined,
+    }),
+  );
+}
 
 const backend = createBackend();
 
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -17,8 +17,7 @@ import {
   pluginIDProviderService,
   rbacDynamicPluginsProvider,
 } from './modules/rbacDynamicPluginsModule';
-import { bootstrap } from 'global-agent';
-import { Agent, ProxyAgent, Dispatcher, setGlobalDispatcher } from 'undici';
+import { configureCorporateProxyAgent } from './corporate-proxy';
 
 // RHIDP-2217: adds support for corporate proxy
 configureCorporateProxyAgent();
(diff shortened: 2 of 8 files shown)


packages/backend/src/index.ts
```diff
@@ -99,3 +104,88 @@ backend.add(import('@internal/plugin-dynamic-plugins-info-backend'));
 backend.add(import('@internal/plugin-scalprum-backend'));
 
 backend.start();
+
+/**
+ * Adds support for corporate proxy to both 'node-fetch' (using 'global-agent') and native 'fetch' (using 'undici') packages.
+ */
+function configureCorporateProxyAgent() {
```
gazarenkov, line 111: Would not make sense to move this function from index.ts to simplify merging in case of upstream changes?

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,100 @@
...
+        value: 'http://10.10.10.105:3128'
+      - name: HTTPS_PROXY
+        value: 'http://10.10.10.106:3128'
+      - name: NO_PROXY
+        value: 'localhost,example.org'
+```
+
+</details>
+
+## Operator deployment
```
gazarenkov, line 48: I think it worth mention that it can be configured as default config on cluster level.
It looks like a better place, indeed, it looks like infrastructural specific and would be applicable for the whole cluster,.

gazarenkov (commented):
Any tests expected?

packages/backend/src/index.ts
rm3l: That makes sense - I'll move it.

rm3l:
> Any tests expected?

This will make sense as E2E tests, which are going to be covered in https://issues.redhat.com/browse/RHIDP-1955.

packages/backend/src/index.ts
rm3l: Done in [`96402f9` (#1225)](https://github.com/janus-idp/backstage-showcase/pull/1225/commits/96402f9284094527742190927f4e98404ef0015d)

showcase-docs/corporate-proxy.md
rm3l: Done in [`27b84b9` (#1225)](https://github.com/janus-idp/backstage-showcase/pull/1225/commits/27b84b9849a72fdcafb45a977ad659f0a926c8a8)

coreydaley:
@rm3l I think that the Local development and Full Test instructions that you wrote in the description for this pull request would be a great addition to a `docs/proxy.md` file as part of this pull request.

packages/backend/src/corporate-proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+      return (opts, handler) => {
+        return isNoProxy(opts.origin?.toString())
+          ? defaultAgent.dispatch(opts, handler)
+          : dispatch(opts, handler);
+      };
+    };
+
+    // Create a proxy agent that will send all requests through the configured proxy, unless the
+    // noProxyInterceptor bypasses it.
+    const proxyAgent = new ProxyAgent({
```
kim-tsao, line 92: I don't think we can assume all HTTP calls will go through the proxyAgent.  For example, customers can customize the homepage and support page with their own links: https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/customization.md.    If users click on those links in a proxied environment, they could be waiting a while for a response (300s seems to the be the default timeout).   We should document these areas where customers need to take some sort of action (remove links, update URL) and/or consider future support for shortening request timeouts

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,102 @@
+# Running the Showcase application behind a corporate proxy
+
+Out of the box, the Showcase application can be run behind a corporate proxy, by setting any of the following environment variables prior to starting the application:
+
+- `HTTP_PROXY`: HTTP proxy to use.
+- `HTTPS_PROXY`: distinct proxy to use for HTTPS requests.
+
+Additionally, you can set the `NO_PROXY` environment variable to exclude certain domains from proxying. The value is a comma-separated list of hostnames that do not require a proxy to get reached, even if one is specified.
+
```
kim-tsao, line 9: Would plugin vendors need to do anything here?  I assume as long as they are using one of `axios`, `fetch`, `node-fetch`, they should get this support for free?  Maybe it's worth calling out in our product docs that these are the only HTTP request clients we support.  In general, we should be following the upstream ADR that recommends the use of `node-fetch` but we don't enforce this.

packages/backend/src/corporate-proxy.ts
coreydaley: Do the links that customers can add get loaded through RHDH? If they open in a new tab then it has nothing to do with the proxy settings, it would have to do with their workstation/network proxy settings I would think.

kim-tsao: Scaffolder is another area that may need to be considered

packages/backend/src/corporate-proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+  // Bootstrap global-agent, which addresses node-fetch proxy-ing.
+  // global-agent purposely uses namespaced env vars to prevent conflicting behavior with other libraries,
+  // but user can set GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE to an empty value for global-agent to use
+  // the conventional HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables.
+  // More details in https://github.com/gajus/global-agent#what-is-the-reason-global-agentbootstrap-does-not-use-http_proxy
+  bootstrap();
+
+  // Configure the undici package, which affects the native 'fetch'. It leverages the same env vars used by global-agent,
+  // or the more conventional HTTP(S)_PROXY ones.
+  const proxyEnv =
```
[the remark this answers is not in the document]
rm3l: Yes, they are processed based on the protocol, hostname and port. The check here is only to determine whether to create the global dispatcher, as long as any of these env vars is set. So the order of precedence shouldn't matter, I think.
However, what can be confusing is the order of precedence between, say, `HTTP_PROXY` and `http_proxy` if both are set. I couldn't find a clear answer to this, as it turns out that it depends on how each library handles that. I documented setting the upper case form because that's what both 'global-agent` honors (and Axios as well).

packages/backend/src/corporate-proxy.ts
rm3l: > I don't think we can assume all HTTP calls will go through the proxyAgent. For example, customers can customize the homepage and support page with their own links: [main/showcase-docs/customization.md](https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/customization.md?rgh-link-date=2024-05-13T18%3A07%3A45Z). If users click on those links in a proxied environment, they could be waiting a while for a response (300s seems to the be the default timeout). We should document these areas where customers need to take some sort of action (remove links, update URL) and/or consider future support for shortening request timeouts

These kinds of customized links will be accessed directly from the browser, no?
The settings here cover requests sent from the backend part of the application. As mentioned by @coreydaley above, client-side calls will typically be covered by browser/OS settings. I assume if their workstations are behind a corporate proxy, they already have instructions for configuring their browsers and systems accordingly.

showcase-docs/corporate-proxy.md
rm3l: There is indeed nothing to do for plugin vendors using Axios, `fetch` or `node-fetch`. I'll mention this in the docs, and make sure it is also called out in our product docs. Thanks.

packages/backend/src/corporate-proxy.ts
rm3l: > Scaffolder is another area that may need to be considered

I just tested with some scaffolding templates (like https://github.com/redhat-developer/red-hat-developer-hub-software-templates/blob/main/templates/github/go-backend/template.yaml or https://github.com/backstage/software-templates/blob/main/scaffolder-templates/springboot-grpc-template/template.yaml), and could see the backend requests sent to the proxy, as expected.

rm3l:
> @rm3l I think that the Local development and Full Test instructions that you wrote in the description for this pull request would be a great addition to a `docs/proxy.md` file as part of this pull request.

Good point. Added in [`bcf8551` (#1225)](https://github.com/janus-idp/backstage-showcase/pull/1225/commits/bcf85510875f048fb91f954b3859e244a6f4ddc4). Thanks!

showcase-docs/corporate-proxy.md
rm3l: Added in [`66a5d93` (#1225)](https://github.com/janus-idp/backstage-showcase/pull/1225/commits/66a5d930aaecf5b0ea2440a20531e19aeb63cfa2)

rm3l:
Rebased and force-pushed to fix a conflict with `yarn.lock`

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
+# Running the Showcase application behind a corporate proxy
+
+Out of the box, the Showcase application can be run behind a corporate proxy, by setting any of the following environment variables prior to starting the application:
+
+- `HTTP_PROXY`: HTTP proxy to use.
+- `HTTPS_PROXY`: distinct proxy to use for HTTPS requests.
```
coreydaley, line 6:
```suggestion
- `HTTP_PROXY`: Proxy to use for HTTP requests.
- `HTTPS_PROXY`: Proxy to use for HTTPS requests.
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+
+You can set the proxy information in your Helm `values` file, like so:
+
+```yaml
+upstream:
+  backstage:
+    extraEnvVars:
+      - name: HTTP_PROXY
+        # HTTP proxy to use
+        value: '<my_http_proxy_url>'
```
coreydaley, line 20:
```suggestion
      # Proxy to use for HTTP requests
      - name: HTTP_PROXY
        value: '<my_http_proxy_url>'
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+      - name: HTTPS_PROXY
+        # Distinct proxy to use for HTTPS requests
+        value: '<my_https_proxy_url>'
```
coreydaley, line 23:
```suggestion
      # Proxy to use for HTTPS requests
      - name: HTTPS_PROXY
        value: '<my_https_proxy_url>'
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+You can set the proxy information in your Custom Resource, like so:
+
+```yaml
+spec:
+  application:
+    extraEnvs:
+      envs:
+        - name: HTTP_PROXY
+          # HTTP proxy to use
+          value: '<my_http_proxy_url>'
```
coreydaley, line 59:
```suggestion
        # Proxy to use for HTTP requests
        - name: HTTP_PROXY
          value: '<my_http_proxy_url>'
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+        - name: HTTPS_PROXY
+          # Distinct proxy to use for HTTPS requests
+          value: '<my_https_proxy_url>'
```
coreydaley, line 62:
```suggestion
        # Proxy to use for HTTPS requests
        - name: HTTPS_PROXY
          value: '<my_https_proxy_url>'
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+  backstage:
+    extraEnvVars:
+      - name: HTTP_PROXY
+        # HTTP proxy to use
+        value: '<my_http_proxy_url>'
+      - name: HTTPS_PROXY
+        # Distinct proxy to use for HTTPS requests
+        value: '<my_https_proxy_url>'
+      - name: NO_PROXY
+        # Pattern of comma-separated URLs that should be excluded from proxying.
```
coreydaley, line 25:
```suggestion
        # List of comma-separated URLs that should be excluded from proxying.
```

showcase-docs/corporate-proxy.md
```diff
@@ -0,0 +1,88 @@
...
+    extraEnvs:
+      envs:
+        - name: HTTP_PROXY
+          # HTTP proxy to use
+          value: '<my_http_proxy_url>'
+        - name: HTTPS_PROXY
+          # Distinct proxy to use for HTTPS requests
+          value: '<my_https_proxy_url>'
+        - name: NO_PROXY
+          # Pattern of comma-separated URLs that should be excluded from proxying.
```
coreydaley, line 64:
```suggestion
          # List of comma-separated URLs that should be excluded from proxying.
```

coreydaley (commented):
Just a few suggestions.

rm3l:
Rebased and force-pushed to fix a conflict with `yarn.lock`.

rm3l:
Rebased and force-pushed to fix a conflict with the following files:
- `.rhdh/docker/Dockerfile`
- `docker/Dockerfile`

PR #1245, opened 2024-05-08
fix(app): RHIDP-2266 add condition error handling

gashcrumb:
## Description

This change adds exception handling to the two points where plugin supplied condition functions are evaluated, once when rendering the catalog entity tabs, and a second time when a plugin's mountpoint configuration is being checked.  This change also makes the ApiHolder context object available as a 2nd argument; available at the second evaluation point.

- Fixes [RHIDP-2266](https://issues.redhat.com/browse/RHIDP-2266)

## PR acceptance criteria

An easy way to test this is to throw an exception from [here](https://github.com/gashcrumb/simple-test-components/blob/main/src/plugin.ts#L35) and use [this](https://github.com/gashcrumb/simple-test-components/tree/main#copy-build-results-to-host-application) plugin.  The configuration for this would look like:

```yaml
    backstage-plugin-simple-test-components:
      mountPoints:
        - mountPoint: entity.page.overview/cards
          importName: SimpleTestComponentsPage
          config:
            layout:
              gridColumn: "span 1"
              gridRow: "span 1"
            if:
              allOf:
                - isAvailableYes
            props:
              text: 'Content Block Two (always available)'
```

### DIFF
--- a/packages/app/src/components/DynamicRoot/DynamicRootContext.tsx
+++ b/packages/app/src/components/DynamicRoot/DynamicRootContext.tsx
@@ -36,7 +36,7 @@ type ScalprumMountPointConfigBase = {
 };
 
 export type ScalprumMountPointConfig = ScalprumMountPointConfigBase & {
-  if: (e: Entity) => boolean | Promise<boolean>;
+  if: (e: Entity) => boolean;
 };
 
 export type ScalprumMountPointConfigRawIf = {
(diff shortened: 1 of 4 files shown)


ciiay:
Verified this is working great. I have this config
```
          if:
            allOf:
            - isAvailableUsingContext
```
As well as an exception in `simple-test-components/src/plugin.ts`
```
export const isAvailableUsingContext = (
  _entity: Entity,
  _context: { apis: ApiHolder }
) => {
  throw new Error("Test exception for isAvailableUsingContext");
};
```
When tested on main branch it throws an error and breaks the entire UI. While with this PR's update it loads the rest of catalog entity tabs only the simple-test-component doesn't get load.

gashcrumb:
github happy path tests seem to have:

image

PR #1248, opened 2024-05-09
Fixing rbac tests execution

gustavolira:
## Description

Update openshift-ci-tests.sh to echo RESULT

Changed the behavior in openshift-ci-tests.sh to echo the result instead of exiting the script using the RESULT variable. This modification allows further actions or analysis after the droute_send function call.

RBAC tests execution

## PR acceptance criteria


.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -225,11 +225,11 @@ run_tests() {
   ansi2html <"/tmp/${LOGFILE}" >"/tmp/${LOGFILE}.html"
   cp -a "/tmp/${LOGFILE}.html" ${ARTIFACT_DIR}/${project}
   cp -a /tmp/backstage-showcase/e2e-tests/playwright-report/* ${ARTIFACT_DIR}/${project}
-  
+
 
   droute_send $release_name $project
 
-  exit ${RESULT}
+  echo ${RESULT}
```
subhashkhileri, line 232: We should return and exit as per the result else it will not mark job as failed because of `set -e` while running tests.

gustavolira: @subhashkhileri ok it's just for test purpose, let's see if rbac tests will be executed

gustavolira: @subhashkhileri do you have any suggestion or can we go with `echo` ?

PR #1306, opened 2024-06-06
[e2e] Verify TLS config with external postgres-db

subhashkhileri:
## Description
[e2e] Verify TLS config with external postgres-db
- Set up an RHDH deployment with Postgres DB installed in a separate namespace
- Configure the certs (TBD, determine how to provision CA certs) by following the comments in the above config file
- Verify the connection between RHDH and the DB is working

https://issues.redhat.com/browse/RHIDP-1237

## PR acceptance criteria


.ibm/pipelines/resources/postgres-db/postgres.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: postgress-external-db
+  namespace: postgress-external-db
+spec:
+  image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi8-14.2-1
+  postgresVersion: 14
```
josephca, line 8: @subhashkhileri Can we change it to use version 15.1? It would be good to use the same postgres version that RHDH 1.2 supports

.ibm/pipelines/resources/postgres-db/postgres-cred.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: postgres-cred
+data:
+  POSTGRES_PASSWORD: dG1w
+  POSTGRES_PORT: NTQzMg==
+  POSTGRES_USER: amFudXMtaWRw
+  POSTGRES_HOST: cG9zdGdyZXNzLWV4dGVybmFsLWRiLXByaW1hcnkucG9zdGdyZXNzLWV4dGVybmFsLWRiLnN2Yy5jbHVzdGVyLmxvY2Fs
+  PGSSLMODE: cmVxdWlyZQ==
```
josephca, line 10: @subhashkhileri Can we add an extra line to avoid red x?

josephca (changes requested):

.ibm/pipelines/resources/postgres-db/postgres.yaml
rnapoles-rh, line 8: Same as Joseph, lets use v16

rnapoles-rh (changes requested):
lets use pg v16

.ibm/pipelines/value_files/values_showcase-rbac.yaml
```diff
@@ -151,9 +151,19 @@ upstream:
...
-          connection:
-            password: ${POSTGRESQL_ADMIN_PASSWORD}
-            user: postgres
+          connection:  # configure Backstage DB connection parameters
+            host: ${POSTGRES_HOST}
+            port: ${POSTGRES_PORT}
+            user: ${POSTGRES_USER}
+            password: ${POSTGRES_PASSWORD}
+            ssl:
+              rejectUnauthorized: false,
```
gazarenkov, line 160: What if you set rejectUnauthorized: true ?

subhashkhileri: https://github.com/backstage/charts/issues/80#issuecomment-1501776357

`rejectUnauthorized: true` also works. but with false it will tell not to fail if the connection isn't verified.

I have update the `rejectUnauthorized` to true.

.ibm/pipelines/resources/postgres-db/postgres.yaml
subhashkhileri: Updated to version 16.

subhashkhileri: Updated the version to 16

PR #1332, opened 2024-06-17
Updated the label and container port in doc on Setting up Metrics Mon…

Fortune-Ndlovu:
- Replaced janus-idp.io/app with rhdh.redhat.com/app
- Replaced the container port name from http-backend to backend
- Made sure the instructions provided work correctly

Why the documentation update?
In the ServiceMonitor resource to create, the label added by the Operator had changed from janus-idp.io/app to rhdh.redhat.com/app a long time ago [(https://github.com/janus-idp/operator/issues/199)](url) before the first version of the operator was released, and we forgot to update this documentation.

Also, the container port endpoint to scrape metrics from is wrong. It should be backend as per the 1.1.x, 1.2.x, or main branches in the operator repo.

As a consequence, the instructions depicted in this [https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/monitoring-and-logging.md#operator-backed-deployment](url) will not work as expected.


rm3l (approved):
Awesome - thanks @Fortune-Ndlovu for fixing this and making sure it works as expected !!

rm3l:
/approve
/override "PR Docker Build"

Doc-only PR

rm3l:
@coreydaley (or anyone else) Can you `/approve` this PR so it can be merged? Looks like I don't have permission to do so. I confirm that the change to the docs looks good.

nickboldt:
instead of all the comments to disable tests, why not just merge this directly? seems more efficient...

rm3l:
> instead of all the comments to disable tests, why not just merge this directly? seems more efficient...

Indeed, I would have merged this simple PR directly, but I don't have permission on this repo.

PR #1429, opened 2024-07-19
provide kubernetesAuthApi in frontend kubernetes dynamic plugin

JessicaJHee:
## Description

The Kubernetes plugin has a default API factory that work right out the box with a few oidcProviders as defined [here](https://github.com/backstage/backstage/blob/d2af76ffb6a105dfec1310733f5ca8e316872116/plugins/kubernetes/src/plugin.ts#L66-L96), but `oidc.oidc` is not included. We are able to override this definition by adding our own implementation. This PR adds an additional utility API to the `backstage-plugin-kubernetes` plugin to allow the use of `oidcTokenProvider: 'oidc'` ([More on `oidcTokenProvider`](https://backstage.io/docs/features/kubernetes/configuration/#clustersoidctokenprovider-optional)).

- Fixes [RHDHBUGS-36](https://issues.redhat.com/browse/RHDHBUGS-36)

## PR acceptance criteria
The error does not appear when configuring OIDC as the auth provider with the Kubernetes plugin like: 
```
kubernetes:
  ...
  clusterLocatorMethods:
    - type: 'config'
      clusters:
          ...
          authProvider: 'oidc'
          oidcTokenProvider: 'oidc'
```
[Note](https://backstage.io/docs/features/kubernetes/configuration/#clustersoidctokenprovider-optional) on `oidcTokenProvider`

image

1. Set up oidc authentication in `auth.providers.oidc`
2. Add annotation to a component to [surface the Kubernetes components as part of an entity](https://backstage.io/docs/features/kubernetes/configuration/#surfacing-your-kubernetes-components-as-part-of-an-entity)
3. Configure the Kubernetes plugins with the following auth and confirm that no error occurs.
``` 
         authProvider: 'oidc'
         oidcTokenProvider: 'oidc'
```


kim-tsao (commented):
The changes look fine but the only drawback with overriding is that we need to be aware of any changes upstream in order to keep in sync.  Since this is modification of existing behaviour, we should document these changes midstream https://github.com/janus-idp/backstage-showcase/tree/main/showcase-docs which product docs can then use

JessicaJHee:
> The changes look fine but the only drawback with overriding is that we need to be aware of any changes upstream in order to keep in sync. Since this is modification of existing behaviour, we should document these changes midstream https://github.com/janus-idp/backstage-showcase/tree/main/showcase-docs which product docs can then use

@kim-tsao Since there is no section for specific plugins, would it make sense to add a note about this change under [this section](https://github.com/janus-idp/backstage-showcase/blob/46f828a9676d38702300ab2687f9d4e28fb945d5/showcase-docs/getting-started.md#running-locally-with-the-optional-plugins)? I've added it there for now, lmk how it looks!

showcase-docs/getting-started.md
```diff
@@ -341,6 +341,10 @@ The easiest and fastest method for getting started: Backstage Showcase app, runn
 
 7. Navigate to <http://localhost:3000>
 
+#### Note: Extension to Kubernetes Backend Plugin
+
+The Kubernetes plugin has a default API factory that work right out the box with a few `oidcProviders` for authentication, but `oidc.oidc` is not included. An additional utility API has been added to the `backstage-plugin-kubernetes` plugin to allow the use of `oidcTokenProvider: 'oidc'` ([More on oidcTokenProvider](https://backstage.io/docs/features/kubernetes/configuration/#clustersoidctokenprovider-optional)).
+
```
kim-tsao, line 347: Ah, we just need to be careful providing links to upstream.  Docs cannot use them as reference.  Can you inline the config snippet as well?

JessicaJHee: I've made the changes, lmk how it looks!

PR #1472, opened 2024-08-08
RHIDP-2128 [Test automation] Establish AKS nightly tests 

zdrapela:
## Description

This PR is making it possible to test RHDH installation on AKS. E2E tests are limited to those that run immediately without further configuration of the cluster (skipped tests defined in: [`e2e-tests/playwright.config.ts`](https://github.com/janus-idp/backstage-showcase/pull/1472/files#diff-8b56f4853df94dcbddfa2770d8b30a8114413e57be15090c3b1534ca57839364)). GitHub would be working, but we reached the maximum number of callback URLs.

To distinguish which configuration should be applied, the `if [[ "$JOB_NAME" == *aks* ]]` condition is used in `.ibm/pipelines/openshift-ci-tests.sh`. 

It uses different value file for AKS instance: `.ibm/pipelines/value_files/values_showcase_AKS.yaml`

This PR adds the Azure CLI to the Dockerfile. The AKS test uses it to start the cluster and then to stop it after the test finishes. The account for Azure CLI was established using https://github.com/adrianriobo/mapt/blob/fix-272/hacks/azure/azure_setup.sh

- Fixes [RHIDP-2128](https://issues.redhat.com/browse/RHIDP-2128)

## PR acceptance criteria


e2e-tests/package.json
```diff
@@ -14,6 +14,7 @@
     "showcase-rbac-1-2-x": "playwright test --project=showcase-rbac",
     "showcase-ci-nightly": "playwright test --project=showcase",
     "showcase-rbac-nightly": "playwright test --project=showcase-rbac",
+    "showcase-aks-ci-nightly": "playwright test --project=showcase-aks",
```
subhashkhileri, line 17: We can use existing project : `showcase-ci-nightly` and `showcase-rbac-nightly`

subhashkhileri, line 17: We can use existing `showcase-ci-nightly` and `showcase-rbac-nightly`

e2e-tests/playwright.config.ts
```diff
@@ -62,6 +62,25 @@ export default defineConfig({
         '**/playwright/e2e/verify-tls-config-with-external-postgres-db.spec.ts',
       ],
     },
+    {
+      name: 'showcase-aks',
```
subhashkhileri, line 66: any specific idea to have new project? I guess we want to run all tests on AKS.

.ibm/pipelines/value_files/values_showcase_AKS.yaml
subhashkhileri, line 1: I guess we are good with using existing value files for rbac and non-rbac. creating new value files will make things complex in future.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -360,23 +398,32 @@ main() {
...
 
   install_oc
-  oc login --token="${K8S_CLUSTER_TOKEN}" --server="${K8S_CLUSTER_URL}"
+  oc login --token="${K8S_CLUSTER_TOKEN}" --server="${K8S_CLUSTER_URL}" --insecure-skip-tls-verify
   echo "OCP version: $(oc version)"
 
   API_SERVER_URL=$(oc whoami --show-server)
-  K8S_CLUSTER_ROUTER_BASE=$(oc get route console -n openshift-console -o=jsonpath='{.spec.host}' | sed 's/^[^.]*\.//')
+  if [[ "$JOB_NAME" == *aks* ]]; then
+    K8S_CLUSTER_ROUTER_BASE=$(kubectl get svc nginx --namespace app-routing-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
subhashkhileri, line 410: what if there is no nginx running, I would fail here.

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -360,23 +398,32 @@ main() {
     NAME_SPACE="showcase-ci-nightly"
     NAME_SPACE_RBAC="showcase-rbac-nightly"
     NAME_SPACE_POSTGRES_DB="postgress-external-db-nightly"
+    NAME_SPACE_AKS="showcase-aks-ci-nightly"
```
subhashkhileri, line 401: new namespace is not needed, I guess idea is to deploy  existing namespace `showcase-ci-nightly` and `showcase-rbac-nightly` on AKS cluster.

.ibm/pipelines/openshift-ci-tests.sh
zdrapela: I agree that it would be better to make sure it runs and enable it if it's not. As it is now, enabling it would be a [manual step](https://redhat-developer.github.io/red-hat-developers-documentation-rhdh/main/install-rhdh-aks/#proc-rhdh-deploy-aks-helm_assembly-install-rhdh-aks) for setting up the cluster before using it for running tests.

e2e-tests/playwright.config.ts
zdrapela: I've got a two reasons for that:
- we can start running tests on AKS before we do the manual setup that is needed for running all the tests ([RHIDP-2128](https://issues.redhat.com/browse/RHIDP-2128) demands installation and Azure integration tests only)
- I believe that some tests might break only on AKS and it is good to have an option to disable them only for AKS

subhashkhileri: Agree, lets keep this way then.

.ibm/pipelines/value_files/values_showcase_AKS.yaml
zdrapela: The config for AKS is slightly different. We might use the same file, but we will have to change it through the Helm install. I agree that maintaining two value files is worse than one. It is not that much values so it can be possible to change the values through the Helm install for AKS and use just one values file (well, one for non-RBAC, one for RBAC).

Screenshot 2024-08-09 at 10 36 52

.ibm/pipelines/openshift-ci-tests.sh
zdrapela: Right now, project that yarn runs the playwright tests for depends on the namespace name. So if we want different project for AKS, we need a different namespace, or change the script to use different value for that
https://github.com/janus-idp/backstage-showcase/blob/5b48b6fadb24b0297015e9c578add6341337f011/e2e-tests/package.json#L17
https://github.com/janus-idp/backstage-showcase/blob/5b48b6fadb24b0297015e9c578add6341337f011/.ibm/pipelines/openshift-ci-tests.sh#L258

.ibm/pipelines/value_files/values_showcase_AKS.yaml
zdrapela: I added `diff-values_showcase_AKS.yaml` file instead and use the `values_showcase.yaml` as a base. Both of the files are applied by `helm upgrade -i` https://github.com/janus-idp/backstage-showcase/pull/1472/files#diff-4d6549ba4c59a94aab379c14838198870ef2e610c26348b6c58c6fbffedc479bR372

.ibm/pipelines/openshift-ci-tests.sh
zdrapela: I added this to make sure it is enabled: https://github.com/janus-idp/backstage-showcase/pull/1472/files#diff-4d6549ba4c59a94aab379c14838198870ef2e610c26348b6c58c6fbffedc479bR392

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -149,14 +155,16 @@ apply_yaml_files() {
...
+    oc new-app https://github.com/janus-qe/test-backstage-customization-provider --namespace="${project}"
+    oc expose svc/test-backstage-customization-provider --namespace="${project}"
+  fi
   oc apply -f "$dir/resources/cluster_role/cluster-role-k8s.yaml" --namespace="${project}"
   oc apply -f "$dir/resources/cluster_role_binding/cluster-role-binding-k8s.yaml" --namespace="${project}"
   oc apply -f "$dir/resources/cluster_role/cluster-role-ocm.yaml" --namespace="${project}"
   oc apply -f "$dir/resources/cluster_role_binding/cluster-role-binding-ocm.yaml" --namespace="${project}"
 
-  sed -i "s/K8S_CLUSTER_API_SERVER_URL:.*/K8S_CLUSTER_API_SERVER_URL: ${ENCODED_API_SERVER_URL}/g" "$dir/auth/secrets-rhdh-secrets.yaml"
+  #sed -i "s/K8S_CLUSTER_API_SERVER_URL:.*/K8S_CLUSTER_API_SERVER_URL: ${ENCODED_API_SERVER_URL}/g" "$dir/auth/secrets-rhdh-secrets.yaml"
```
subhashkhileri, line 167: if its no longer needed can we remove it?

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -170,15 +178,19 @@ apply_yaml_files() {
...
+    oc apply -f "$dir/resources/pipeline-run/hello-world-pipeline-run.yaml"
+  fi
 }
 
 droute_send() {
   # Skipping ReportPortal for nightly jobs on OCP v4.14 and v4.13 for now, as new clusters are not behind the RH VPN.
   if [[ "$JOB_NAME" == *ocp-v4* ]]; then
     return 0
+  elif [[ "$JOB_NAME" == *aks* ]]; then
+    return 0
```
subhashkhileri, line 193:
```suggestion
  if [[ "$JOB_NAME" == *ocp-v4* || "$JOB_NAME" == *aks* ]]; then
    return 0
```

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -364,23 +425,37 @@ main() {
...
   echo "K8S_CLUSTER_ROUTER_BASE : $K8S_CLUSTER_ROUTER_BASE"
 
   ENCODED_API_SERVER_URL=$(echo "${API_SERVER_URL}" | base64)
   ENCODED_CLUSTER_NAME=$(echo "my-cluster" | base64)
 
   initiate_deployments
-  check_and_test "${RELEASE_NAME}" "${NAME_SPACE}"
-  check_and_test "${RELEASE_NAME_RBAC}" "${NAME_SPACE_RBAC}"
+  if [[ "$JOB_NAME" == *aks* ]]; then
+    check_and_test "${RELEASE_NAME}" "${NAME_SPACE_AKS}"
```
subhashkhileri, line 454: I am not seeing the rbac tests here : `check_and_test "${RELEASE_NAME_RBAC}" "${NAME_SPACE_RBAC}"`

.ibm/pipelines/value_files/diff-values_showcase_AKS.yaml
subhashkhileri, line 1: why do we have diff in this file name? 🤔

.ibm/pipelines/value_files/values_showcase.yaml
```diff
@@ -111,6 +111,8 @@ global:
           disabled: false
         - package: ./dynamic-plugins/dist/janus-idp-backstage-plugin-topology
           disabled: false
+        - package: './dynamic-plugins/dist/janus-idp-backstage-plugin-analytics-provider-segment'
+          disabled: true
```
subhashkhileri, line 115: this plugin seems to be disabled 🤔

.ibm/pipelines/openshift-ci-tests.sh
zdrapela: It was disabled for debugging and forgotten. I uncommented it - good catch!

.ibm/pipelines/value_files/diff-values_showcase_AKS.yaml
zdrapela: I wanted to highlight, that it is not a standalone file with all the values needed, but instead, it is a file that contains differences against the main `values_showcase.yaml` value file.

.ibm/pipelines/openshift-ci-tests.sh
zdrapela: I initially wanted to enable just non-RBAC tests (the requirement was just for installation on AKS) and skip RBAC.
But I will try to enable the RBAC ones also (if I don't encounter any issue) and let you know.

PR #1469, opened 2024-08-07
WIP feat: add support for plugins packaged in oci images

kadel:
## Description
Adds support for loading plugins packaged inside OCI images

config example
```yaml
    plugins:
      - disabled: false
        package: >-
          oci://quay.io/tkral/simple-chat:v0.0.1!internal-backstage-plugin-simple-chat-backend-dynamic
      - disabled: false
        package: >-
          oci://quay.io/tkral/simple-chat:v0.0.1!internal-backstage-plugin-simple-chat
        pluginConfig:
          dynamicPlugins:
            frontend:
              internal.backstage-plugin-simple-chat:
                appIcons:
                  - importName: ChatIcon
                    name: chatIcon
                dynamicRoutes:
                  - importName: SimpleChatPage
                    menuItem:
                      icon: chatIcon
                      text: Simple Chat
                    path: /simple-chat
```

## Which issue(s) does this PR fix

- Fixes #?

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

PR #1479, opened 2024-08-13
add install method of plugin from container registry

durandom:
## Description
Adds support for loading plugins packaged inside OCI images.

Rework of #1469 

Instead of the oras python library, which does not work out of the box with ghcr or docker registries (due to authentication and requesting temporary tokens for auth), this PR uses `skopeo`

The image is downloaded only once, even if it contains multiple plugins.
We're still assuming there's only one layer in the image and only inspect the first layer.

config example
```yaml
    plugins:
      - disabled: false
        package: >-
          oci://quay.io/tkral/simple-chat:v0.0.1!internal-backstage-plugin-simple-chat-backend-dynamic
      - disabled: false
        package: >-
          oci://quay.io/tkral/simple-chat:v0.0.1!internal-backstage-plugin-simple-chat
        pluginConfig:
          dynamicPlugins:
            frontend:
              internal.backstage-plugin-simple-chat:
                appIcons:
                  - importName: ChatIcon
                    name: chatIcon
                dynamicRoutes:
                  - importName: SimpleChatPage
                    menuItem:
                      icon: chatIcon
                      text: Simple Chat
                    path: /simple-chat
```

## related PRs
 
https://github.com/redhat-developer/rhdh-chart/pull/41  - auth.json volume mount in helm chart

Signed-off-by: Marcel Hild <hild@b4mad.net>

### DIFF
--- a/.rhdh/docker/Dockerfile
+++ b/.rhdh/docker/Dockerfile
@@ -191,7 +191,7 @@ WORKDIR $CONTAINER_SOURCE/
 COPY $REMOTE_SOURCES/upstream2 $REMOTE_SOURCES_DIR/upstream2/
 # hadolint ignore=DL3013,DL3041,SC2086
 RUN microdnf update --setopt=install_weak_deps=0 -y && \
-  microdnf install -y python3.11 python3.11-pip python3.11-devel make cmake cpp gcc gcc-c++; \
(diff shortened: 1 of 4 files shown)


nickboldt (commented):
excited to have this land in 1.3 so we can start testing its integration with the new registry container

did not review the code (my golang-fu is weak), but happy with the idea.

PR #1523, opened 2024-08-22
Fixing e2e tests

gustavolira:
## Description

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
+++ b/e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
@@ -44,9 +44,9 @@ test.describe.serial('Link Scaffolded Templates to Catalog Items', () => {
   test('Create a React App using the newly registered Template', async () => {
     await uiHelper.openSidebar('Catalog');
     await uiHelper.clickButton('Create');
-    await uiHelper.searchInputPlaceholder('Create React');
-
+    await uiHelper.searchInputPlaceholder('Create React App Template');
     await uiHelper.verifyText('Create React App Template');
+    await uiHelper.waitForTextDisappear('Add ArgoCD to an existing project');
     await uiHelper.clickButton('Choose');
 
     await uiHelper.fillTextInputByLabel('Name', reactAppDetails.componentName);
@@ -59,6 +59,7 @@ test.describe.serial('Link Scaffolded Templates to Catalog Items', () => {
 
     await uiHelper.fillTextInputByLabel('Owner', reactAppDetails.repoOwner);
     await uiHelper.fillTextInputByLabel('Repository', reactAppDetails.repo);
+    await uiHelper.pressTab();
     await uiHelper.clickButton('Review');
 
     await uiHelper.verifyRowInTableByUniqueText('Owner', [
--- a/e2e-tests/playwright/support/pages/rbac.ts
+++ b/e2e-tests/playwright/support/pages/rbac.ts
@@ -37,8 +37,14 @@ export class Roles {
     return [policies];
   }
 
+  //Depending on the version of the Backstage, it can be 'Permission Policies' or 'Accessible Plugins'
+  // Accepts either term
   static getRolesListColumnsText() {
-    return ['Name', 'Users and groups', 'Permission Policies', 'Actions'];
+    return [
+      /^Name$/,
+      /^Users and groups$/,
+      /Permission Policies|Accessible plugins/,
+    ];
   }
 
   static getUsersAndGroupsListColumnsText() {
--- a/e2e-tests/playwright/utils/UIhelper.ts
+++ b/e2e-tests/playwright/utils/UIhelper.ts
@@ -22,6 +22,10 @@ export class UIhelper {
     await this.page.fill('input[placeholder="Search"]', searchText);
   }
 
+  async pressTab() {
+    await this.page.keyboard.press('Tab');
+  }
+
   async waitForHeaderTitle() {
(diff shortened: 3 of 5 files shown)


nickboldt:
I guess we can link this PR to https://issues.redhat.com/browse/RHIDP-2793

josephca (approved):
/lgtm
/approve

e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
```diff
@@ -44,8 +44,8 @@ test.describe.serial('Link Scaffolded Templates to Catalog Items', () => {
   test('Create a React App using the newly registered Template', async () => {
     await uiHelper.openSidebar('Catalog');
     await uiHelper.clickButton('Create');
-    await uiHelper.searchInputPlaceholder('Create React');
-
+    await uiHelper.searchInputPlaceholder('Create React App Template');
+    await page.waitForLoadState('networkidle');
```
subhashkhileri, line 48: @gustavolira can we use alternative to `networkidle`, its not recommended, we aren't using it in our tests as it makes test more flaky.

e2e-tests/playwright/support/pages/rbac.ts
```diff
@@ -37,8 +37,15 @@ export class Roles {
...
   }
 
+  //Depending on the version of the Backstage, it can be 'Permission Policies' or 'Accessible Plugins'
+  // Accepts either term
   static getRolesListColumnsText() {
-    return ['Name', 'Users and groups', 'Permission Policies', 'Actions'];
+    return [
+      'Name',
+      'Users and groups',
+      ['Permission Policies', 'Accessible plugins'],
```
subhashkhileri, line 46: its not affecting 1.2.x, can we keep only the latest version for 1.3 and avoid these nested columns?

e2e-tests/playwright/utils/UIhelper.ts
```diff
@@ -176,25 +180,54 @@ export class UIhelper {
...
-        .locator(`tr>th`)
-        .getByText(rowText, { exact: exact })
-        .first();
-      await rowLocator.waitFor({ state: 'visible' });
-      await rowLocator.scrollIntoViewIfNeeded();
-      await expect(rowLocator).toBeVisible();
+      if (Array.isArray(rowText)) {
+        let found = false;
+
+        for (const text of rowText) {
```
subhashkhileri, line 190: Nested loop is not needed, you can pass OR operator in regex to verify either of it for eg : /Permission Policies|Accessible plugins/

e2e-tests/playwright/utils/UIhelper.ts
```diff
@@ -176,25 +180,54 @@ export class UIhelper {
...
     }
   }
 
-  async verifyHeading(heading: string) {
+  async verifyHeading(heading: string, exact: boolean = true) {
     const headingLocator = this.page
       .locator('h1, h2, h3, h4, h5, h6')
-      .filter({ hasText: heading })
-      .first();
+      .getByText(heading, { exact });
```
subhashkhileri, line 229: with a same text there can be multiple Headings, lets keep `.first();` to make test less flaky.

e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
gustavolira: @subhashkhileri The only way I was able to make this work was by using networkidle or waitForTimeout(1000). So, comparing those two options, I think networkidle is the lesser of two evils

e2e-tests/playwright/support/pages/rbac.ts
gustavolira: @subhashkhileri  It will not break changes for 1.2 since I updated the `verifyColumnHeading`

e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
subhashkhileri: Well in some/many scenarios `networkidle` would never happen as application continuously keeps making new requests, it can be token refresh or other things.

For eg. we have faced this issue post github login and test used to fail due to timeout errors

e2e-tests/playwright/support/pages/rbac.ts
subhashkhileri: @gustavolira what I mean is we can keep only the latest column name and remove the old one instead of checking either of them.

e2e-tests/playwright/e2e/catalog-scaffoldedfromLink.spec.ts
gustavolira: Do you have any other suggestion, because I could only make it work with networkidle or waitForTimeout(1000)

subhashkhileri: I've tried running it a couple of times, and it works fine without using `networkidle` or a timeout. Since there is already an `await uiHelper.verifyText('Create React App Template');`, it should automatically wait for the card to be visible.

gustavolira: The implicit wait for await uiHelper.verifyText('Create React App Template'); won't work because the title 'Create React App Template' is always present, even without performing the search. The purpose of the search is to narrow down the results to a single item, allowing for the correct selection of the Choose button.

PR #1576, opened 2024-09-09
[e2e] Switch `droute` pod for RP and update

zdrapela:
## Description

- Switch to a better-configured pod to send test results to ReportPortal.
- Update `droute` to `1.2`
- Use new `droute` functionality to finalize the status in ReportPortal automatically.

- Fixes [RHIDP-4011](https://issues.redhat.com/browse/RHIDP-4011)

## PR acceptance criteria


subhashkhileri:
@zdrapela I was thinking if we should add `set +e` at start of `droute_send()` and `set -e` at the end? so in case of droute failure it should not stop the rbac tests from running and also it will not mark e2e-tests as failure just because droute failed to upload result.

zdrapela:
@subhashkhileri Yes, I agree, droute should have a lower priority. The drawback is, that we might miss when droute gets broken.

subhashkhileri:
> @subhashkhileri Yes, I agree, droute should have a lower priority when it comes to marking tests as failed. The drawback is, that we might miss when droute gets broken.

Okay, you are adding rsync, does this lowers chances of failure? if yes then we can go with your changes, else we need to think on adding `set +e`

zdrapela:
@subhashkhileri Yes, now the pod uses an image with rsync and rh certs baked in. That should be much more reliable.

Let me also use this PR to update the `droute` to `1.2`

subhashkhileri:
@zdrapela can you please move droute to utils.sh

zdrapela:
Sure! Done and rebased

josephca (commented):
/lgtm
/approve

subhashkhileri:
@zdrapela It will need a manual cherry-pick. in 1.3.x droute is set +e, you can make appropriate changes and also remove the cluster redirect change that made as pr-os is working fine now.

zdrapela:
Ok, opened PR here https://github.com/janus-idp/backstage-showcase/pull/1657

PR #1565, opened 2024-09-06
Add RBAC file watcher and alias tests

josephca:
## Description

Add e2e tests for new RBAC features

- load RBAC policy and conditional permission  from files 
- test aliases : $currentUser, $ownerRefs 

## Which issue(s) does this PR fix

- Fixes [RHIDP-3482](https://issues.redhat.com/browse/RHIDP-3482)

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

PR #1596, opened 2024-09-12
[e2e] Fix failure of `catalog-timestamp`

zdrapela:
## Description

This is a follow-up to #1565, which fixed one test, but broke the following one. The second text expects the search field to be empty, but it is not cleared automatically.

- Fixes [RHIDP-3967](https://issues.redhat.com/browse/RHIDP-3967)

## PR acceptance criteria

### DIFF
--- a/e2e-tests/playwright/e2e/catalog-timestamp.spec.ts
+++ b/e2e-tests/playwright/e2e/catalog-timestamp.spec.ts
@@ -41,6 +41,7 @@ test.describe('Test timestamp column on Catalog', () => {
     await uiHelper.verifyRowInTableByUniqueText('timestamp-test', [
       /^\d{1,2}\/\d{1,2}\/\d{1,4}, \d:\d{1,2}:\d{1,2} (AM|PM)$/g,
     ]);
+    await uiHelper.searchInputPlaceholder('');
   });
 
   test('Toggle ‘CREATED AT’ to see if the component list can be sorted in ascending/decending order', async () => {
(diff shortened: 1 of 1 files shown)


e2e-tests/playwright/e2e/catalog-timestamp.spec.ts
```diff
@@ -41,6 +41,7 @@ test.describe('Test timestamp column on Catalog', () => {
     await uiHelper.verifyRowInTableByUniqueText('timestamp-test', [
       /^\d{1,2}\/\d{1,2}\/\d{1,4}, \d:\d{1,2}:\d{1,2} (AM|PM)$/g,
     ]);
+    await uiHelper.searchInputPlaceholder('');
```
subhashkhileri, line 44: we can again go into issue if all the row in catalog page have CreatedAt value.
instead of clearing the search I would say lets add a component without CreatedAt value which have similar name to the one used in this test. for eg: `timestamp-test-2` and do the partial search `timestamp-test` so both components will be visible and toggle assertion should work as expected.

zdrapela: Ok, agreed. It will be more bullet-proof your way 😄

zdrapela: Updated https://github.com/janus-qe/custom-catalog-entities/blob/main/timestamp-catalog-info.yaml and closing this PR.

zdrapela: We are gonna need to update the test anyway. The `timestamp-test` contained the `Created at` and it was at the first place (without sorting). And the second test requires it to be empty at the first place. So I updated the `timestamp-catalog-info.yaml` again and we need this PR to address that.

josephca:
opened a [PR]( https://github.com/janus-idp/backstage-showcase/pull/1604) for cherrypick 1.3.x

PR #1607, opened 2024-09-12
[WIP] [Test Automation] Automate E2E scenario with use of Operator for installation

Omar-AlJaljuli:
## Description

[Test Automation] Automate E2E scenario with use of Operator for installation

https://issues.redhat.com/browse/RHIDP-2919

## PR acceptance criteria


.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -97,6 +97,16 @@ configure_namespace() {
   oc config set-context --current --namespace="${project}"
 }
 
+configure_namespace_if_nonexistent() {
```
subhashkhileri, line 100: @Omar-AlJaljuli I don't see this method being used. can you please remove?

.ibm/pipelines/resources/rhdh-operator/installation/rhdh-subscription.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: rhdh-operator
```
subhashkhileri, line 4: @Omar-AlJaljuli Does this deploy the latest and new rhdh build on `main`?

.ibm/pipelines/resources/rhdh-operator/deployment/rhdh-start.yaml
```diff
@@ -0,0 +1,18 @@
...
+      configMaps:
+        - name: app-config-rhdh
+      mountPath: /opt/app-root/src
+    dynamicPluginsConfigMapName: dynamic-plugins
+    extraEnvs:
+      secrets:
+        - name: rhdh-secrets
+    extraFiles:
+      configMaps:
+        - name: rbac-policy
```
subhashkhileri, line 17: @Omar-AlJaljuli rbac-policy is not needed on non rbac instance.

subhashkhileri (changes requested):

subhashkhileri:
@Omar-AlJaljuli Have you tested in CI?

.ibm/pipelines/resources/rhdh-operator/installation/rhdh-subscription.yaml
Omar-AlJaljuli: This installs the latest GA version of RHDH. I was told that only GA will work with IBM, since IBM is a self-hosted cluster.

gustavolira:
I think the Operator could have it's own script instead of using openshift-ci-tests.sh. 
@josephca @subhashkhileri wdyt ?

Omar-AlJaljuli:
> I think the Operator could have it's own script instead of using openshift-ci-tests.sh. @josephca @subhashkhileri wdyt ?

@gustavolira I could create a new script specific to just the operator installation and e2e tests, but my thinking was that the new script will be majorly similar to the original script as it will have to reuse the majority of the functions; Also this way we can keep the operator e2e test config up to date and we won't have to update two scripts every time we create a test automation.

.ibm/pipelines/openshift-ci-tests.sh
gustavolira: @Omar-AlJaljuli  Why it was marked as resolved since this code is still here and not being used?

.ibm/pipelines/auth/secrets-rhdh-secrets.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: rhdh-secrets
 data:
   # the followings are place holders with dummy values that will be overwritten by CI
+  GITHUB_APP_CLIENT_ID_FLEX: dGVtcA==
```
gustavolira, line 7: what is GITHUB_APP_CLIENT_ID_FLEX ?

gustavolira:
> > I think the Operator could have it's own script instead of using openshift-ci-tests.sh. @josephca @subhashkhileri wdyt ?
> 
> @gustavolira I could create a new script specific to just the operator installation and e2e tests, but my thinking was that the new script will be majorly similar to the original script as it will have to reuse the majority of the functions; Also this way we can keep the operator e2e test config up to date and we won't have to update two scripts every time we create a test automation. What do you think?

I added only some nitpicks, but in general, it looks great!

.ibm/pipelines/auth/secrets-rhdh-secrets.yaml
Omar-AlJaljuli: Since there is now more than one app that we use to house the callback url, I created this to be the CLIENT_ID used for authentication, so we can keep the CLIENT_ID used for GH integration the same across all projects.

Omar-AlJaljuli: I think I overthought the GH integration tests. I realize I can use APP_3 secrets for the GH org integration as well since APP_3 and APP_2 and APP_1 share the same org.

.ibm/images/Dockerfile
```diff
@@ -100,4 +100,18 @@ RUN go install github.com/kadel/pr-commenter@latest && \
...
     ibmcloud plugin install -f kubernetes-service
 
+# Install skopeo
+RUN apt-get update -y && \
+    apt-get install -y skopeo
+
+# Install umoci
+RUN curl -LO "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" && \ 
+    chmod +x umoci.amd64 && \
+    mv umoci.amd64 /usr/local/bin/opm
```
rm3l, line 110:
```suggestion
    mv umoci.amd64 /usr/local/bin/umoci
```

Copy-paste issue?

.ibm/images/Dockerfile
```diff
@@ -100,4 +100,18 @@ RUN go install github.com/kadel/pr-commenter@latest && \
...
+RUN apt-get update -y && \
+    apt-get install -y skopeo
+
+# Install umoci
+RUN curl -LO "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" && \ 
+    chmod +x umoci.amd64 && \
+    mv umoci.amd64 /usr/local/bin/umoci
+
+# Install opm
+RUN curl -LO "https://github.com/operator-framework/operator-registry/releases/download/v1.19.5/linux-amd64-opm" && \
```
rm3l, line 113:
```suggestion
RUN curl -LO "https://github.com/operator-framework/operator-registry/releases/download/v1.23.0/linux-amd64-opm" && \ 
```

Otherwise, the `opm generate dockerfile` command used in the install script is not recognized. This also aligns with the version of `opm` we are using in the operator repo: https://github.com/redhat-developer/rhdh-operator/blob/main/Makefile#L468

rm3l (changes requested):

.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -343,6 +354,30 @@ install_tekton_pipelines() {
...
 
+install_rhdh_operator() {
+  local dir=$1
+  local namespace=$2
+
+  configure_namespace $namespace
+  
+  # Make sure script is up to date
+  rm -f /tmp/install-rhdh-catalog-source.sh
+  curl -L https://raw.githubusercontent.com/rm3l/redhat-developer-hub-operator/refs/heads/RHIDP-4934--unable-to-use-ci-install-script-in-prow-containers-where-podman-is-restricted/.rhdh/scripts/install-rhdh-catalog-source.sh > /tmp/install-rhdh-catalog-source.sh
```
rm3l, line 365:
```suggestion
  curl -L https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/main/.rhdh/scripts/install-rhdh-catalog-source.sh > /tmp/install-rhdh-catalog-source.sh
```

I've just merged https://github.com/redhat-developer/rhdh-operator/pull/404

Omar-AlJaljuli:
Duplicate: https://github.com/janus-idp/backstage-showcase/pull/2088

PR #1336, opened 2024-06-20
[Test automation] Configure and verify GitHub integration with Org data fetching

Omar-AlJaljuli:
## Description
Configured the environment to integrate the Janus-qe organization into the catalog. Created an automated test file that checks if the integration works. https://issues.redhat.com/browse/RHIDP-753.

Configuration followed from: https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.1/html-single/getting_started_with_red_hat_developer_hub/index#proc-add-source-control-rhdh-catalog_rhdh-getting-started

## Which issue(s) does this PR fix

[RHIDP-753](https://issues.redhat.com/browse/RHIDP-753)

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

PR #1625, opened 2024-09-16
[e2e][RHIDP-754] Github discovery UI test

nilgaar:
## Description

adding test for [RHIDP-754](https://issues.redhat.com//browse/RHIDP-754)

- Fixes [RHIDP-754](https://issues.redhat.com//browse/RHIDP-754)

## PR acceptance criteria


e2e-tests/package.json
```diff
@@ -28,6 +28,7 @@
     "typescript": "5.5.4"
   },
   "dependencies": {
+    "axios": "^1.7.7",
```
subhashkhileri, line 31: It's better to use the built-in functions from @playwright/test for API calls instead of a third-party library. One key reason is that it provides better logging in the HTML reporter and Trace View, making debugging easier.

e2e-tests/playwright/support/api/github.ts
```diff
@@ -0,0 +1,48 @@
+import axios, { AxiosResponse } from 'axios';
+import { getOrganizationResponse } from './github-structures';
+import { JANUS_ORG } from '../../utils/constants';
+
+export default class GithubApi {
```
subhashkhileri, line 5: @nilgaar There is `githubRequest()` method in APIHelper that also takes care of headers. You may  consider using it.

e2e-tests/playwright/support/api/github.ts
```diff
@@ -0,0 +1,48 @@
...
+  private static API_URL = 'https://api.github.com';
+  private static API_VERSION = '2022-11-28';
+  private static AUTH_HEADER = {
+    Accept: 'application/vnd.github+json',
+    Authorization: `Bearer ${process.env.GH_RHDH_QE_USER_TOKEN}`,
+    'X-GitHub-Api-Version': GithubApi.API_VERSION,
+  };
+
+  public async getOrganization(
```
subhashkhileri, line 14: @nilgaar can we move/add this to `APIHelper`.

e2e-tests/playwright/utils/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const GITHUB_URL = 'https://github.com/';
+export const JANUS_ORG = 'janus-idp';
+export const JANUS_QE_ORG = 'janus-qe';
+export const SHOWCASE_REPO = `${JANUS_ORG}/backstage-showcase`;
```
subhashkhileri, line 4: @nilgaar  we have API end points at one place `APIEndpoints.ts` you can consider adding it to `githubAPIEndpoints`

e2e-tests/playwright/e2e/github-discovery.spec.ts
subhashkhileri, line 1: @nilgaar There is a test for auto-discovery located at `e2e-tests/playwright/e2e/github-integration-org-fetch.spec.ts`. Is this discovery different, or can it be added to the old spec and moved?

e2e-tests/playwright/support/pages/Catalog.ts
```diff
@@ -1,14 +1,20 @@
...
 //${BASE_URL}/catalog page
 export class Catalog {
   private page: Page;
   private uiHelper: UIhelper;
+  private searchField: Locator;
 
   constructor(page: Page) {
     this.page = page;
     this.uiHelper = new UIhelper(page);
+    this.searchField = page.locator('#input-with-icon-adornment');
```
subhashkhileri, line 13: @nilgaar can we move the locators to `e2e-tests/playwright/support/pageObjects`?

subhashkhileri (changes requested):

e2e-tests/playwright/support/api/github.ts
nilgaar: I think that a class that contains both the logic and the urls and that's able to accept parameters on every component of the URL may be more versatile. The advantages that `githubRequest` currently have can be accomplished by a `request.newContext` inside the `GithubApi` class.

e2e-tests/playwright/support/pages/Catalog.ts
nilgaar: usually, in POM both locators and actions are placed in the same class. I am not sure if placing them in different filed has any advantage.

e2e-tests/playwright/utils/constants.ts
nilgaar: I see that majority of endpoints under `githubAPIEndpoints` refer to `backstageShowcaseAPI`. Not sure if my constants fit there. Besides, those constant are intend to be used to avoid misspelling some strings that we may use frequently, and are not indeed endpoints themselves.

nilgaar (commented):
thanks for the feedback! The `axios` requests have been deleted in favor of native Playwright  api requests. Some other things may be open for discussion. Thanks!

e2e-tests/package.json
nilgaar: agree! fixed!

e2e-tests/playwright/e2e/github-discovery.spec.ts
nilgaar: I do think that it is using the Github Organization Discovery here, as I understand from the PR in which the test was created: https://github.com/janus-idp/backstage-showcase/pull/1336/files.
For context: https://docs.google.com/document/d/1X40OGE7h0UW-nNGnkqW1ymsmgsK4zICmMuHvzFHlUzw/edit#heading=h.qucs7foygo5j

e2e-tests/playwright/support/pages/Catalog.ts
subhashkhileri: @nilgaar , I see your point. It could be an advantage if there are repetitive locators in different places and they can be utilized. since it seems to be just one locator in this case, I'm okay with it.

e2e-tests/playwright/utils/constants.ts
subhashkhileri: Intention behind `APIEndpoints.ts` was to have APIEndpoints at one place, I did some refactor in #1588 , Any further  Improvements are welcome.

subhashkhileri:
@nilgaar can you ask @josephca to add you to janus-idp org, so all check on this PR can run.

.ibm/pipelines/resources/config_map/configmap-app-config-rhdh-rbac.yaml
```diff
@@ -88,6 +88,13 @@ data:
           rules:
             - allow: [User, Group]
       providers:
+        github:
```
subhashkhileri, line 91: @nilgaar looks like your test will be running on showcase instance not on the showcase-rbac, if this is correct then this config update is not needed here, [configmap-app-config-rhdh.yaml](https://github.com/janus-idp/backstage-showcase/pull/1625/files#diff-d3959db965d3482fe2336650085090712166a0db3e7d9be769a42ff23eb70f5e) is enough

e2e-tests/playwright/utils/constants.ts
nilgaar: Sounds good! I think we can discuss an standardized approach to code APIs in the project, as we will have Github's, RHDH and we may have more in the future.

e2e-tests/playwright/support/pages/Catalog.ts
nilgaar: Absolutely. At the moment, I was thinking about adhering as much as possible to the POM described on Playwright https://playwright.dev/docs/pom

.ibm/pipelines/resources/config_map/configmap-app-config-rhdh-rbac.yaml
nilgaar: fixed!! thank you for the insight!

PR #1643, opened 2024-09-17
e2e: add scenarios to standardize auth providers tests

albarbaro:
## Description

PR adds scenarios for the authentication providers: Basic authentication, RHSSO 7.6, Microsoft EntraI and Github.
A following PR will integrate these scenarios with CI (more tuning is needed due to the long execution times)

[RHIDP-3499](https://issues.redhat.com/browse/RHIDP-3499)
[RHIDP-3500](https://issues.redhat.com/browse/RHIDP-3500)
[RHIDP-3750](https://issues.redhat.com/browse/RHIDP-3750)
[RHIDP-3538](https://issues.redhat.com/browse/RHIDP-3538)

- Fixes

## PR acceptance criteria

Set and export the envs defined in the utils/authenticationProviders/constants file, run the showcase-auth-providers project with 1 worker: 
`npx playwright test --project=showcase-auth-providers --retries=2 --workers=1`
The suite will automatically create the namespace, configmap and secret required to setup the instance.


e2e-tests/playwright/utils/authenticationProviders/rhssoHelper.ts
```diff
@@ -0,0 +1,222 @@
...
+    });
+    throw new Error('RHSSO setup failed: ' + JSON.stringify(e));
+  }
+  return {
+    usersCreated,
+    groupsCreated,
+  };
+}
+
+export async function clearUserSessions(username: string, realm: string) {
```
subhashkhileri, line 110: @albarbaro can we remove all unused methods from the PR? there seems to be many

e2e-tests/playwright/e2e/plugins/bulk-import.spec.ts
```diff
@@ -18,7 +18,7 @@ test.describe.serial('Bulk Import plugin', () => {
   let uiHelper: UIhelper;
   let common: Common;
   let bulkimport: BulkImport;
-  let backstageShowcase: BackstageShowcase;
+  let backstageShowcase: BackstageShowcase; // eslint-disable-line @typescript-eslint/no-unused-vars
```
subhashkhileri, line 21: @albarbaro you may remove this line, as its not being used in this test anymore.

e2e-tests/package.json
```diff
@@ -28,6 +30,13 @@
...
   },
   "dependencies": {
-    "node-fetch": "3.3.2"
+    "@azure/identity": "^4.4.1",
+    "@keycloak/keycloak-admin-client": "^25.0.2",
+    "@kubernetes/client-node": "^0.21.0",
+    "@microsoft/microsoft-graph-client": "^3.0.7",
+    "isomorphic-fetch": "^3.0.0",
+    "node-fetch": "3.3.2",
+    "octokit": "^4.0.2",
```
subhashkhileri, line 39: can we replace `isomorphic-fetch`, `node-fetch` and `octokit` and use the built-in functions from @playwright/test for API calls instead of a third-party library. One key reason is that it provides better assertions/logging in the HTML reporter and Trace View, and makes debugging easier.

albarbaro: `isomorphic-fetch` and `node-fetch` are both required by the msgraph and keycloak-admin client which is essential to manage the azure and rhsso users and groups during the execution. 
`octokit` is the only official client I found for github teams/users/orgs management, which is also essential for these scenarios. 
Re-writing them as standard rest call would be a lot of work and will need to maintained, but it's critical it can be done. Maybe we can just improve the logs: all the steps are logging to an external fine since the instances are destroyed and recreated multiple times.

test-results/basic-authentication/0fdf3f05aaa3f4323ceee380c7dc4475.webm
kadel, line 1: I guess this shouldn't be here

e2e-tests/playwright/utils/authenticationProviders/helper.ts
```diff
@@ -0,0 +1,320 @@
...
+  CHART: string,
+  NAMESPACE: string,
+  VALUES: string,
+  CHART_VERSION: string,
+  QUAY_REPO: string,
+  TAG_NAME: string,
+  FLAGS: Array<string>,
+) {
+  logger.info(`Deleting any exisitng helm release ${RELEASE}`);
+  await deleteHelmReleaseWithWait(RELEASE, NAMESPACE);
```
kadel, line 42: Is there a reason why you need to delete this first?
`helm upgrade` should be able to handle all changes in deployment without deleting release first

e2e-tests/playwright/utils/authenticationProviders/k8sHelper.ts
kadel, line 1: could this file be moved to `utils`? I don't see anything specific to authenticationProviders here, and there will be other tests that could function from here.

e2e-tests/playwright/utils/authenticationProviders/helper.ts
kadel, line 1: same as my comment for k8sHelper.ts, could we make this a generic helper?

e2e-tests/playwright/utils/authenticationProviders/helper.ts
albarbaro: Some of the test cases are applying changes at runtime to the authentication providers and checking the sync process and its results: the helm upgrade doesn't clear the data previously synced, unless you delete the postgres database, and that lead to inconsistencies in some tests, especially when they were retried for some reason. 
It also helps ensuring that each test case can be retried, mitigating any flakiness caused by the target cluster we are testing against. 
I guess there are different ways to handle this, but this was the simplest one that would also reduce the execution time of the whole suite; deleting and re-creating straight away ensures I can only wait for the first sync to happen, for example, without having to estimate when the next iteration will be or waiting for a longer time than needed.

e2e-tests/playwright/e2e/authProviders/basic-authentication.spec.ts
```diff
@@ -0,0 +1,168 @@
+import { test, Page, expect } from '@playwright/test';
+import { Common, setupBrowser } from '../../utils/Common';
+import { UIhelper } from '../../utils/UIhelper';
+import * as constants from '../../utils/authenticationProviders/constants';
+import { logger } from '../../utils/authenticationProviders/Logger';
```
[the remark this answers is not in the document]
albarbaro: It is writing to a file. It was requested by the rest of QE team since the configurations will be changed and destroyed during runtime; it's a way to keep event or config dumps from instances that do not exists anymore once the test finishes. This could probably be a common approach to all tests other than just printing to console.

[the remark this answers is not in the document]
kadel: If this is approach that other tests cases should adopt wouldn't ti be better to make it more generic so it can be re-used by other scenarios?

e2e-tests/playwright/utils/helper.ts
```diff
@@ -0,0 +1,320 @@
...
+    process.stdout.on('data', data => {
+      result = data;
+    });
+    process.stderr.on('data', data => {
+      logger.log({
+        level: 'error',
+        message: `Error occurred while running command`,
+        dump: data,
+      });
+      reject();
```
kadel, line 24: If the program writes to stderr, it doesn't mean that there was an error.
Stderr is sometimes used to output debugging or tracing information. 

The right way to determine if there was an error while executing the program is to exit the code of the program. 



 I would suggest something like this:
 
 ```javascript
export async function runShellCmd(command: string) {
  return new Promise<string>((resolve, reject) => {
    logger.info(`Executing command ${command}`);
    const process = exec(command);
    let result: string;
    process.stdout.on('data', data => {
      logger.log({
        level: 'debug',
        message: `Stdout output:`,
        dump: data,
      });
      result = data;
    });
    process.stderr.on('data', data => {
      logger.log({
        level: 'debug',
        message: `Stderr output:`,
        dump: data,
      });
    });
    process.on('exit', code => {
      if (code == 0) {
        resolve(result)
      } else {
        reject();
        throw Error('Error executing shell command');
      }
    });
  });
```

Comment (albarbaro) [reply e2e-tests/playwright/utils/helper.ts:24]: Updated to check the exit code.

File: e2e-tests/playwright/utils/helper.ts [hunk 7]
```diff
@@ -0,0 +1,321 @@
...
+  );
+  logger.log({
+    level: 'info',
+    message: `Release delete returned: `,
+    dump: result,
+  });
+  return result;
+}
+
+export async function dumpPodLog(POD: string, NAMESPACE: string) {
```
subhashkhileri, line 89: I don't see usage of this method.

e2e-tests/playwright/utils/helper.ts
```diff
@@ -0,0 +1,321 @@
...
+    );
+    expect(_syncTime).not.toBeNull();
+    expect(_syncTime).toBeGreaterThan(syncTime);
+  }).toPass({
+    intervals: [1_000, 2_000, 10_000],
+    timeout: SYNC__TIME * 2 * 1000,
+  });
+}
+
+export async function appendRBACPolicyToFileConfigMap(
```
subhashkhileri, line 147: No usage. can we remove?

test-results/basic-authentication/0fdf3f05aaa3f4323ceee380c7dc4475.webm
subhashkhileri, line 1: please remove this .webm

subhashkhileri (changes requested):

.ibm/pipelines/value_files/values-showcase_auth-providers.yaml
```diff
@@ -0,0 +1,239 @@
...
+            apps:
+              - appId: ${AUTH_ORG_APP_ID}
+                clientId: ${AUTH_ORG_CLIENT_ID}
+                clientSecret: ${AUTH_ORG_CLIENT_SECRET}
+                privateKey: ${AUTH_ORG1_PRIVATE_KEY}
+                webhookSecret: ${AUTH_ORG_WEBHOOK_SECRET}
+      catalog:
+        locations:
+          - type: url
+            target: https://github.com/albarbaro-org/some-test/blob/main/location.yaml
```
subhashkhileri, line 59: can you please move this to janus-qe org, so in future if needed anyone can update.

PR #1682, opened 2024-09-20
revert change to re-enable e2e tests

josephca:
## Description

Revert the temporary change to re-enable e2e tests
https://github.com/janus-idp/backstage-showcase/pull/1675

### DIFF
--- a/.ibm/pipelines/openshift-ci-tests.sh
+++ b/.ibm/pipelines/openshift-ci-tests.sh
@@ -1,6 +1,5 @@
 #!/bin/sh
 echo "Enable e2e-tests on release-1.3"
-exit 0 #temp change revert once /pj-rehearse auto-ack pass on relase PR:56923
 set -xe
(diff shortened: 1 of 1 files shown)


.ibm/pipelines/openshift-ci-tests.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
 echo "Enable e2e-tests on release-1.3"
```
nickboldt, line 2: does this pipeline ONLY run on one branch at a time? can we not us this for main AND release-1.3 in parallel?

gustavolira (approved):

.ibm/pipelines/openshift-ci-tests.sh
josephca: Yes, this pipeline runs only on one branch at a time thus, it doesn't run on main and release-1.3 in parallel.

PR #1719, opened 2024-10-02
[WIP] Switch AKS tests to `mapt` and spot instances

zdrapela:
## Description

This PR is switching the AKS nightly tests from long running cluster (that goes to sleep and wake up) to a spot instance (that is created and then destroyed). This approach allows us to save costs on Azure and gives us a fresh cluster every time, to avoid hiccups that the long running clusters might encounter.

This PR depends on having the https://github.com/janus-idp/backstage-showcase/pull/1643 merged first, to be able to use different auth provider than GitHub for majority of tests. GitHub requires a static callback URL (which is not achievable with `mapt`, but might be in the future: https://github.com/redhat-developer/mapt/issues/273). GitHub is also not the most stable auth provider and tests sometimes fail because of it.

The same approach as for AKS will be possible for GKE after `mapt` adds support: https://github.com/redhat-developer/mapt/issues/290

- Fixes [RHIDP-4268](https://issues.redhat.com/browse/RHIDP-3500)

## PR acceptance criteria

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,7 @@ dynamic-plugins/*/dist-dynamic/src
 .webpack-cache
 .idea/
 .idea/
+
+# mapt temporary files
+.ibm/pipelines/.pulumi
+.ibm/pipelines/kubeconfig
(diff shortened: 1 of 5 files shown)


zdrapela:
Closing in favor of https://github.com/openshift/release/pull/59507/files

PR #1749, opened 2024-10-08
chore: Move techdocs plugin from dependency to resolutions

coreydaley:
The techdocs update was inadvertently placed in the dependencies section of the package.json instead of the resolutions section, which is corrected by this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


nickboldt:
Presumably closed due to discussion in https://redhat.enterprise.slack.com/archives/C04CUSD4JSG/p1728408381871749?thread_ts=1728408381.871749&cid=C04CUSD4JSG

PR #1761, opened 2024-10-10
feat(app): introduce flavor configuration value

dzemanov:
## Description

Introduces new configuration value:
```
developerHub:
  flavor: <flavor>
```
Default value is set to `rhdh`.

This configuration value can be used by plugins to know whether they are running on RHDH, RHTAP, or Backstage.

- Fixes [RHIDP-4414](https://issues.redhat.com/browse/RHIDP-4414)

## PR acceptance criteria

- [X] Documentation is updated if necessary (requirement for new features)


packages/app/config.d.ts
```diff
@@ -6,6 +6,11 @@ export interface Config {
      * @visibility frontend
      */
     proxyPath?: string;
+    /**
+     * Name of the Backstage flavor (e.g. backstage, rhdh, rhtap)
+     * @visibility frontend
+     */
+    flavor?: string;
```
schultzp2020, line 13:
```suggestion
    flavor?: "backstage" | "rhdh" | "rhtap" | (string & {});
```

I'd suggest making this change because it autocompletes `"backstage" | "rhdh" | "rhtap"` while still allowing any generic string.

dzemanov: Makes sense, thank you

PatAKnight: @dzemanov @schultzp2020, It looks like this change is causing `yarn build` to fail with the following:

```
Error: Could not resolve values for path:"type". They are probably incompatible. Values: 
"object"
"string"
```

PatAKnight (approved):
/lgtm

AndrienkoAleksandr (approved):

PR #1779, opened 2024-10-17
chore: add link checker workflow

kadel:
## Description

add github action to validate links in markdown files


.github/workflows/link-checker.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Get all docs files that have changed
+      id: changed-files-yaml
+      uses: tj-actions/changed-files@v45
```
rm3l, line 15: Instead of this, how about just using the [`paths` filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) on the `pull_request` event? This way, the Workflow won't even run if those files are not modified.. And the different `if: ` statements in the steps will not be needed anymore.

```yaml
on:
	pull_request:
		paths:
			- '**.md'
			- 'docs/**'
```

.github/workflows/link-checker.yaml
```diff
@@ -0,0 +1,44 @@
...
+        echo "One or more doc file(s) has changed."
+        echo "List all the files that have changed: $DOC_ALL_CHANGED_FILES"
+
+    # If one file is changed, we can check all the files
+    - name: Check markdown links in docs
+      if: steps.changed-files-yaml.outputs.doc_any_changed == 'true'
+      uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        config-file: '.github/workflows/link-checker.config.json'
+        folder-path: 'docs/'
```
rm3l, line 36: Does this mean that the markdown files at the root (like `README.md`) will not be checked?

kadel: They are checked by the next step. It is a little bit hacky, but the reason is that we  don't want to check all the md files in the whole repository, as there are random md files that we don't control directly in this repo (like readme.md for static plugins etc...). 

It has been a while since I initially opened this so I need to check what is happening if I let it run on all md files

.github/workflows/link-checker.yaml
kadel: Good point. I don't remember why I did it this way 😇 
I'll try to change it see what happens

.github/workflows/link-checker.yaml
kadel: There is a lot of broken links in other md files like in `./packages/app/CHANGELOG.md` (see. https://github.com/janus-idp/backstage-showcase/actions/runs/11859232155/job/33051687669?pr=1779) 

I can fix them, but I'm not sure if going back to Changelog and fixing no longer existing links makes sense. I consider Changelog as a snapshot in a given time, not sure if going back and updating or removing links makes sense

rm3l (approved):

rm3l (approved):

rm3l (approved):

e2e-tests/README.md
```diff
@@ -2,14 +2,14 @@
 
 ### How to setup backstage configuration during the pipeline
 
-[configmap-app-config-rhdh.yaml](.ibm/pipelines/resources/config_map/configmap-app-config-rhdh.yaml) is the
+[configmap-app-config-rhdh.yaml](../.ibm/pipelines/resources/config_map/configmap-app-config-rhdh.yaml) is the
```
gustavolira, line 5: now the name of the file is `app-config-rhdh.yaml` `.ibm/pipelines/resources/config_map/app-config-rhdh.yaml`

PR #1818, opened 2024-10-24
feat: update to backstage 1.32

schultzp2020:
Signed-off-by: Paul Schultz <pschultz@pobox.com>


packages/backend/src/index.ts
```diff
@@ -40,10 +30,19 @@ backend.add(
         'configSchema.json',
       );
     },
+    moduleLoader: logger => new CommonJSModuleLoader(logger),
+    logger: config => {
+      const auditLogConfig = config?.getOptionalConfig('auditLog');
+      return {
+        transports: [
```
davidfestal, line 37: Didn't you miss the [audit log specific format](https://github.com/janus-idp/backstage-showcase/pull/1818/files#diff-77fa0019c0266a3c6ce82e398cc562233f1cbeec51c7dbd299724ee7c800212aL106) here ?

schultzp2020: It's not necessary.

davidfestal: why ?

ciiay:
Hi @schultzp2020 , I'm trying out this PR, and when I run `export-dynamic`, it generates changes of `package.json` and `yarn.lock` in each plugin wrapper to upgrade the supported backstage version. Shall we include all that changes in this PR as well? 

image

schultzp2020:
> Hi @schultzp2020 , I'm trying out this PR, and when I run `export-dynamic`, it generates changes of `package.json` and `yarn.lock` in each plugin wrapper to upgrade the supported backstage version. Shall we include all that changes in this PR as well? 
> 
> image
> 

I should change this PR to a draft since it's WIP.

packages/backend/src/index.ts
schultzp2020: `format` is the fallback is it `transports` doesn't work.

schultzp2020:
/override "Main Job"

Workflow has been renamed.

schultzp2020:
/override "PR Docker Build"

Workflow has been renamed.

PR #1871, opened 2024-11-01
chore(metadata): fix support states

kim-tsao:
## Description

Updates the plugin support values that were erroneously set as "tech-preview"

- Fixes #?
https://issues.redhat.com/browse/RHIDP-4797
## PR acceptance criteria


nickboldt:
Will any of these drop back to tech preview when they're no longer bundled/wrapped inside RHDH but on the plugin registry? If so might be better to keep them declaratively TP instead of GA.

nickboldt (approved):
Checked the current list at https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html-single/dynamic_plugins_reference/index#red-hat-supported-plugins and your changes look good -- only plugins we already list as GA are being updated here.

PR #1881, opened 2024-11-05
feat(backend): add service factories by default

gashcrumb:
## Description

This change explicitely adds the default service factories to the backend statically to prevent dynamic plugins from being able to override them by default.  It's possible to override each statically added service factory via an environment variable derived from the service factory ID.  So for example to add a "core.rootHttpService" service factory configuration from a dynamic plugin, set ENABLE_CORE_ROOTHTTPROUTER_OVERRIDE to "true".  This change also adds a logger to the backend main.  Finally, a unit test has been added that checks the installed backend-defaults value for the defaultServiceFactories list against what this change adds to catch future regressions.

- Fixes [RHIDP-4786](https://issues.redhat.com/browse/RHIDP-4786)

## PR acceptance criteria

It's enough to just check that this doesn't break anything, however an example exists [here](https://github.com/gashcrumb/dynamic-plugins-root-http-middleware) that has a plugin that can be used to try this out.

### DIFF
--- a/packages/backend/src/defaultServiceFactories.test.ts
+++ b/packages/backend/src/defaultServiceFactories.test.ts
@@ -1,0 +1,38 @@
+import { ServiceFactory } from '@backstage/backend-plugin-api';
+
+import { defaultServiceFactories } from './defaultServiceFactories';
+
+// explicitly check this against the module inside the installed package
+const {
+  defaultServiceFactories: upstreamDefaultServiceFactories,
+  // eslint-disable-next-line
+} = require('../../../node_modules/@backstage/backend-defaults/dist/CreateBackend.cjs.js');
+
+function findDifference(a1: string[], a2: string[]) {
+  const set = new Set(a2);
+  return a1.filter(i => !set.has(i));
+}
+
+function findSymmetricDifference(a1: string[], a2: string[]) {
+  return [...new Set([...findDifference(a1, a2), ...findDifference(a2, a1)])];
+}
+
+/**
+ * Validate that the installed backend-defaults package contains the expected
+ * list of default service factories.  A failure in this test indicates that
+ * either the export was removed, the list was moved, or the list in
+ * "defaultServiceFactories" should be updated.
+ */
+describe('Default service factory list comparison', () => {
+  it('Should produce an expected difference of service factories as compared to the upstream implementation', () => {
+    const upstreamServiceFactoryIds = upstreamDefaultServiceFactories.map(
+      (serviceFactory: ServiceFactory) => serviceFactory.service.id,
+    );
+    const serviceFactoryIds = defaultServiceFactories.map(
+      (serviceFactory: ServiceFactory) => serviceFactory.service.id,
+    );
+    expect(
+      findSymmetricDifference(upstreamServiceFactoryIds, serviceFactoryIds),
+    ).toEqual(['core.rootLogger']);
+  });
+});
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -5,18 +5,30 @@ import { PackageRoles } from '@backstage/cli-node';
 import * as path from 'path';
 
 import { configureCorporateProxyAgent } from './corporate-proxy';
+import { getDefaultServiceFactories } from './defaultServiceFactories';
 import { CommonJSModuleLoader } from './loader';
-import { transports } from './logger';
+import { createStaticLogger, transports } from './logger';
 import {
   healthCheckPlugin,
   pluginIDProviderService,
   rbacDynamicPluginsProvider,
 } from './modules';
 
+// Create a logger to cover logging static initialization tasks
+const staticLogger = createStaticLogger({ service: 'developer-hub-init' });
+staticLogger.info('Starting Developer Hub backend');
+
 // RHIDP-2217: adds support for corporate proxy
 configureCorporateProxyAgent();
 
 const backend = createBackend();
+
+const defaultServiceFactories = getDefaultServiceFactories({
+  logger: staticLogger,
+});
+defaultServiceFactories.forEach(serviceFactory => {
+  backend.add(serviceFactory);
+});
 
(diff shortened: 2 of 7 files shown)


packages/backend/src/index.ts
```diff
@@ -1,24 +1,58 @@
...
+// Install a static root HttpRouterServiceFactory configure function that can
+// be explicitly overridden by the user
+if (
+  (process.env.DISABLE_STATIC_ROOT_HTTP_ROUTER_CONFIG || '').toLowerCase() !==
+  'true'
+) {
+  staticLogger.info('Using static root HttpRouterServiceFactory configuration');
+  backend.add(
+    rootHttpRouterServiceFactory({
+      configure({ app, middleware, routes }) {
```
davidfestal, line 36: Do we really need this configure method, if it is simply the default service that we add back ?

davidfestal (changes requested):
Shouldn't we do something more general than that, for all the default services added in the `createdBackend()`method ?
With environment variables like:
ENABLE_OVERRIDING_SERVICE_<serviceName> (serviceName being the service name mentioned in the list of default services)

packages/backend/src/index.ts
gashcrumb: Ah, I was not aware this could be optional which would be better, let me give that a try and update this.  Thanks!

gashcrumb:
> Shouldn't we do something more general than that, for all the [default services](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/CreateBackend.ts?rgh-link-date=2024-11-06T22%3A10%3A27Z#L37) added in the `createdBackend()`method ? With environment variables like: ENABLE_OVERRIDING_SERVICE_ (serviceFactoryName being the service factory name mentioned in the list of default services: [backstage/backstage@`master`/packages/backend-defaults/src/CreateBackend.ts#L37](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/CreateBackend.ts?rgh-link-date=2024-11-06T22%3A10%3A27Z#L37))

I can look at doing that, certainly.  I didn't see the same customization options offered for some of the other services when I was looking through them in preparation for this, so I opted to keep the scope for this much more narrow.  But I'll take another look.  Thanks!

davidfestal:
I like the last commit: https://github.com/janus-idp/backstage-showcase/pull/1881/commits/4a76ffa96c62f7295631540cd616010217df9326

Would it be possible to add a unit test that would simply extract the upstream source file based on the `backstage.json` of the showcase, and check that the service factories are the same ?

gashcrumb:
> I like the last commit: [4a76ffa](https://github.com/janus-idp/backstage-showcase/commit/4a76ffa96c62f7295631540cd616010217df9326)

Cool!

> Would it be possible to add a unit test that would simply extract the upstream source file based on the `backstage.json` of the showcase, and check that the service factories are the same ?

yeah good call, lemme look into that.

packages/backend/src/defaultServiceFactories.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  it('Should produce an expected difference of service factories as compared to the upstream implementation', () => {
+    const upstreamServiceFactoryIds = upstreamDefaultServiceFactories.map(
+      (serviceFactory: ServiceFactory) => serviceFactory.service.id,
+    );
+    const serviceFactoryIds = defaultServiceFactories.map(
+      (serviceFactory: ServiceFactory) => serviceFactory.service.id,
+    );
+    expect(
+      findSymmetricDifference(upstreamServiceFactoryIds, serviceFactoryIds),
+    ).toEqual(expect.arrayContaining(['core.rootLogger']));
```
davidfestal, line 36: Should this be `arrayContaining` or one of the `equal` flavors. I guess we don't want to accept a list of differences that contains **_both `core.rootLogger` and something else_** ?

davidfestal (commented):
Do we need:
- a test that expects the backend application to fail when some custom core service is added as a dynamic plugin (and the corresponding environment variable is not set)
- a test that expects the backend application to effectively let add some custom service as a dynamic plugin when the expected environment variable has been provided.

gashcrumb:
> Do we need:
> 
> * a test that expects the backend application to fail when some custom core service is added as a dynamic plugin (and the corresponding environment variable is not set)
> * a test that expects the backend application to effectively let add some custom service as a dynamic plugin when the expected environment variable has been provided.

We do!  I created a separate epic to do this later in [RHIDP-4522](https://issues.redhat.com/browse/RHIDP-4522)

gashcrumb:
@subhashkhileri something seems amiss with the test env, I wonder if the persistent store with the dynamic plugins isn't cleaned up between the non-rbac and rbac tests.  Looking at the failed tests, I see the [rbac showcase instance](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/janus-idp_backstage-showcase/1881/pull-ci-janus-idp-backstage-showcase-main-e2e-tests/1858736355492761600/artifacts/e2e-tests/janus-idp-backstage-showcase/artifacts/showcase-rbac/pod_logs/rhdh-backstage-576f884f5f-8cthl_backstage-backend.log) loaded the segment plugin:

```
"level":"\u001b[32minfo\u001b[39m","message":"Loaded dynamic frontend plugin 'janus-idp-backstage-plugin-analytics-provider-segment' from 'file:///opt/app-root/src/dynamic-plugins-root/janus-idp-backstage-plugin-analytics-provider-segment-1.10.0' ","plugin":"scalprum","service":"backstage","timestamp":"2024-11-19 05:26:00"}
```

But I can see it's disabled [here](https://github.com/janus-idp/backstage-showcase/blob/main/.ibm/pipelines/value_files/values_showcase-rbac.yaml#L106-L107).  Is there a chance maybe the PV holding the dynamic-plugins-root is getting reused or could this be a symptom of trying to run tests concurrently?

PR #1903, opened 2024-11-07
fix(corporate-proxy): use the built-in `EnvHttpProxyAgent` from `undici` to fix the proxy issues reported

rm3l:
## Description

Undici has added native support for proxy handling, so it is no longer necessary for us to have our own custom proxy handling for `fetch` calls.
This was introduced in Undici v6.14.0 and enhanced in the next versions and appears to fix the issues faced with the handling of the `NO_PROXY` env var by `fetch` [1]

This PR simplifies the code by removing such custom handling and also clarifies the expectations in the documentation (based on some manual testing I've done).

Note that [`global-agent`](https://github.com/gajus/global-agent) should still continue to handle proxy settings for `node-fetch`.

[1] https://github.com/nodejs/undici/releases/tag/v6.14.0

- Fixes https://issues.redhat.com/browse/RHDHBUGS-106
- Fixes https://issues.redhat.com/browse/RHIDP-4646

## PR acceptance criteria

See https://github.com/redhat-developer/rhdh-local#testing-rhdh-in-a-simulated-corporate-proxy-setup
```shell
$ git clone https://github.com/redhat-developer/rhdh-local.git rhdh-local && cd rhdh-local
$ RHDH_IMAGE=quay.io/janus-idp/backstage-showcase:pr-1903 \
  podman-compose -f compose.yaml -f compose-with-corporate-proxy.yaml up -d
```

### DIFF
--- a/.rhdh/docker/Dockerfile
+++ b/.rhdh/docker/Dockerfile
@@ -364,12 +364,17 @@ ENV NPM_CONFIG_ignore-scripts='true'
 ENV SEGMENT_WRITE_KEY=gGVM6sYRK0D0ndVX22BOtS7NRcxPej8t
 ENV SEGMENT_TEST_MODE=false
 
-# RHIDP-2217: corporate proxy support (configured using 'global-agent')
+# RHIDP-2217: corporate proxy support (configured using 'global-agent' for 'node-fetch' calls and 'undici' for 'fetch' calls)
 # This is to avoid having to define several environment variables for the same purpose,
 # i.e, GLOBAL_AGENT_HTTP(S)_PROXY (for 'global-agent') and the conventional HTTP(S)_PROXY (honored by other libraries like Axios).
 # By setting GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE to an empty value,
 # 'global-agent' will use the same HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables.
 ENV GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE=''
+
+# RHDHBUGS-106,RHIDP-4646: requests to the loopback interface should bypass the corporate proxy if set.
+# Note that NO_PROXY will take effect only if the 'HTTP(S)_PROXY' environment variables are set.
+# Users can still override this when running the image.
+ENV NO_PROXY='localhost,127.0.0.1'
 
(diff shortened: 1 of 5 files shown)


durandom:
I ❤️ this PR, a lot of ✂️ on code, but addition in documentation.

One thing: I wonder if we should add localhost always to the NO_PROXY list, or at least mention it in the docs? Since for the backend to function correctly, the localhost request *must not* be proxied, IIUC

rm3l:
> I ❤️ this PR, a lot of ✂️ on code, but addition in documentation.
> 
> One thing: I wonder if we should add localhost always to the NO_PROXY list, or at least mention it in the docs? Since for the backend to function correctly, the localhost request _must not_ be proxied, IIUC

That's a good point. I agree it makes sense. I'm thinking of adding localhost (and the loopback IP address) as the default value for `NO_PROXY` in the image since `NO_PROXY` will anyway only take effect if the `HTTP(S)_PROXY` env vars are set. I'll update this and also mention this in the docs. Thanks.

rm3l:
> > I ❤️ this PR, a lot of ✂️ on code, but addition in documentation.
> > One thing: I wonder if we should add localhost always to the NO_PROXY list, or at least mention it in the docs? Since for the backend to function correctly, the localhost request _must not_ be proxied, IIUC
> 
> That's a good point. I agree it makes sense. I'm thinking of adding localhost (and the loopback IP address) as the default value for `NO_PROXY` in the image since `NO_PROXY` will anyway only take effect if the `HTTP(S)_PROXY` env vars are set. I'll update this and also mention this in the docs. Thanks.

@durandom PTAL at [`81003ea` (#1903)](https://github.com/janus-idp/backstage-showcase/pull/1903/commits/81003ea78cdc8b98c003719a321ccaeefc032ac8) which adds `localhost,127.0.0.1` as the default value for `NO_PROXY` in the container image. Users still have the ability to override it if needed when deploying RHDH, and it is documented that they should include at least `localhost`.
Once this PR is merged, I will create another PR to also update the related [downstream docs](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/administration_guide_for_red_hat_developer_hub/assembly-running-rhdh-behind-a-proxy#assembly-running-rhdh-behind-a-proxy).

durandom (approved):
/approve

looks good, good job.

kadel:
👍 I tested this with my local testing setup, where I was able to reproduce the issue, and this  fixes it 🎉 

/approve

PR #1905, opened 2024-11-07
Optimize GitHub login by caching authentication state

gustavolira:
Introduced logic to reuse existing authentication state from a local file to speed up subsequent logins. If the file exists, it loads the state and avoids redundant login steps, otherwise performs the login and saves the state for future use.

## Description

- Fixes #?

## PR acceptance criteria


e2e-tests/playwright/utils/Common.ts
```diff
@@ -80,12 +81,30 @@ export class Common {
...
   async loginAsGithubUser(userid: string = process.env.GH_USER_ID) {
-    await this.logintoGithub(userid);
-    await this.page.goto("/");
-    await this.waitForLoad(240000);
-    await this.uiHelper.clickButton("Sign In");
-    await this.checkAndReauthorizeGithubApp();
-    await this.uiHelper.waitForSideBarVisible();
+    const projectName = test.info().project.name;
+    const isNotRbacProject = !projectName.includes("rbac");
+    if (isNotRbacProject && fs.existsSync("authState.json")) {
```
nilgaar, line 86: As I understand, this won't able the use of a second github user? As it will load the cookies of the first github user who logs in.

gustavolira: @nilgaar you are correctly, but we only need a second github user for RBAC, and because of it I will not restore the auth state for rbac

nilgaar: I see, that makes sense.
Even though, can we dynamically write two different files, one for rbac and another for non-rbac?
If this is outside the scope of the PR, for me abstracting the string "authState.json" into a constant will be enough to merge.

gustavolira: @nilgaar  I didn’t quite understand. Could you clarify what you mean by "two different files"?
As for the constant, I’ll definitely take care of it. I’ll switch to using a constant for the string as soon as I get the test results. I can’t make the change right now, as it would place me at the end of the test execution queue.

nilgaar: What I was thinking about is: can we dynamically create two files: like "authState_user1.json" and "authState_user2.json", that are loaded depending on the _userid_ parameter?
Maybe it has implications that I am not considering though, on the paper it looks possible.

gustavolira: @nilgaar now I get your point, and you totally right, I changed the code to use it. let's see the results!

PR #1924, opened 2024-11-13
chore(e2e):  optimize .ibm Dockerfile for Improved Build Performance and Dependency Updates

gustavolira:
https://issues.redhat.com/browse/RHIDP-4956

Optimized the Dockerfile by combining multiple RUN commands into a single layer to reduce the size and improve build efficiency. Set essential environment variables together, consolidated installations, and added comments for clarity. This refactor enhances maintainability and performance of the Docker build process.

### DIFF
--- a/.ibm/images/Dockerfile
+++ b/.ibm/images/Dockerfile
@@ -1,106 +1,54 @@
 FROM mcr.microsoft.com/playwright:v1.47.2-jammy
 
+# Set environment variables for the container
 ENV CI=1 \
     QT_X11_NO_MITSHM=1 \
     _X11_NO_MITSHM=1 \
     _MITSHM=0 \
-    NODE_PATH=/usr/local/lib/node_modules
+    NODE_PATH=/usr/local/lib/node_modules \
+    HELM_VERSION="v3.12.3" \
+    OC_VERSION="4.14.3" \
+    OCM_VERSION="0.1.76"
 
-# Define Helm and OpenShift CLI (oc) versions
-ENV HELM_VERSION="v3.12.3"
-ENV OC_VERSION="4.14.3"
-ENV OCM_VERSION="0.1.76"
+# Install essential dependencies and Node.js tools
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    nodejs \
+    openssl libssl-dev ca-certificates \
+    curl wget jq colorized-logs && \
+    npm install -g typescript yarn@latest && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    echo "Node: $(node -v)\nNPM: $(npm -v)\nYarn: $(yarn -v)\nTypeScript: $(tsc -v)"
 
-ARG CI_XBUILD
-
-RUN apt-get update && \
-    apt-get install -y nodejs-dev nodejs \
-    openssl libssl-dev ca-certificates make cmake cpp gcc g++ zlib1g zlib1g-dev brotli libbrotli-dev python3 && \
-    apt-get clean && rm -rf /var/lib/apt/lists/*
-
-# should be root user
+# Ensure the current user is root and verify permissions
 RUN echo "whoami: $(whoami)" \
-    # command "id" should print:
-    # uid=0(root) gid=0(root) groups=0(root)
-    # which means the current user is root
+    # The "id" command should confirm the user is root
     && id \
-    && npm install -g typescript \
-    # give every user read access to the "/root" folder where the binary is cached
-    # we really only need to worry about the top folder, fortunately
-    && ls -la /root \
+    # Make the "/root" folder readable for all users
     && chmod 755 /root \
-    # always grab the latest Yarn
-    # otherwise the base image might have old versions
-    # NPM does not need to be installed as it is already included with Node.
-    && npm i -g yarn@latest \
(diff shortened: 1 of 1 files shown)


nickboldt:
Would be cool if you could also switch this to be based on UBI9 so we're not using Ubuntu to test RH product :D

zdrapela:
> Would be cool if you could also switch this to be based on UBI9 so we're not using Ubuntu to test RH product :D 

Unfortunately, Playwright only supports Ubuntu and Debian from Linux distros 😢

.ibm/images/Dockerfile
```diff
@@ -1,103 +1,62 @@
...
+RUN curl -fsSL "https://github.com/mikefarah/yq/releases/download/v4.43.1/yq_linux_amd64.tar.gz" | tar -xz && \
+    mv yq_linux_amd64 /usr/local/bin/yq && \
+    curl -sL https://aka.ms/InstallAzureCLIDeb | bash && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# Install additional tools and IBM Cloud CLI plugins
 RUN go install github.com/kadel/pr-commenter@latest && \
     ibmcloud plugin install -f cloud-object-storage && \
-    ibmcloud plugin install -f kubernetes-service
+    ibmcloud plugin install -f kubernetes-service && \
```
zdrapela, line 58: Why not join the `Install IBM Cloud CLI` with `IBM Cloud CLI plugins`?

.ibm/images/Dockerfile
```diff
@@ -1,103 +1,62 @@
...
+    ibmcloud --version
 
-# Set environment variables to make Go work correctly
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+# Install yq and Azure CLI
+RUN curl -fsSL "https://github.com/mikefarah/yq/releases/download/v4.43.1/yq_linux_amd64.tar.gz" | tar -xz && \
+    mv yq_linux_amd64 /usr/local/bin/yq && \
+    curl -sL https://aka.ms/InstallAzureCLIDeb | bash && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
zdrapela, line 53: Why not keep Azure CLI alone and join `yq` with `# Install Go, Helm, OpenShift CLI, and ocm-cli`?

zdrapela (commented):
Hi @gustavolira, 
have you measured how much that speeds up the build process?
Some of the optimization is definitely needed and helps with speeding up, but I believe that having too many `RUN` commands squashed together reduces the maintainability and readability of the `Dockerfile`.
The `# Install essential dependencies and Node.js tools` and `# Install Go, Helm, OpenShift CLI, and ocm-cli` seem to be quite packed up.

.ibm/images/Dockerfile
```diff
@@ -1,103 +1,62 @@
...
-ENV GO_SHA256 464b6b66591f6cf055bc5df90a9750bf5fbc9d038722bb84a9d56a2bea974be6
-
-# Install Go and other tools used by the pipeline
-RUN apt-get update && \
-    apt-get install -y curl && \
-    curl -LO "https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz" && \
-    echo "${GO_SHA256} go${GO_VERSION}.linux-amd64.tar.gz" | sha256sum -c - && \
+# Install Go, Helm, OpenShift CLI, ocm-cli, and yq
+RUN curl -LO "https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz" && \
+    echo "464b6b66591f6cf055bc5df90a9750bf5fbc9d038722bb84a9d56a2bea974be6 go${GO_VERSION}.linux-amd64.tar.gz" | sha256sum -c - && \
```
zdrapela, line 37: I think that the `GO_SHA256` should be kept as a variable close to `GO_VERSION`. You would be updating them both together. The comment `# Set Go version and the expected SHA256 hash for verification` seems to be useful and I would keep it around.

zdrapela:
I added one more comment regarding Go. Apart from that I'm good with the changes once the PR is green

.ibm/images/Dockerfile
```diff
@@ -1,103 +1,62 @@
...
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install Azure CLI
-RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
-
-# Set environment variables to make Go work correctly
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
-
-RUN go install github.com/kadel/pr-commenter@latest && \
```
zdrapela, line 99: Is this safe to remove, or is it used somewhere? @kadel

kadel: If I recall correctly, this was used to post PR comment with logs in CI jobs running on IBM cloud.
If nothing is running there anymore, this should be safe to delete.

josephca: IIRC, the pr commenter was used a long time ago and no longer the case. So, it can be removed safely.

gustavolira: removed `GO` and `pr-commenter`

josephca (approved):
/lgtm

PR #1953, opened 2024-11-19
chore(e2e): rhidp-4958 re-enable the flaky test - catalog-entity refresh is denied after DELETE

gustavolira:
https://issues.redhat.com/browse/RHIDP-4958

Added error tracking and logging to beforeEach, afterEach, and afterAll hooks to improve robustness and clarity. Skipped cleanup actions if errors occurred during retries to avoid inconsistent state.

## Description

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
+++ b/e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
@@ -10,9 +10,13 @@ test.describe("Test RBAC plugin REST API", () => {
   let uiHelper: UIhelper;
   let page: Page;
   let responseHelper: Response;
+  // Variable to track errors or test states
+  let hasErrors = false;
+  let retriesRemaining: number;
 
   test.beforeAll(async ({ browser }, testInfo) => {
     page = (await setupBrowser(browser, testInfo)).page;
+    retriesRemaining = testInfo.project.retries;
 
     uiHelper = new UIhelper(page);
     common = new Common(page);
@@ -23,9 +27,12 @@ test.describe("Test RBAC plugin REST API", () => {
     responseHelper = new Response(apiToken);
   });
 
-  test.beforeEach(
-    async () => await new Common(page).checkAndClickOnGHloginPopup(),
-  );
+  // eslint-disable-next-line no-empty-pattern
+  test.beforeEach(async ({}, testInfo) => {
+    console.log(
+      `beforeEach: Attempting setup for ${testInfo.title}, retry: ${testInfo.retry}`,
+    );
+  });
 
   test("Test that roles and policies from GET request are what expected", async ({
     request,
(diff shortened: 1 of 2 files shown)


gustavolira:
> It might cause problems down the line if we have tests that run after the RBAC api tests. If the RBAC api tests fail and the afterhooks don't trigger then there might be policies that are unaccounted for. But as long as the api tests remain as the last tests ran in the RBAC test suite, it should be fine.

I updated the code to check if it's the last retry -> execute afterAll

nilgaar (changes requested):
If a particular test requires a element to be present in the catalog, can't we check it or add it through an API or similar? I think we shall try a more isolated, resilient approach; as the current ones looks hard to read and maintain.

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
```diff
@@ -23,9 +27,18 @@ test.describe("Test RBAC plugin REST API", () => {
...
-  test.beforeEach(
-    async () => await new Common(page).checkAndClickOnGHloginPopup(),
-  );
+  // eslint-disable-next-line no-empty-pattern
+  test.beforeEach(async ({}, testInfo) => {
+    try {
+      console.log(
+        `beforeEach: Attempting setup for ${testInfo.title}, retry: ${testInfo.retry}`,
+      );
+      await new Common(page).checkAndClickOnGHloginPopup();
```
subhashkhileri, line 36: rbac test doesn't need github login. we may remove this.

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
```diff
@@ -23,9 +27,18 @@ test.describe("Test RBAC plugin REST API", () => {
...
+    } catch (error) {
+      console.error(`beforeEach failed for ${testInfo.title}:`, error);
+      throw error; // Ensure Playwright handles the failure correctly for retries
```
subhashkhileri, line 39: if you throw the error from catch its the same logic without the try-catch block, try catch may not be needed

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
```diff
@@ -267,8 +303,10 @@ test.describe("Test RBAC plugin REST API", () => {
...
       );
 
-      expect(deleteRemainingPolicies.ok());
-      expect(deleteRole.ok());
-    },
-  );
+      expect(deleteRemainingPolicies.ok()).toBeTruthy();
+      expect(deleteRole.ok()).toBeTruthy();
+    } catch (error) {
+      console.error("Error during cleanup in afterAll:", error);
```
subhashkhileri, line 309: console.error will not report test failure, it will give false positive afaik.

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
gustavolira: I agree that throwing the error directly could simplify the code. I'll adjust it to remove unnecessary try-catch blocks.

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
gustavolira: You’re right. I will replace console.error with assertions or other mechanisms that correctly fail the test if cleanup encounters issues. Thank you for pointing that out!

e2e-tests/playwright/e2e/plugins/rbac/rbac-api.spec.ts
Omar-AlJaljuli: Not really sure if I'm understanding this comment right. Is this about something specifically related to GH login popup, or are you saying the user won't need to be logged in for these tests?
These tests focus specifically on a role created in the beginning for the rhdh-qe user.

    const members = ["user:default/rhdh-qe"];

    const newPolicy = {
      entityReference: "role:default/test",
      permission: "catalog-entity",
      policy: "read",
      effect: "deny",
    };`

gustavolira:
> If a particular test requires an element to be present in the catalog, can't we check it or add it through an API or similar? I think we shall try a more isolated, resilient approach; as the current ones looks hacky and hard to read and maintain.

@Omar-AlJaljuli wrote this test. To be honest, I'm not entirely familiar with how it works, and I have never run it manually. However, we could consider creating a ticket to improve this test with your suggestion, @nilgaar . Can you create a ticket for it in terms of improvements and reducing flakiness ?

PatAKnight (approved):

PR #2022, opened 2024-11-29
fix: add mui v5 workaround in wrapper code

christoph-jerolimov:
## Description

Big thanks and kudos to @gashcrumb who solved the issue https://github.com/janus-idp/backstage-plugins/pull/2183 for dynamic plugins with a comment. But it needs to be applied to all plugins.

I'm adding this here now (also with respect to the 1.4 timeline) to these frontend plugins that allowed me to add it.

I would expect that a better solution is to automatically apply this rule when bundling a plugin! But for me the code looks save to be called twice and I'm fine to revert this in 1.4 or 1.5 as soon as we're sure that we include this somewhere else. Tbc. and tbd.  :smirk: 

This fixes at least our ad-hoc issues with Bulk import and RBAC.

* https://issues.redhat.com/browse/RHIDP-5104 Bulk import layout issues with MUI v5 components (missing gaps)

## Changes / Verification

I tested these plugins and compared `next` with this PR on a cluster.

* [x] ACR (MUI v4) => No difference
* [x] Bulk Import (MUI v5 and had issues before) => Issues are solved
* [x] Home page => No difference
* [x] RBAC (MUI v5 and had issues before)
* [x] Tech Radar (MUI v4 afaik) => No difference
* [x] Kubernetes (just because I wanted test Tekton and Topology :D)
* [x] Tekton (PF + MUI v4) => No difference
* [x] Topology (PF + MUI v5 and had issues before that @ciiay fixed in the plugin itself) => no double padding or something. Two small changes: Download log button is now blue and the log dialog uses now 80% instead of 100% of the screen width (both changes lgtm)
* [x] Tech docs => No difference, but unluckily it does not fix the selection bug

Here are the differences (and I think fixes) that I noticed:

### Bulk import

| Area | Before | After |
| --- | --- | --- |
| Bulk import page size selector is fixed | image | image |
| Padding and collapse icon on the Add repo tutorial is fixed | image | image |
| Repository / Orginzation ToggleButton is fixed | image | image |
| Button padding is restored | image | image |
| Added repository table has no more padding due bigger buttons | image | image |
| More aligned codeowner button | image | image |

### RBAC

| Area | Before | After |
| --- | --- | --- |
| Download button padding is fixed | image | image |
| Action buttons have a better padding | image | image |
| Configuration access button has a font color now (I'm hovering it here to show also the background) | image | image |
| Configuration access > Not options has now the right padding | image | image |
| Toast X is now better aligned | 
image | 
image |

/cc @divyanshiGupta @debsmita1 @ShiranHi @ciiay @invincibleJai 

## PR acceptance criteria

### DIFF
--- a/dynamic-plugins/wrappers/backstage-community-plugin-acr/src/index.ts
+++ b/dynamic-plugins/wrappers/backstage-community-plugin-acr/src/index.ts
@@ -1,2 +1,10 @@
+import { unstable_ClassNameGenerator as ClassNameGenerator } from '@mui/material/className';
+
+ClassNameGenerator.configure(componentName => {
+  return componentName.startsWith('v5-')
+    ? componentName
+    : `v5-${componentName}`;
+});
+
 export * from '@backstage-community/plugin-acr';
 
--- a/dynamic-plugins/wrappers/backstage-community-plugin-jfrog-artifactory/src/index.ts
+++ b/dynamic-plugins/wrappers/backstage-community-plugin-jfrog-artifactory/src/index.ts
@@ -1,2 +1,10 @@
+import { unstable_ClassNameGenerator as ClassNameGenerator } from '@mui/material/className';
+
+ClassNameGenerator.configure(componentName => {
+  return componentName.startsWith('v5-')
+    ? componentName
+    : `v5-${componentName}`;
+});
+
 export * from '@backstage-community/plugin-jfrog-artifactory';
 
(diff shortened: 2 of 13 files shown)


ShiranHi:
Looks great, thank you @christoph-jerolimov

gashcrumb (approved):
Yep, if the imported code wasn't `unstable_*` I would have considered adding this code to plugins automatically.  We could still consider a CLI flag to inject this code into exported plugin however for the medium term, but for now this is the way.

christoph-jerolimov:
At least images job should pass

/retest

PR #2036, opened 2024-12-02
chore(ci): build multiplatform next container image

kadel:
## Description

This updates the build process for the next image (build from main branch)

It adds the arm64 image build. arm64 and amd64 are built in parallel.

Platform specific images are tagged with `next-{platform}`. This is mainly to make the amd64 image available quickly (takes less than 40 min) for testing, as arm64 build can take 4 hours.

After both builds are done, the `next` multiplatform tag is pushed and can be used.

Screenshot 2024-12-02 at 14 50 21

Screenshot 2024-12-02 at 14 52 13

- Fixes https://issues.redhat.com/browse/RHIDP-5027

### DIFF
--- a/.github/actions/docker-build/action.yaml
+++ b/.github/actions/docker-build/action.yaml
@@ -36,6 +36,15 @@ inputs:
   push:
     description: Whether to push the image
     required: true
+  platform:
+    description: "Target given CPU platform architecture (default: linux/amd64)"
+    required: false
+    default: linux/amd64
+
+outputs:
+  digest:
+    value: ${{ steps.build.outputs.digest }}
+
 
 runs:
   using: composite
@@ -77,6 +86,7 @@ runs:
 
     - name: Build and push Docker image
       uses: docker/build-push-action@v5
+      id: build
       with:
         context: .
         file: docker/Dockerfile
@@ -84,3 +94,4 @@ runs:
         provenance: false
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
+        platforms: ${{ inputs.platform }}
--- a/.github/workflows/next-build-image.yaml
+++ b/.github/workflows/next-build-image.yaml
@@ -25,7 +25,7 @@ concurrency:
 
 env:
   REGISTRY: quay.io
-  REGISTRY_IMAGE: ${{ github.repository }}
+  REGISTRY_IMAGE: rhdh-community/rhdh
 
 jobs:
   build-image:
@@ -64,11 +64,7 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
-<<<<<<< HEAD
-          imageName: rhdh-community/rhdh
-=======
           imageName: ${{ env.REGISTRY_IMAGE }}
->>>>>>> fbed803d (chore(ci): build multiplatform next container image)
           imageTags: |
             type=raw,value=next-${{ env.PLATFORM_ARCH }}
             type=sha,prefix=next-,suffix=-${{ env.PLATFORM_ARCH }}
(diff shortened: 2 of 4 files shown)


nickboldt:
Do we want PR builds on arm too, in case you're developing something locally on your arm64 mac?

kadel:
> Do we want PR builds on arm too, in case you're developing something locally on your arm64 mac?

I'm working on that as well. I'll do it in a separate PR. e2e tests depend on those images, so I want to make sure that they are not delayed by arm image builds

.github/workflows/next-build-image.yaml
```diff
@@ -40,26 +47,99 @@ jobs:
...
+
 
       - name: Build and Push with Buildx
         uses: ./.github/actions/docker-build
+        id: build
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
+<<<<<<< HEAD
```
nickboldt, line 67: this looks sus 
```suggestion
```

.github/workflows/next-build-image.yaml
```diff
@@ -40,26 +47,99 @@ jobs:
...
           imageName: rhdh-community/rhdh
```
nickboldt, line 68:
```suggestion
```

.github/workflows/next-build-image.yaml
```diff
@@ -40,26 +47,99 @@ jobs:
...
+=======
```
nickboldt, line 69:
```suggestion
```

.github/workflows/next-build-image.yaml
```diff
@@ -40,26 +47,99 @@ jobs:
...
+          imageName: ${{ env.REGISTRY_IMAGE }}
+>>>>>>> fbed803d (chore(ci): build multiplatform next container image)
```
nickboldt, line 71:
```suggestion
```

.github/workflows/pr-build-image.yaml
```diff
@@ -76,6 +68,7 @@ jobs:
           # to autodelete PR image tags, set an expiry date
           imageLabels: quay.expires-after=14d
           push: true
+          platform: linux/amd64
```
nickboldt, line 71: so PR builds are single-arch but next builds are dual-arch? 🤔 What if your contributor is on a Mac and wants to test their PR build?

.github/actions/docker-build/action.yaml
```diff
@@ -36,6 +36,15 @@ inputs:
   push:
     description: Whether to push the image
     required: true
+  platform:
```
nickboldt, line 39: Because the intent here is multiple arches should this be "platforms" ?

.github/actions/docker-build/action.yaml
```diff
@@ -36,6 +36,15 @@ inputs:
...
+    description: "Target given CPU platform architecture (default: linux/amd64)"
```
nickboldt, line 40: How do I list more than one? Or can this action only run on a single arch, but is then called by another workflow?

.github/actions/docker-build/action.yaml
kadel: This is intentional. This should be run for only run platform.
The idea is that this action gets triggered in workflow for each platform in parallel

.github/actions/docker-build/action.yaml
kadel: more info at https://github.com/redhat-developer/rhdh/pull/2036/files/7b9d684bcf52bcf4ad16565736616256989ce678#r1914593017

.github/workflows/pr-build-image.yaml
kadel: Yes, for now yes.

My plan is to do the PR images in followup PR.
As it will require small updates to e2e tests as well.
e2e tests should **not** wait for the whole multiplatform image to be built, as it can take 4+ hours. So we will need to reconfigure them to wait for the platform specific tag.

nickboldt:
lasr PR build failed with
```
ERROR: failed to solve: failed to push quay.io/rhdh-community/rhdh:pr-2036: failed to copy: unexpected status from PUT request to https://quay.io/v2/rhdh-community/rhdh/blobs/uploads/9a90c668-b180-4b0d-b97c-c01d71271356?digest=sha256%3A8b76ee4a0a25c4b269af66290f7fdb1d0c56383c3c67d60a32778039bb0f0a28: 413 Request Entity Too Large
```

kadel:
@nickboldt can you give this another look?

nickboldt (approved):

PR #2045, opened 2024-12-03
test(e2e): fixin rbac tests

nilgaar:
## Description

enabling RBAC tests
also replaced github logins with keylock login

- Fixes https://issues.redhat.com/browse/RHIDP-5208
- and https://issues.redhat.com/browse/RHIDP-4993

## PR acceptance criteria


e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
```diff
@@ -134,220 +103,419 @@ test.describe.skip(
...
+
+      if (!policiesResponse.ok()) {
+        throw Error(
+          `RBAC policiesResponse API call failed with status code ${policiesResponse.status()}`,
+        );
+      }
+
+      await Response.checkResponse(
+        rolesResponse,
+        RbacConstants.getExpectedRoles(),
```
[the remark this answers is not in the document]
nilgaar: this is a good take! I just pushed a change to compare them as objects instead of strings. Let's see if it works as expected.
Thank you!

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
```diff
@@ -10,31 +11,29 @@ import { Common, setupBrowser } from "../../../utils/common";
...
+    Note that:
+    The policies generated from a policy.csv or ConfigMap file cannot be edited or deleted using the Developer Hub Web UI. 
+    https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/authorization/managing-authorizations-by-using-the-web-ui#proc-rbac-ui-edit-role_title-authorization
+*/
+test.describe.serial("Test RBAC", () => {
+  test.describe
+    .serial("Test RBAC plugin: load permission policies and conditions from files", () => {
+    test.beforeEach(async ({ page }) => {
+      await new Common(page).loginAsKeycloakUser();
+      await page.goto("/rbac");
```
subhashkhileri, line 29: In test at least once we should cover the rbac navigation from sidebar `Administration>RBAC`

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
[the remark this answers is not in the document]
subhashkhileri: it was string in function return but it was being converted to object using JSON.parse so I guess it still would same thing.

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
nilgaar: totally. Just pushed it, thanks!

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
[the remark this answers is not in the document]
nilgaar: I see... now all RBAC test will run sequentially, as I saw collisions between the APIs and UIs tests.
So, new or deleted elements shouldn't be a problem. I replaced the equality with `toContainEqual`, which I understand doesn't care about the order of the elements in the array. This shall help us too.

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
```diff
@@ -10,31 +11,41 @@ import { Common, setupBrowser } from "../../../utils/common";
...
+/*
+    Note that:
+    The policies generated from a policy.csv or ConfigMap file cannot be edited or deleted using the Developer Hub Web UI. 
+    https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/authorization/managing-authorizations-by-using-the-web-ui#proc-rbac-ui-edit-role_title-authorization
+*/
+test.describe.serial("Test RBAC", () => {
+  test.describe
+    .serial("Test RBAC plugin: load permission policies and conditions from files", () => {
+    test.beforeEach(async ({ page }) => {
+      await new Common(page).loginAsKeycloakUser();
```
subhashkhileri, line 28: I am not aware of if there is any reason but I guess performing login after each test will increase the test execution time.
does it makes sense to move login to beforeAll()?

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
```diff
@@ -134,220 +115,419 @@ test.describe.skip(
...
-    await uiHelper.openSidebar("RBAC");
-    await uiHelper.verifyHeading("RBAC");
   });
 
-  test.beforeEach(
-    async () => await new Common(page).checkAndClickOnGHloginPopup(),
-  );
+  test.describe("Test RBAC plugin as an admin user", () => {
+    test.beforeEach(async ({ page }) => {
+      await new Common(page).loginAsKeycloakUser();
```
subhashkhileri, line 122: https://github.com/janus-idp/backstage-showcase/pull/2045/files#r1893990996

e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts
nilgaar: Well. There are indeed a couple of reasons for it.
In the first place, this may help us to keep the tests as isolated as possible; it is true that now or rbac test are quite coupled together, but at some point, maybe with the keyloack or something similar, we can try to make them completely isolated, which shall be our objective in my opinion.
On the other hand, I saw this in the [docs](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/authorization/enabling-and-giving-access-to-rbac#enabling-and-giving-access-to-rbac) about the admin account:
Sign out from the existing Red Hat Developer Hub session and log in again using the declared policy administrator account. 
Apparently, it regards only about admin account setup, but I think it may make sense to logout and login again when making changes on RBAC elements, to ensure the changes are correctly pulled into the frontend.

subhashkhileri: Okay got it. thanks

PR #2056, opened 2024-12-04
chore: remove deprecated rotate file

schultzp2020:
Signed-off-by: Paul Schultz <pschultz@pobox.com>


kadel:
Is there a Jira for this? We are basically removing a feature here, so we will need at least a release note for this

kadel:
This is documented downstream, so we also need Jira to track docs update and make sure that docs are updated before next release



https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/audit_log/index#ref-audit-log-file-rotation-overview_assembly-audit-log

kadel:
It looks like this issue is related to https://issues.redhat.com/browse/RHIDP-4852

PR #2113, opened 2024-12-17
fix(wrappers): embed runtime dependencies

gashcrumb:
## Description

This change updates the wrappers for the notification backend and catalog backend module for gitlab org to explicitly embed some packages that are needed at runtime, as the normal behavior of the janus-idp CLI for these dependencies are to move them to peer dependencies.

- Fixes [RHIDP-5308](https://issues.redhat.com/browse/RHIDP-5308)
- Fixes https://issues.redhat.com/browse/RHIDP-5319


kadel:
Tested is with the image build from this PR and it fixes the issue. I can see the gitlab-org plugin being correctly loaded

kadel (approved):

PR #2129, opened 2024-12-19
chore(plugins): bump dynamic plugin wrappers to the latest x.y.z version (RHIDP-5317)

nickboldt:
### What does this PR do?

chore(plugins): bump dynamic plugin wrappers to the latest x.y.z version [release-1.4]

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### Screenshot/screencast of this PR
N/A

### What issues does this PR fix or reference?
https://issues.redhat.com/browse/RHIDP-5317

See also https://github.com/janus-idp/backstage-showcase/pull/2128 for the main branch **(which includes minor and patch bumps)**

### How to test this PR?
N/A

### PR Checklist

As the author of this Pull Request I made sure that:

- [x] Code produced is complete
- [ ] Code builds without errors
- [ ] Tests are covering the bugfix
- [ ] Relevant user documentation updated
- [ ] Relevant contributing documentation updated

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

PR #2139, opened 2024-12-26
chore: update Dynamic Plugins versions for RHDH 1.4

jesuino:
Updating Dynamic Plugins versions for RHDH 1.4.

Kindly double check the versions


gashcrumb:
This seems like a candidate for automating perhaps?  If we did that we could probably also then use these version lists as a basis to maybe start doing more up-front dependency verification/correction in the CLI when exporting dynamic plugins.

nickboldt:
See related PRs https://github.com/janus-idp/backstage-showcase/pull/2128 and https://github.com/janus-idp/backstage-showcase/pull/2129

> This seems like a candidate for automating perhaps? If we did that we could probably also then use these version lists as a basis to maybe start doing more up-front dependency verification/correction in the CLI when exporting dynamic plugins.

See also https://github.com/janus-idp/backstage-showcase/pull/2129/files#diff-f1efe2420dcffbb51ddd91a1041f899eda7398b3794b9158239c3e639b14db66R10 ... something we could trigger on PRs or on pushes.

docs/dynamic-plugins/versions.md
```diff
@@ -1,17 +1,17 @@
...
-## RHDH 1.4 (pre-release, versions can change for final release)
+## RHDH 1.4
 
 <!-- source
-https://github.com/janus-idp/backstage-showcase/blob/main/backstage.json
+https://github.com/janus-idp/backstage-showcase/blob/release-1.4/backstage.json
 -->
 
-Based on [Backstage 1.31.3](https://backstage.io/docs/releases/v1.31.0)
+Based on [Backstage 1.32.6](https://backstage.io/docs/releases/v1.32.0)
```
nickboldt, line 9: this is for 1.4 but your changes are targetting `main` branch (1.5), not `release-1.4`.

nickboldt:
Should we abandon this PR as stale?

PR #2210, opened 2025-01-20
feat: support configuration to include transient group ownership during auth

JessicaJHee:
## Description
* Adds `InheritedGroupOwnershipResolver` to provide the ability to add optionally transient parent groups into the resolved user group membership
* Enabled by this config: `includeInheritedGroupOwnership?: boolean`
* Adds unit tests for `AuthProviderModule` and `InheritedGroupOwnershipResolver`

With `includeTransientGroupOwnership` to true:
image
* Includes `maintainers` group which is a parent of `maintainers-plugins` and `maintainers-showcase`

With `includeTransientGroupOwnership` to false (by default):
image
* only direct group ownership is included

- Fixes [RHIDP-5505](https://issues.redhat.com/browse/RHIDP-5505) and [RHIDP-5506](https://issues.redhat.com/browse/RHIDP-5506)

## PR acceptance criteria

### DIFF
--- a/packages/backend/src/transientGroupOwnershipResolver.ts
+++ b/packages/backend/src/transientGroupOwnershipResolver.ts
@@ -1,0 +1,111 @@
+import { AuthService, DiscoveryService } from '@backstage/backend-plugin-api';
+import { CatalogApi, CatalogClient } from '@backstage/catalog-client';
+import {
+  Entity,
+  GroupEntity,
+  RELATION_CHILD_OF,
+  RELATION_MEMBER_OF,
+  stringifyEntityRef,
+} from '@backstage/catalog-model';
+import { Config } from '@backstage/config';
+import { AuthOwnershipResolver } from '@backstage/plugin-auth-node';
+
+export class TransientGroupOwnershipResolver implements AuthOwnershipResolver {
+  private readonly catalogApi: CatalogApi;
+  private readonly config: Config;
+  private readonly auth: AuthService;
+
+  constructor(deps: {
+    discovery: DiscoveryService;
+    config: Config;
+    auth: AuthService;
+    catalog?: CatalogApi;
+  }) {
+    this.catalogApi = deps.catalog
+      ? deps.catalog
+      : new CatalogClient({ discoveryApi: deps.discovery });
+    this.config = deps.config;
+    this.auth = deps.auth;
+  }
+
+  private async resolveParentGroups(
+    groupRefs: string[],
+    processedGroups: Set<string> = new Set(),
+  ): Promise<string[]> {
+    const allTransientGroupRefs = new Set<string>();
+
+    for (const groupRef of groupRefs) {
+      if (processedGroups.has(groupRef)) continue;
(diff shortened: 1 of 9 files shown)


packages/backend/src/transientGroupOwnershipResolver.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    for (const groupRef of groupRefs) {
+      if (allTransientGroupRefs.has(groupRef)) continue;
+
+      allTransientGroupRefs.add(groupRef);
+
+      const res = await this.catalogApi.getEntitiesByRefs({
+        entityRefs: [groupRef],
+        fields: ['kind', 'relations'],
+      });
```
PatAKnight, line 39: When the permission framework is enabled, this will return a deny and only the direct membership refs will be recorded. You will need to retrieve a token to pass to the catalog to avoid this problem.

I have an example [here](https://github.com/backstage/community-plugins/blob/main/workspaces/rbac/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts#L101:L111) from the RBAC backend plugin, if needed.

packages/backend/src/transientGroupOwnershipResolver.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      const groupEntity = res.items?.find(
+        e => e?.kind.toLowerCase() === 'group',
```
PatAKnight, line 42: I wonder if it would be safer to do this as `e?.kind.toLocaleLowerCase('en-US')}`?

PatAKnight (commented):
Overall, it looks fairly good to me. A couple of requests / questions from me.

- Could you add some docs [here](https://github.com/redhat-developer/rhdh/blob/main/docs/auth.md). That way we have something for the docs team to reference.
- I know there was an issue with large tokens, I believe that it had something to do with ownership refs, is that still a problem today?

Also, I did run into an infinite loop while testing that I think might be worth addressing. I tested with a user that had cyclic behavior in the group memberships, something like `user -> group_a <=> group_b`

JessicaJHee:
@PatAKnight Thank you very much for the review! I've addressed  your feedback, PTAL :)

> I know there was an issue with large tokens, I believe that it had something to do with ownership refs, is that still a problem today?

I'm not too familiar with this issue, the only thing I found was [this](https://issues.redhat.com/browse/RHIDP-2412) and it looks like it's been resolved in 1.3 👍

PatAKnight (approved):
Tested and it worked for me.
/lgtm

PatAKnight:
@JessicaJHee, I just mentioned this in another [PR](https://github.com/redhat-developer/rhdh/pull/2238#issuecomment-2642991218). One of the links is now dead after the deprecation of the aap plugin. I recommended updating the [deadlink](https://github.com/redhat-developer/rhdh/blob/main/docs/dynamic-plugins/export-derived-package.md?plain=1#L46) to something like https://github.com/backstage/community-plugins/blob/main/workspaces/3scale/plugins/3scale-backend/src/module.ts.

PR #2218, opened 2025-01-21
chore(ci): build images from tags, build next nightly

kadel:
Modifies .github/workflows/next-build-image.yaml
- to make it more generic, so it can be triggered even on the tags and build images from tag
- builds next nightly instead of after every push to main branch. arm build takes almost 5 hours, so we won't be able to build more than 4 of them a day anyway 

related to https://issues.redhat.com/browse/RHIDP-5027


.github/workflows/next-build-image.yaml
```diff
@@ -12,12 +12,15 @@
...
+name: Build Next and Tag Image
 
 on:
+  schedule:
+    # run at 3:18 UTC every day
+    - cron: "18 3 * * *"
   push:
-    branches:
-      - main
+    tags:
```
nickboldt, line 22: we previously decided to not push tagged container images in the comunity repo, choosing instead to focus on the downstream Konflux builds. Why do we want to return to tagged images?

.github/workflows/next-build-image.yaml
```diff
@@ -12,12 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Build Next Image
+name: Build Next and Tag Image
 
 on:
+  schedule:
+    # run at 3:18 UTC every day
+    - cron: "18 3 * * *"
```
nickboldt, line 20: I'm confused why you want to switch from "on every push to main" to "once a day"... are we trying to reduce noise ? server usage? data centre heat causing global warming?

.github/workflows/next-build-image.yaml
kadel: We need arm images for rhdh-local users and plugin developers using Macs.
We need tagged (versioned) because when developing or testing plugins they need to target a specific version.

.github/workflows/next-build-image.yaml
kadel: If you do "on every push to main" than every time, there is a new commit on main, the running job gets canceled, and a new one is trigerred. Which makes sense because otherwise you are building something that is already out of date. 

To complete a full multiplatform image build, which typically takes over 5 hours, you’d need at least five consecutive hours without merging any pull requests (PRs) into main.
During regular workdays, with multiple PRs being merged daily, this job will likely get triggered and canceled repeatedly until nighttime when there’s less activity.

.github/workflows/next-build-image.yaml
nickboldt: hmm. OK, seems like going backwards but +1.

.github/workflows/next-build-image.yaml
nickboldt: FIVE HOURS? What's it doing, compiling every single dependent package from sources??

nickboldt: also you can turn off the concurrency settings to cancel in-progress builds if you don't want things to be cancelled and started over:

https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/next-build-image.yaml#L24

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

kadel: yes, but if we don't cancel it than we are building something that is  is never going to be used

kadel:
/hold

lets get agreement that everyone is ok with switching to daily builds

rm3l (commented):
+1. Just a small comment below:

PR #2219, opened 2025-01-21
feat: update Backstage to 1.34.2

04kash:
## Description

Fixes: https://issues.redhat.com/browse/RHIDP-5187


04kash:
When I tested locally, I was able to successfully test the homepage and tech radar plugins. I will try testing on a cluster as well.

PR #2232, opened 2025-01-22
chore(ci): label stale PRs and clean its cache

kadel:
This adds GH action that labels PR as "Stale" when there is no activity on PR for 7 days.
It doesn't close them, just adds label.

When the PR is labeled as "Stale" it will also trigger a workflow that cleans its GitHub Actions cache. This should prevent problems that we sometimes have with cache getting filled. 

fixes https://issues.redhat.com/browse/RHIDP-5610


.github/workflows/cache-cleanup.yaml
```diff
@@ -5,12 +5,15 @@ on:
...
       - closed
+      - labeled
 
 jobs:
   cleanup:
     runs-on: ubuntu-latest
     steps:
       - name: Cleanup
+        # when the PR is closed or labeled as 'Stale'
+        if: ${{ github.event.action == 'closed' || github.event.label.name == 'Stale' }}
```
nickboldt, line 16: does the case of the strings here matter? you have lowercase `closed` and Titlecase `Stale`...

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,20 @@
...
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-pr-message: 'This PR is stale because it has been open 7 days with no activity.'
+          days-before-pr-stale: 7
+          days-before-pr-close: -1
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity.'
+          days-before-issue-stale: 30
```
nickboldt, line 19: Since we don't use GH issues for tracking things we could stale/close issues immediately. We use JIRA for this repo. 
```suggestion
          days-before-issue-stale: 1
```

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,20 @@
...
+          days-before-issue-close: -1
```
nickboldt, line 20:
```suggestion
          days-before-issue-close: 1
```

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,20 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-pr-message: 'This PR is stale because it has been open 7 days with no activity.'
+          days-before-pr-stale: 7
+          days-before-pr-close: -1
```
nickboldt, line 17:
```suggestion
          days-before-pr-close: 21
```

.github/workflows/stale.yaml
kadel: I was thinking about this as well. All issues except 2 had last updated more than 30 days ago, so even setting this to 30 will label all existing issues as stale :-D 

Ideally, we should try to close all of them and disable issues on this repo.

.github/workflows/stale.yaml
kadel: If we do this, we should probably update the message to something like 
`This issue is stale because it has been open 30 days with no activity. This project switched to Jira at https://issues.redhat.com/projects/RHIDP/ for issue tracking. If this issues is still relevant, please reopen it there.'

kadel: I changed it 7 days be a little bit nicer

nickboldt:
If we want to close *issues* as well as *PRs* then we probably need some sort of exception for https://github.com/redhat-developer/rhdh/issues/320 as that's the renovate dashboard.

nickboldt (approved):

.github/workflows/cache-cleanup.yaml
nickboldt: if the event is being generated as "Stale" then let's jusst use that.

PR #2238, opened 2025-01-23
feat(app): dynamic Techdocs addons

dzemanov:
## Description

This change adds support for dynamically loading [Techdocs Addons](https://backstage.io/docs/features/techdocs/addons/). It creates a wrapper for [techdocs-module-addons-contrib](https://github.com/backstage/backstage/tree/master/plugins/techdocs-module-addons-contrib) which contains techdocs extensions:
- LightBox
- ReportIssue
- TextSize
- ExpandableNavigation

 A plugin can export a Techdocs extension like:
```
export const TextSize = techdocsModuleAddonsContribPlugin.provide(
  createTechDocsAddonExtension({
    name: 'TextSize',
    location: TechDocsAddonLocations.Settings,
    component: TextSizeAddon,
  }),
);
```

Use new configuration option:
```
frontend:
  backstage.plugin-techdocs-module-addons-contrib:
    techdocsAddons:
      - importName: TextSize
```  

- Fixes [RHIDP-5499](https://issues.redhat.com/browse/RHIDP-5499)

## PR acceptance criteria

You can add to `app-config.yaml` this config:

```
techdocs:
  builder: 'local'

dynamicPlugins:
  rootDirectory: dynamic-plugins-root
  frontend:
    backstage.plugin-techdocs:
      routeBindings:
        targets:
          - importName: techdocsPlugin
        bindings:
          - bindTarget: catalogPlugin.externalRoutes
            bindMap:
              viewTechDoc: techdocsPlugin.routes.docRoot
          - bindTarget: scaffolderPlugin.externalRoutes
            bindMap:
              viewTechDoc: techdocsPlugin.routes.docRoot
      dynamicRoutes:
        - path: /docs
          importName: TechDocsIndexPage
          menuItem:
            icon: docs
            text: Docs
        - path: /docs/:namespace/:kind/:name/*
          importName: TechDocsReaderPage
      mountPoints:
        - mountPoint: entity.page.docs/cards
          importName: EntityTechdocsContent
          config:
            layout:
              gridColumn: "1 / -1"
            if:
              allOf:
                - isTechDocsAvailable
        - mountPoint: search.page.results
          importName: TechDocsSearchResultListItem
        - mountPoint: search.page.filters
          importName: TechdocsSearchFilter
        - mountPoint: search.page.types
          importName: techdocsSearchType
          config:
            props:
              name: Documentation
              icon: docs

    backstage.plugin-techdocs-module-addons-contrib:
      techdocsAddons:
        - importName: TextSize
        - importName: LightBox
        - importName: ReportIssue
        - importName: ExpandableNavigation
```

https://github.com/user-attachments/assets/7c17b0d4-34dc-43f9-a234-8ce312b0e065


dzemanov:
ReportIssue bug is fixed.
Not sure if I can use `importName` for the filtering for TechDocs dynamic components or is there something better / safer to use.

dzemanov:
Pinging again as it was not picked
/cc @gashcrumb

packages/app/src/components/DynamicRoot/DynamicRoot.tsx
```diff
@@ -365,6 +370,30 @@ export const DynamicRoot = ({
       return acc;
     }, []);
 
+    const techdocsAddonComponents = techdocsAddons.reduce<TechdocsAddon[]>(
```
gashcrumb, line 373: On one hand I kind wonder if we shouldn't consider a more general approach for things that have add-ons like this techdocs thingie and the scaffolder field extensions etc.  But knowing that this could be addressed by the new frontend system I wouldn't want to make it more difficult to map our current configuration to the new frontend system.  @davidfestal what do you think?

PatAKnight:
@dzemanov It looks like when the aap plugin got [deprecated](https://github.com/janus-idp/backstage-plugins/pull/2628), it broke a link in our docs. Tried to find a replacement. 

Update this link [here](https://github.com/redhat-developer/rhdh/blob/main/docs/dynamic-plugins/export-derived-package.md?plain=1#L46) to this: https://github.com/backstage/community-plugins/blob/main/workspaces/3scale/plugins/3scale-backend/src/module.ts.

packages/app/src/components/DynamicRoot/DynamicRoot.tsx
dzemanov: Discussed this with dynamic plugins and it was decided it would be better to have these be separate.

christoph-jerolimov (approved):
Hi @dzemanov, @gashcrumb, @PatAKnight,

this is a really nice feature and the code change looks overall good to me.

@gashcrumb, do you want merge it this way?

I've tested it with [pr-2238-4015364c](https://quay.io/rhdh-community/rhdh:pr-2238-4015364c) on a cluster, and everything works as expected.

1. Default looks good to me, but `ReportIssues` was missing in that image. I guess this should be added to `dynamic-plugins.default.yaml`
2. So disabled plugin looks currently same as the default. ;)

   image

3. Currently I need to enable the plugin and the default `ReportIssue` manually. I've done it this way:

   ```yaml
   global:
     dynamic:
       plugins:
         - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-module-addons-contrib
           disabled: false
   
   upstream:
     backstage:
       appConfig:
         dynamicPlugins:
           frontend:
             backstage.plugin-techdocs-module-addons-contrib:
               techdocsAddons:
                 # - importName: TextSize
                 # - importName: LightBox
                 - importName: ReportIssue
                 # - importName: ExpandableNavigation
   ```

   It also looks good. The interessting part is that the 2nd button  (Leave feedback) disappears. So with this change it looks like before.

   image

   I  think that was always the case. Personally I think that addon should not remove the button and add just the selection-overlay option instead. Users without a mouse has now a feature less!

   Selection works fine:

   image

4. Enabling the addons in the configuration above works also fine:

   image

   * Lightbox works also fine.
   * I've discussed with @dzemanov that ExpandableNavigation doesn't work yet and that she should create an issue for that.

/lgtm
/approve

christoph-jerolimov:
For some reason I didn't understand yet, the latest version didn't worked on a cluster.

Absolutely no dynamic plugin is loaded. We need to check if its this PR or a general issue. But to be save I mark this as

/hold

app-config.dynamic-plugins.yaml
```diff
@@ -471,6 +471,9 @@ dynamicPlugins:
             props:
               name: Documentation
               icon: docs
+    backstage.plugin-techdocs-module-addons-contrib:
```
gashcrumb, line 474: I've seen both a `.` and a `-` used as the org/package separator.  I think we should consider that inconsistency here confuses users, and I wonder if we shouldn't just stick with using `-`.  I'll have a 2nd comment on this in the plugin's `scalprum` config :-)

gashcrumb (approved):
I did a quick local test using the PR image and could see the add-ons installed, very nice!  A couple comments but nothing that needs to be addressed in this PR.

dynamic-plugins/wrappers/backstage-plugin-techdocs-module-addons-contrib/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@backstage/cli": "0.29.6",
+    "@janus-idp/cli": "3.2.0",
+    "react": "18.3.1",
+    "typescript": "5.7.3"
+  },
+  "files": [
+    "dist",
+    "dist-scalprum"
+  ],
+  "scalprum": {
```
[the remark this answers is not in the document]
dzemanov: Yeah, not necessary. I have seen both versions being used, it would be great to unify this.

app-config.dynamic-plugins.yaml
dzemanov: I would maybe let it be this way for now, but we can do a following bulk update for all the plugins that have it this way.

gashcrumb: yep, I don't know if there's any formal convention here so maybe one needs to be established.

christoph-jerolimov:
I've tested it now again on a new cluster and couldn't reproduce my earlier issue. :man_shrugging: 

/lgtm
/unhold

PR #2249, opened 2025-01-24
chore: fix PR builder so when new commits are added, new builds happen (check for pr-123-SHA instead of pr-123)

gustavolira:
## Description

chore: fix PR builder so that when new commits are added to an existing PR, new builds happen. Previously it was assumed that the presence of a pr-123 tag meant the build was already done, but this could be for a previous commit SHA, so we need to check for the presence of a pr-123-SHA image tag, instead of pr-123. 

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/.github/workflows/pr-build-image.yaml
+++ b/.github/workflows/pr-build-image.yaml
@@ -93,7 +93,7 @@ jobs:
         uses: ./.github/actions/get-sha
 
       - name: Build and Push with Buildx
-        if: env.relevant_changes == 'true' || env.image_exists == 'false'
+        if: env.relevant_changes == 'true' && env.image_exists == 'false'
         uses: ./.github/actions/docker-build
         with:
           registry: ${{ env.REGISTRY }}
@@ -108,7 +108,7 @@ jobs:
           platform: linux/amd64
 
       - name: Comment the image pull link
-        if: env.relevant_changes == 'true' || env.image_exists == 'false'
+        if: env.relevant_changes == 'true' && env.image_exists == 'false'
         uses: actions/github-script@v7
         with:
           script: |
--- a/.github/workflows/pr-build-image.yaml
+++ b/.github/workflows/pr-build-image.yaml
@@ -55,17 +55,26 @@ jobs:
       - name: Check if Image Already Exists
         id: image-check
         run: |
-          IMAGE_TAG="pr-${{ github.event.number }}-${{ env.SHORT_SHA }}"
-          IMAGE_NAME="${{ env.REGISTRY }}/rhdh-community/rhdh:${IMAGE_TAG}"
+          IMAGE_TAG_PR="pr-${{ github.event.number }}"
+          IMAGE_TAG_COMMIT="pr-${{ github.event.number }}-${{ env.SHORT_SHA }}"
 
-          # Check if an image with the exact tag exists
-          IMAGE_TAGS=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG" '.tags[] | select(.name == $tag) | .name')
+          IMAGE_NAME_PR="${{ env.REGISTRY }}/rhdh-community/rhdh:${IMAGE_TAG_PR}"
+          IMAGE_NAME_COMMIT="${{ env.REGISTRY }}/rhdh-community/rhdh:${IMAGE_TAG_COMMIT}"
 
-          if [ -n "$IMAGE_TAGS" ]; then
-            echo "Image $IMAGE_NAME already exists for the current commit."
+          # Check if any image exists for the PR
+          IMAGE_EXISTS_PR=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG_PR" '.tags[] | select(.name == $tag) | .name')
+
+          # Check if any image exists for the specific commit
+          IMAGE_EXISTS_COMMIT=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG_COMMIT" '.tags[] | select(.name == $tag) | .name')
+
+          if [ -n "$IMAGE_EXISTS_COMMIT" ]; then
+            echo "Image $IMAGE_NAME_COMMIT already exists for the current commit."
             echo "image_exists=true" >> $GITHUB_ENV
+          elif [ -n "$IMAGE_EXISTS_PR" ]; then
+            echo "Image $IMAGE_NAME_PR exists for the PR, but not for the current commit."
+            echo "image_exists=false" >> $GITHUB_ENV
           else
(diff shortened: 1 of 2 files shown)


04kash (approved):

.github/workflows/pr-build-image.yaml
```diff
@@ -92,25 +92,8 @@ jobs:
...
-            echo "Image $IMAGE_NAME or its variants already exist. Skipping the build."
-            echo "image_exists=true" >> $GITHUB_ENV
-          else
-            echo "Image $IMAGE_NAME does not exist. Proceeding with the build."
-            echo "image_exists=false" >> $GITHUB_ENV
-          fi
-
       - name: Build and Push with Buildx
-        if: env.proceed_with_build == 'true' && env.image_exists == 'false'
+        if: env.relevant_changes == 'true' || env.image_exists == 'false'
```
04kash, line 96:
```suggestion
        if: env.relevant_changes == 'true' || env.image_exists == 'false'
```
Shouldn't this be the same as before? Wouldn't we want a new image only when we have relevant changes and the image corresponding to these changes doesn't exist?

gustavolira: The current logic ensures that a new image is only built when there are relevant changes (proceed_with_build == 'true') and an image corresponding to these changes does not already exist (image_exists == 'false). This avoids unnecessary builds while still ensuring that new changes result in a fresh image. Altering this logic would either trigger unnecessary builds or skip builds when they are actually required.

04kash: Yeah, so shouldn't we still be using && instead of || ?

.github/workflows/pr-build-image.yaml
```diff
@@ -92,25 +92,8 @@ jobs:
       - name: Get the last commit short SHA of the PR
         uses: ./.github/actions/get-sha
 
-      - name: Check if Image Already Exists
-        if: env.proceed_with_build == 'true'
-        run: |
-          IMAGE_TAG="pr-${{ github.event.number }}"
```
04kash, line 98: Can we change IMAGE_TAG="pr-${{ github.event.number }}" to IMAGE_TAG="pr-${{ github.event.number }}-${{ env.SHORT_SHA }}" so that when new commits are added to a PR, we get a new image?

.github/workflows/pr-build-image.yaml
```diff
@@ -92,25 +92,8 @@ jobs:
...
-          IMAGE_TAGS=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG" '.tags[] | select(.name | startswith($tag)) | .name')
-
-          if [ -n "$IMAGE_TAGS" ]; then
-            echo "Image $IMAGE_NAME or its variants already exist. Skipping the build."
-            echo "image_exists=true" >> $GITHUB_ENV
-          else
-            echo "Image $IMAGE_NAME does not exist. Proceeding with the build."
-            echo "image_exists=false" >> $GITHUB_ENV
-          fi
-
```
04kash, line 111: Can we keep this one, and remove the 1st mention of `Check if Image Already Exists` on line 55?

.github/workflows/pr-build-image.yaml
```diff
@@ -55,17 +55,26 @@ jobs:
...
+            echo "Image $IMAGE_NAME_COMMIT already exists for the current commit."
             echo "image_exists=true" >> $GITHUB_ENV
+          elif [ -n "$IMAGE_EXISTS_PR" ]; then
+            echo "Image $IMAGE_NAME_PR exists for the PR, but not for the current commit."
+            echo "image_exists=false" >> $GITHUB_ENV
           else
-            echo "Image $IMAGE_NAME does not exist."
+            echo "No image exists for the PR or the current commit."
             echo "image_exists=false" >> $GITHUB_ENV
           fi
```
04kash, line 79: You might have to place this after the `Get the last commit short SHA of the PR` step

.github/workflows/pr-build-image.yaml
```diff
@@ -55,17 +55,26 @@ jobs:
...
+          # Check if any image exists for the PR
+          IMAGE_EXISTS_PR=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG_PR" '.tags[] | select(.name == $tag) | .name')
+
+          # Check if any image exists for the specific commit
+          IMAGE_EXISTS_COMMIT=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh/tag/" | jq -r --arg tag "$IMAGE_TAG_COMMIT" '.tags[] | select(.name == $tag) | .name')
+
+          if [ -n "$IMAGE_EXISTS_COMMIT" ]; then
+            echo "Image $IMAGE_NAME_COMMIT already exists for the current commit."
             echo "image_exists=true" >> $GITHUB_ENV
+          elif [ -n "$IMAGE_EXISTS_PR" ]; then
```
nickboldt, line 73: why check for the floating PR tag, if we need only the `IMAGE_TAG_COMMIT` to exist and therefore `IMAGE_EXISTS_COMMIT` to be true?

.github/workflows/pr-build-image.yaml
nickboldt: Yes, you do. See https://github.com/redhat-developer/rhdh/blob/main/.github/actions/get-sha/action.yml#L41

.github/workflows/pr-build-image.yaml
gustavolira: Hi @04kash, you're absolutely correct! Using IMAGE_TAG="pr-${{ github.event.number }}-${{ env.SHORT_SHA }}" ensures that a unique image is created for each commit, avoiding the reuse of the same image across multiple commits in the same PR. I've updated the workflow to reflect this change. Thank you for pointing this out

.github/workflows/pr-build-image.yaml
gustavolira: Hi @04kash, I agree that having two separate "Check if Image Already Exists" steps is redundant. I've consolidated the logic into a single step, which will come after "Get the last commit short SHA of the PR." This simplifies the workflow and ensures the process is streamlined.

.github/workflows/pr-build-image.yaml
gustavolira: he SHORT_SHA variable is required for the image tag, so the "Check if Image Already Exists" step must come after "Get the last commit short SHA of the PR." I've updated the workflow to ensure the correct sequence. Great catch!

.github/workflows/pr-build-image.yaml
gustavolira: Hi @nickboldt, great point! Checking only IMAGE_TAG_COMMIT simplifies the logic and ensures we only focus on commit-specific images. I've updated the workflow to remove the redundant check for the floating PR tag (IMAGE_TAG_PR). Let me know if you have any additional thoughts on this!

04kash (approved):

nickboldt (approved):
/lgtm

PR #2269, opened 2025-01-28
chore: use local files instead of latest from main branch

christoph-jerolimov:
## Description

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/app-config.yaml
+++ b/app-config.yaml
@@ -150,8 +150,15 @@ catalog:
   rules:
     - allow: [Component, System, Group, Resource, Location, Template, API]
   locations:
-    - type: url
-      target: https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/all.yaml
+    - type: file
+      target: ../../catalog-info.yaml
+      rules:
+        - allow: [Component, System]
+
+    - type: file
+      target: ../../catalog-entities/all.yaml
+      rules:
+        - allow: [Component, System, Group, Resource, Location, Template, API]
 
     - type: url
       target: https://github.com/redhat-developer/red-hat-developer-hub-software-templates/blob/main/templates.yaml


app-config.yaml
```diff
@@ -150,8 +150,15 @@ catalog:
   rules:
     - allow: [Component, System, Group, Resource, Location, Template, API]
   locations:
-    - type: url
-      target: https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/all.yaml
+    - type: file
+      target: ../../catalog-info.yaml
```
logonoff, line 154: when this app-config is being parsed, is the CWD in `./packages/app`? Just making sure that it's okay to put  `../../` in a file that is already on the root repo directory

christoph-jerolimov: Hey, the default is `packages/backend`, the default Backstage scaffolder contains a comment like this:

https://github.com/christoph-jerolimov/backstage-history/blob/main/app-config.yaml#L80-L95

That was removed here. I think this app-config.yaml is only used in local development. Production installation will not see all these "test entities".

But it's maybe worthful to check that again with someone that is more involved into the build process.

app-config.yaml
```diff
@@ -150,8 +150,15 @@ catalog:
...
+      rules:
+        - allow: [Component, System]
+
+    - type: file
+      target: ../../catalog-entities/all.yaml
+      rules:
+        - allow: [Component, System, Group, Resource, Location, Template, API]
```
ciiay, line 161: If we want to allow all entity types here, we can remove the rules right?

ciiay (approved):
The changes look good and verified working. Just want to understand why we want to use the local file instead of latest main branch file 🤔

logonoff (approved):
/lgtm

app-config.yaml
```diff
@@ -150,8 +150,15 @@ catalog:
...
   locations:
-    - type: url
-      target: https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/all.yaml
+    - type: file
+      target: ../../catalog-info.yaml
+      rules:
+        - allow: [Component, System]
+
+    - type: file
+      target: ../../catalog-entities/all.yaml
```
kadel, line 159: When testing this change, I get 
```
{"entity":"location:default/generated-28cca00eef20648da516bda20fe0e71b7b7364ba","level":"\u001b[33mwarn\u001b[39m","location":"file:/opt/catalog-entities/all.yaml","message":"file /opt/catalog-entities/all.yaml does not exist","plugin":"catalog","service":"backstage","timestamp":"2025-02-26 11:24:42"}
```

I think that it should not go two-levels up (`../../`).

When I tried to search for `all.yaml` or `catalog-info.yaml` in the container image I haven't found anything. This might require updating Dockerfile to add the files into the  container iamge

PR #2284, opened 2025-01-29
chore: linking index.html to index.html.tmpl

coreydaley:
## Description

The existence of the index.html.tmpl tells Backstage to take that file and load it into memory, inject the needed appConfig (instead of writing it into a random *.chunk.js file, and then serve that file from memory so that we can set the readOnlyRootFilesystem: true option for the RHDH container, further improving the security of our offering.

https://issues.redhat.com/browse/RHIDP-5761

## PR acceptance criteria


nickboldt:
Concept: +1. Please update the Dockerfile in .rhdh/ too so the same fix lands downstream.

PR #2199, opened 2025-01-16
feat(fab): add new mountpoints support for providers and listeners

debsmita1:
## Description:
https://issues.redhat.com/browse/RHIDP-5478
https://issues.redhat.com/browse/RHIDP-5399

### Testing in your local

- In your local `rhdh` repo, add the following snippet in your `app-config.yaml` file
```
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
  frontend:
    red-hat-developer-hub.backstage-plugin-bulk-import:
      mountPoints:
         - mountPoint: global.floatingactionbutton/component
          importName: BulkImportPage
          config:
            slot: 'page-end'
            icon: bulkImportIcon
            label: 'Bulk import'
            toolTip: 'Register multiple repositories in bulk'
            to: /bulk-import/repositories
            excludeOnPaths: ['/catalog']
      appIcons:
        - name: bulkImportIcon
          importName: BulkImportIcon
      dynamicRoutes:
        - path: /bulk-import/repositories
          importName: BulkImportPage
          menuItem:
            icon: bulkImportIcon
            text: Bulk import
    red-hat-developer-hub.backstage-plugin-global-floating-action-button:
      mountPoints:
        - mountPoint: application/listener
          importName: DynamicGlobalFloatingActionButton
        - mountPoint: global.floatingactionbutton/component
          importName: NullComponent
          config:
            icon: github
            label: 'Git'
            toolTip: 'Github'
            to: https://github.com/redhat-developer/rhdh
```

- In your local `rhdh` repo, run `yarn install` followed by `yarn dev`

### Testing on a cluster

```
dynamic-plugins.yaml: |
    includes:
    - dynamic-plugins.default.yaml
    plugins:
      - package: '@red-hat-developer-hub/backstage-plugin-application-provider-test@0.0.1'
        integrity: 'sha512-4k/D6YjVfLX8jiCOlZJ+CG0hcE55hk2UxyKO/SwsRWc+qjwuny1A+MO0vz+KgKcwkFvg1tw7kkPCjV8w1k4lUg=='
        pluginConfig:
          dynamicPlugins:
            frontend:
              red-hat-developer-hub.backstage-plugin-application-provider-test:
                dynamicRoutes:
                  - path: /countpage
                    importName: CountPage
                mountPoints:
                  - mountPoint: application/provider
                    importName: CountProvider
      - package: ./dynamic-plugins/dist/red-hat-developer-hub-backstage-plugin-bulk-import
        disabled: false
        pluginConfig:
[...]

PR #2312, opened 2025-02-04
feat(fab): add new mountpoints support for providers and listeners (fixed CI issue)

christoph-jerolimov:
Same as #2199 but rebased andresolve yarn.lock merge conflict by just run it again.

After rebasing the PR it was required to update the backstage/cli devDependency version in `dynamic-plugins/wrappers/red-hat-developer-hub-backstage-plugin-global-header/package.json` to the same version that all plugins use.

/cc @debsmita1 @ciiay


ciiay (approved):
/lgtm

Thank you!

PR #2263, opened 2025-01-27
feat(global-header): added config for default header components

ciiay:
## Description

- Added config for default application header components
- Added extra common icons
- ~Updated support to "tech-preview"~
- Updated support to "production" as the new plugin version will be ready for production

## Which issue(s) does this PR fix

- Fixes [RHIDP-5252](https://issues.redhat.com/browse/RHIDP-5252)

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [x] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [x] Add a screenshot if the change is UX/UI related

## How to test changes / Special notes to the reviewer
1. Pull changes in [rhdh-plugins PR](https://github.com/redhat-developer/rhdh-plugins/pull/373) to your local and in `workspaces/global-header/plugins/global-header/` directory run `npx --yes @janus-idp/cli package export-dynamic-plugin --dynamic-plugins-root <path-to-your-rhdh>/dynamic-plugins-root --dev`
2. Pull changes in [theme PR](https://github.com/redhat-developer/red-hat-developer-hub-theme/pull/53) to your local 
3. Pull changes in this PR to your local rhdh, in `packages/app/package.json`, replace `"@redhat-developer/red-hat-developer-hub-theme": "0.4.0",` with `"@redhat-developer/red-hat-developer-hub-theme": "file:path/to/your/red-hat-developer-hub-theme",`
4. In your rhdh `app-config.local` file:
<details>
<summary>Config</summary>
```
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
  frontend:
    default.main-menu-items:
      menuItems:
        default.create:
          title: ''
    red-hat-developer-hub.backstage-plugin-global-header:
      mountPoints:
        - mountPoint: application/header
          importName: GlobalHeader
          config:
            position: above-main-content # above-main-content | below-main-content
        - mountPoint: global.header/component
          importName: SearchComponent
          config:
            priority: 100
        - mountPoint: global.header/component
          importName: Spacer
          config:
            priority: 99
            props:
              growFactor: 0
        - mountPoint: global.header/component
          importName: HeaderIconButton
          config:
            priority: 90
            props:
              title: Create...
              icon: add
              to: create
        - mountPoint: global.header/component
          importName: SupportButton
          config:
            priority: 80
        - mountPoint: global.header/component
          importName: NotificationButton
          config:
[...]

PR #2384, opened 2025-02-14
fix: align application header mountpoint with other mount points

christoph-jerolimov:
## Description

Mount point `config.layout` is for catalog cards used to apply CSS properties to the surrounding Box. So I think its better to use `config.position` instead of `config.layout.position` for this.

It supports in this first version also the old attribute. This might not be necessary.

I also applied the position as prop to the header component. Just for the case that a header needs this.

- Fixes #?

## PR acceptance criteria


ciiay (approved):
/lgtm

I have cherry-picked this PR into #2263 . Thank you!

PR #2385, opened 2025-02-14
chore(e2e): techdocs addons tests

dzemanov:
## Description

Introduces techdocs addon tests for ReportIssue.
Adds entity tab techdocs test.

- Fixes [RHIDP-5223](https://issues.redhat.com/browse/RHIDP-5223)

## PR acceptance criteria

### DIFF
--- a/.ibm/pipelines/value_files/sanity-check-plugins.yaml
+++ b/.ibm/pipelines/value_files/sanity-check-plugins.yaml
@@ -91,6 +91,14 @@ global:
         disabled: false
       - package: ./dynamic-plugins/dist/backstage-plugin-techdocs
         disabled: false
+      - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-module-addons-contrib
+        disabled: false
+        pluginConfig:
+          dynamicPlugins:
+            frontend:
+              backstage.plugin-techdocs-module-addons-contrib:
+                techdocsAddons:
+                  - importName: ReportIssue
       - package: ./dynamic-plugins/dist/pagerduty-backstage-plugin
         disabled: false
       - package: ./dynamic-plugins/dist/pagerduty-backstage-plugin-backend-dynamic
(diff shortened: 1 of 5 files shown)


.ibm/pipelines/value_files/sanity-check-plugins.yaml
```diff
@@ -91,6 +91,14 @@ global:
         disabled: false
       - package: ./dynamic-plugins/dist/backstage-plugin-techdocs
         disabled: false
+      - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-module-addons-contrib
```
zdrapela, line 94: Hello, I'm confused about the `sanity-check-plugins.yaml` value file for the deployment. I think this is only used in nightly job in a separate deployment (that @gustavolira recently re-introduced here https://github.com/redhat-developer/rhdh/pull/2248), but I see that `techdocs-backend-dynamic` plugin is also here. But I'm unsure how the tests passed here. Maybe I'm missing something 😅

dzemanov: I thought `techdocs-backend-dynamic` and `techodocs` are enabled for the deployment as they have [here](https://github.com/redhat-developer/rhdh/blob/8454e1d6ce892056a0f8b5b5d5924a3e5c618889/.ibm/pipelines/value_files/sanity-check-plugins.yaml#L90-#L93) `disabled: false`. Is the config taken from somewhere else then? I added this to the sanity check as the plugin with ReportIssue is enabled by default in RHDH. I can remove it.

zdrapela: The `sanity-check-plugins.yaml` is only used here for another deployment that runs only on nightly: https://github.com/redhat-developer/rhdh/blob/23e5e483e3d5c89e2be08ae3fcbf7dbede0a43a2/.ibm/pipelines/utils.sh#L813 
Please add the plugin into https://github.com/redhat-developer/rhdh/blob/main/.ibm/pipelines/value_files/values_showcase.yaml
If the plugin is enabled by default, that explains why the tests are passing now.

dzemanov: Yeah, I didn't add the config [here](https://github.com/redhat-developer/rhdh/blob/main/.ibm/pipelines/value_files/values_showcase.yaml) as I thought it's not needed for the defaultly enabled, as techdocs are not here either. But will add it to make it explicit. Thank you @zdrapela.

PR #2451, opened 2025-02-25
chore(e2e): add RHSSO sub claim resolvers to auth e2e tests

JessicaJHee:
## Description

Adds e2e tests for the `oidcSubClaimMatchingKeycloakUserId` OIDC auth resolver
* sets this resolver as the default one for RHSSO auth e2e tests

- Fixes [RHIDP-6322](https://issues.redhat.com/browse/RHIDP-6322)

## PR acceptance criteria


JessicaJHee:
@albarbaro I did notice that on the first run of `RHBK - remove a user from RHDH` it fails on this on the first run:
https://github.com/redhat-developer/rhdh/blob/836cc5e5392511e713acba17606375d415905ebf/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts#L609

But passes on the retry. Any ideas on why this might be?

albarbaro:
> @albarbaro I did notice that on the first run of `RHBK - remove a user from RHDH` it fails on this on the first run:
> 
> https://github.com/redhat-developer/rhdh/blob/836cc5e5392511e713acba17606375d415905ebf/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts#L609
> 
> But passes on the retry. Any ideas on why this might be?

I did notice the same recently but I wasn't able to figure it out yet. My guess is this is due to a sync issue after the previous test cases are run, since on retry the users are re-created and the sync happens again, but I haven't verified it yet.

kim-tsao (commented):
Just a small nit.  If this change is intended for 1.6 and we have plans to remove RHSSO 7.6, does it make sense to remove the test cases as well?

JessicaJHee:
@kim-tsao From my understanding the tests are already using RHBK but the code/names of constants need to be updated to make this clear.

The check is done [here](https://github.com/redhat-developer/rhdh/blob/main/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts#L77-L84), and it only runs the `RHBK` tests:
https://github.com/redhat-developer/rhdh/blob/ca0c16bfb181dac0787c55fb841fbf79a212b705/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts#L22

I think it'd be a good idea to clean this up in a separate PR. @albarbaro Do you mind confirming? WDYT?

albarbaro:
That's correct, RHBK tests are the only ones running, but the filename and the variables still have to be updated. 
I have a refactoring PR planned and this is something that will be targeted too.

PR #2575, opened 2025-03-13
chore(e2): rhidp-6404  Add image check and relevant changes action to workflows

gustavolira:
https://issues.redhat.com/browse/RHIDP-6404

## Description

- Fixes #?

## PR acceptance criteria


.github/actions/check-image-and-changes/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+    value: ${{ steps.sha.outputs.short_sha }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get the last commit short SHA of the PR
+      id: sha
+      shell: bash
+      run: |
+        SHORT_SHA=$(git rev-parse --short HEAD)
```
nickboldt, line 21: probably better to reuse this action for consistency https://github.com/redhat-developer/rhdh/tree/main/.github/actions/get-sha

.github/workflows/pr-build-image.yaml
```diff
@@ -52,69 +52,41 @@ jobs:
...
               repo: context.repo.repo,
-              body: 'The image is available at:\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}) or\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }})'
+              body: 'The image is available at:\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}) or\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ steps.check-image.outputs.short_sha }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ steps.check-image.outputs.short_sha }})'
             })
+
+      - name: Skip build notification
+        if: ${{ steps.check-image.outputs.relevant_changes == 'false' && steps.check-image.outputs.image_exists == 'true' }}
+        run: |
+          echo "Skipping build: Image already exists and no relevant changes detected."
+
```
nickboldt, line 91:
```suggestion
```

.github/workflows/pr.yaml
```diff
@@ -26,9 +26,47 @@ concurrency:
...
+          cache-prefix: ${{ runner.os }}-v22
+
+      - name: Run Prettier
+        run: yarn prettier:check --continue --affected
+
+      - name: Run Lint
+        run: yarn run lint:check --continue --affected
+
   build:
     name: Build with Node.js ${{ matrix.node-version }}
```
nickboldt, line 66: this is a required test. Can we run the step but have it immediately exit 0 if there's nothing to do?

image

.github/workflows/pr.yaml
```diff
@@ -61,6 +99,8 @@ jobs:
   test:
     name: Test with Node.js ${{ matrix.node-version }}
```
nickboldt, line 100: this is a required test. Can we run the step but have it immediately exit 0 if there's nothing to do?

image

.github/actions/check-image-and-changes/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+  steps:
+    - name: Get the last commit short SHA of the PR
+      id: get-sha
+      uses: ./.github/actions/get-sha  # Chamando a action existente
+
+    - name: Check if Image Already Exists
+      id: image-check
+      shell: bash
+      run: |
+        IMAGE_TAG_COMMIT="pr-${{ github.event.number }}-${{ steps.get-sha.outputs.short_sha }}"
```
nickboldt, line 26: since the get-sha action commits an env var, you can just use...

```suggestion
        IMAGE_TAG_COMMIT="pr-${{ github.event.number }}-${{ env.SHORT_SHA }}"
```

.github/workflows/pr-build-image.yaml
```diff
@@ -52,69 +57,37 @@ jobs:
...
-        if: env.relevant_changes == 'true' && env.image_exists == 'false'
         uses: ./.github/actions/docker-build
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
           imageName: rhdh-community/rhdh
           imageTags: |
-            type=ref,prefix=pr-,suffix=-${{ env.SHORT_SHA }},event=pr
+            type=ref,prefix=pr-,suffix=-${{ needs.check-image.outputs.short_sha }},event=pr
```
nickboldt, line 68:
```suggestion
            type=ref,prefix=pr-,suffix=-${{ env.SHORT_SHA }},event=pr
```

.github/workflows/pr-build-image.yaml
```diff
@@ -52,69 +57,37 @@ jobs:
...
-        if: env.relevant_changes == 'true' && env.image_exists == 'false'
         uses: actions/github-script@v7
         with:
           script: |
             github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'The image is available at:\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}) or\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }})'
+              body: 'The image is available at:\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}) or\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ needs.check-image.outputs.short_sha }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ needs.check-image.outputs.short_sha }})'
```
nickboldt, line 82:
```suggestion
              body: 'The image is available at:\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}) or\n* [`quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }}`](https://quay.io/rhdh-community/rhdh:pr-${{ github.event.number }}-${{ env.SHORT_SHA }})'
```

.github/workflows/pr.yaml
```diff
@@ -60,30 +98,38 @@ jobs:
 
   test:
     name: Test with Node.js ${{ matrix.node-version }}
+    needs: check-image
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [22]
+        node-version: [ 20, 22 ]
```
nickboldt, line 105: Why are we re-adding node 20?

```suggestion
        node-version: [ 22 ]
```

.github/workflows/pr.yaml
```diff
@@ -60,30 +98,38 @@ jobs:
...
 
+      - name: Skip Test if No Relevant Changes
+        run: |
+          if [[ "${{ needs.check-image.outputs.relevant_changes }}" == "false" ]]; then
+            echo "Skipping tests: No relevant changes detected."
+            exit 0
+          fi
+
       - name: Setup Node.js
-        uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4
```
nickboldt, line 74: these changes will just trigger renovate updates. Please revert.

.github/workflows/pr.yaml
```diff
@@ -26,9 +26,47 @@ concurrency:
...
+        run: yarn prettier:check --continue --affected
+
+      - name: Run Lint
+        run: yarn run lint:check --continue --affected
+
   build:
     name: Build with Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
+    needs: check-image
+    if: ${{ needs.check-image.outputs.relevant_changes == 'true' || needs.check-image.outputs.image_exists == 'false' }}
```
nickboldt, line 69: looks like the build step will be completely skipped here. can we do a similar check as you have in pr-build-image.yaml line 61? https://github.com/redhat-developer/rhdh/pull/2575/files#diff-2b0fff9a121c4bf67a04b4a82b44d2f2a5cc48a69d8c299b46072978d7cf1212R61

.github/workflows/pr-build-image.yaml
```diff
@@ -52,48 +56,14 @@ jobs:
...
-          # Check if changes are only in .ibm/ or e2e-tests/
-          if echo "$CHANGED_FILES" | grep -qvE '^(e2e-tests/|\.ibm/)'; then
-            echo "Changes detected outside the e2e-tests or .ibm folders. Build required."
-            echo "relevant_changes=true" >> $GITHUB_ENV
-          else
-            echo "No relevant changes detected."
-            echo "relevant_changes=false" >> $GITHUB_ENV
+          if [[ "${{ needs.check-image.outputs.relevant_changes }}" == "false" && "${{ needs.check-image.outputs.image_exists }}" == "true" ]]; then
+            echo "Skipping build: Image already exists and no relevant changes detected."
+            exit 0
```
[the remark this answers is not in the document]
gustavolira: @nickboldt  The skip-build job is just a notification step and does not control execution flow. The build-image job already has an if: condition that ensures it only runs if relevant changes exist or the image does not exist. So, this workflow will not proceed with an unnecessary build.

.github/workflows/pr.yaml
gustavolira: @nickboldt  You're right! The current logic would completely skip the build step if no relevant changes are detected and the image already exists.

I have updated the workflow to ensure that the job always runs (if: always()) and added an explicit check inside the first step. If the conditions indicate that no build is needed, the job exits early with exit 0, similar to what we did in pr-build-image.yaml at line 61.

This way, the workflow structure remains intact, and we avoid unnecessary builds while still ensuring required jobs are executed properly.

Let me know if you have any concerns! Thanks for the feedback.

.github/workflows/pr.yaml
```diff
@@ -26,9 +26,47 @@ concurrency:
...
+        run: yarn prettier:check --continue --affected
+
+      - name: Run Lint
+        run: yarn run lint:check --continue --affected
+
   build:
     name: Build with Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
+    needs: check-image
+    if: always()
```
nickboldt, line 69: do we need this line if the check-if-skip is on line 79?

.github/workflows/pr-build-image.yaml
```diff
@@ -27,8 +13,27 @@ env:
...
+          fetch-depth: 0
+
+      - name: Check Image and Relevant Changes
+        id: check-image
+        uses: ./.github/actions/check-image-and-changes
+
   build-image:
     name: Build Image
+    needs: check-image
+    if: ${{ needs.check-image.outputs.relevant_changes == 'true' || needs.check-image.outputs.image_exists == 'false' }}
```
nickboldt, line 36: this is a required test, so the same "must run and if skipped should exit 0" rule applies here too. 

image

.github/workflows/pr-build-image.yaml
nickboldt, line 68: since the check-image stage sets `${{ env.SHORT_SHA }}` why not just use that env?

.github/workflows/pr.yaml
```diff
@@ -26,9 +26,47 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    outputs:
+      relevant_changes: ${{ steps.check-image.outputs.relevant_changes }}
+      image_exists: ${{ steps.check-image.outputs.image_exists }}
+    steps:
+      - name: Run image & changes check
+        uses: ./.github/actions/check-image-and-changes
+
+  linting:
+    name: Run Linting & Prettier
```
nickboldt, line 40:
```suggestion
    name: Checkout, install deps, run Linting & Prettier
```

.github/workflows/pr-build-image.yaml
```diff
@@ -1,17 +1,3 @@
-# Copyright Red Hat, Inc.
```
nickboldt, line 1: bit weird to remove a copyright statement... but I guess if you have a reason for this...?

.github/workflows/pr.yaml
gustavolira: @nickboldt The if: always() condition isn't needed since we already have a step inside the job that checks if it should be skipped. I'll remove always() and rely on the existing logic to avoid unnecessary executions. Thanks for pointing it out!

nickboldt (commented):
Only thing I don't like is that we're now doing the same GH checkout twice per action (first to check if we have to build, then to do the build)

:thinking: 

if we moved the call to `.github/actions/check-image-and-changes` out of the `check-image` job and into the `build-image` job, then you might be able to reuse the same checkout for the build, and the step `Check if build should be skipped` would look at  variable `build-image.check-image.outputs.relevant_changes` instead of `needs.check-image.outputs.relevant_changes` ?

nickboldt:
Might have a typo somewhere... 

image

https://github.com/redhat-developer/rhdh/actions/runs/13928032057/job/38977666439 

why was the job skipped? why is it incorrectly formatted?

.github/workflows/pr.yaml
```diff
@@ -105,50 +72,43 @@ jobs:
...
       - name: Use app-config.example.yaml
         run: rm app-config.yaml && mv app-config.example.yaml app-config.yaml
 
       - name: Install dependencies
-        uses: backstage/actions/yarn-install@b3c1841fd69e1658ac631afafd0fb140a2309024 # v0.6.17
+        uses: backstage/actions/yarn-install@v0.6.17
         with:
           cache-prefix: ${{ runner.os }}-v${{ matrix.node-version }}
 
-      - name: Run monorepo tools
```
nickboldt, line 96: why did you remove this and the dockerfile regen check? please explain in the issue description -- this is a new change today that was not in the original PR, where you only removed the prettier and linter checks. 

https://github.com/redhat-developer/rhdh/pull/2575/commits/373dac41c690909770704703f9ebfc1b09a30a52#diff-1eb4e5fd5611777d4e597ef299a1cb5ba8050c28a2dabbd4fbc56205d69e5ddaL90

gustavolira:
> Only thing I don't like is that we're now doing the same GH checkout twice per action (first to check if we have to build, then to do the build)
> 
> 🤔
> 
> if we moved the call to `.github/actions/check-image-and-changes` out of the `check-image` job and into the `build-image` job, then you might be able to reuse the same checkout for the build, and the step `Check if build should be skipped` would look at variable `build-image.check-image.outputs.relevant_changes` instead of `needs.check-image.outputs.relevant_changes` ?

Thanks for the suggestion! We’ve implemented the improvement by moving the .github/actions/check-image-and-changes call directly into the build-image job. This allows us to reuse the same checkout step, avoiding duplicate operations. Now, the Check if build should be skipped step references build-image.check-image.outputs.relevant_changes instead of needs.check-image.outputs.relevant_changes

.github/workflows/pr.yaml
gustavolira: Apologies for the removal, it was unintentional. I got a bit lost during the refactoring while juggling the test day and updating the PR at the same time. In the process, I mistakenly removed the Dockerfile regeneration check. I've now updated the PR and restored everything that was removed. Thanks for pointing it out!

.github/workflows/pr-build-image.yaml
```diff
@@ -43,6 +43,17 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
 
+      - name: Check Image and Relevant Changes
+        id: check-image
+        uses: ./.github/actions/check-image-and-changes
+
+      - name: Check if build should be skipped
```
nickboldt, line 50: a further simplication could be to move this check into the action itself.

then the output of the `check-image-and-changes` is not just two booleans, but an exit code of 0 if the job can be terminated right there. 

Of course your logged message would need to be something like

```
            echo "::notice::Skipping job: Image already exists and no relevant changes detected."
```

nickboldt:
Did a quick google for how to exit a GH workflow if something fails and it seems we won't get the results we want with an exit 0:

https://stackoverflow.com/questions/60589373/how-to-force-job-to-exit-in-github-actions-step

Instead of exiting 0, we might have to label all the subsequent steps with a conditional check to see if they should run...

```
steps.check.conclusion == 'failure'
```

gustavolira:
> Did a quick google for how to exit a GH workflow if something fails and it seems we won't get the results we want with an exit 0:
> 
> https://stackoverflow.com/questions/60589373/how-to-force-job-to-exit-in-github-actions-step
> 
> Instead of exiting 0, we might have to label all the subsequent steps with a conditional check to see if they should run...
> 
> ```
> steps.check.conclusion == 'failure'
> ```

Thanks for the suggestion! We explored that direction, but unfortunately GitHub Actions doesn't currently support terminating a job from within a composite action using `exit 0`—it only ends the current step, not the job. This is also documented in [this StackOverflow thread](https://stackoverflow.com/questions/60589373/how-to-force-job-to-exit-in-github-actions-step) which you sent

To maintain clarity and avoid repeating `if` conditions on every step, we’ve kept the check inside the first step of the job (`Check if build should be skipped`). This gives us the flexibility to skip the entire job cleanly while still keeping the logic centralized in the action.

That said, we're definitely keeping this idea in mind if GitHub Actions evolves to support true early exits!

.github/actions/check-image-and-changes/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+  short_sha:
+    description: "Short SHA of the latest commit"
+    value: ${{ steps.get-sha.outputs.short_sha }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get the last commit short SHA of the PR
+      id: get-sha
+      uses: ./.github/actions/get-sha  # Chamando a action existente
```
nickboldt, line 20:
```suggestion
      uses: ./.github/actions/get-sha  # Use existing action
```

nickboldt:
> To maintain clarity and avoid repeating if conditions on every step, we’ve kept the check inside the first step of the job (Check if build should be skipped). This gives us the flexibility to skip the entire job cleanly while still keeping the logic centralized in the action.

Makes sense, as long as the check returning 0 DOES actually end the job. I guess we can merge this and confirm it does what we want.

PR #2594, opened 2025-03-17
feat(defaultAppComponents): add custom 404 page

logonoff:
## Description

before:
image

after:
image

- Fixes https://issues.redhat.com/browse/RHIDP-1317

## PR acceptance criteria


logonoff:
bump

packages/app/src/components/ErrorPages/NotFoundErrorPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+  <ErrorPage
+    title={
+      <>
+        <strong>404</strong> We couldn't find that page
+      </>
+    }
+    message={
+      <>
+        Try adding an <strong>index.md</strong> file in the root of the docs
+        directory of this repository.
```
christoph-jerolimov, line 19: Hmm, this is a generic error page but it talks about tech docs.

We should have a more generic message here.

If you want have a message for techdocs as well, then you can check the current pathname of course, but we should show this only when we're sure that it's a "good recommendation".

This messages could be used for example if the path matches one of these two:

* /docs/...
* /catalog/:namespace/:kind/:name/docs/...

cc @ShiranHi

christoph-jerolimov (changes requested):

logonoff:
bump

logonoff:
bump!

ShiranHi:
Looks good to me, thank you!

packages/app/src/components/ErrorPages/illustrations/illustration.tsx
```diff
@@ -0,0 +1,28 @@
...
+        src={isDarkMode ? dark : light}
+        // presentational image
+        alt=""
+        aria-hidden="true"
+        {...props}
+      />
+    );
+  };
+
+  Illustration.displayName = displayName || 'Illustration';
```
christoph-jerolimov, line 25: It's not a big issue, but this `displayName` arg and need to set a displayName is a hint for me, that the composition order could be easier (in the other direction).

Like:

NotFoundErrorPage use CollaborationIllustration which could use Illustration with two props.

But I don't want delay this PR more, so let us merge it, its not wrong and actually "just" used on an error page.

christoph-jerolimov (approved):
Thanks @logonoff. This is a great improvement.

I like all the small details!

1. The Back button that is only displayed when it does something
2. The Support link that picks up the support link
3. That the design is working 

I'm sorry that this review takes too much time! :crying_cat_face: 

Lgtm!

PR #2637, opened 2025-03-26
chore(marketplace): various catalog metadata fixes

benwilcock:
## Description

Using the CSV created by Jackson Lee, I was able to spot a number of minor inconsistencies and missing data items in the various YAML files used to provide content to the RHDH Extensions Catalog plugin.

One notable change is that the Azure Container Registry plugin had it's support level changed from production down to tech-preview. To the best of mu knowledge based on the data I have this is correct, but you would be wise to double check this with the build manager and product owner.

### DIFF
--- a/catalog-entities/marketplace/plugins/3scale.yaml
+++ b/catalog-entities/marketplace/plugins/3scale.yaml
@@ -69,7 +69,6 @@ spec:
 
   categories:
     - API Management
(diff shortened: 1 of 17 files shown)


christoph-jerolimov (approved):
Thanks Ben, lgtm!

/lgtm
/approve
/cherry-pick release-1.5

PR #2694, opened 2025-04-03
chore: set default OIDC default resolver to resolve by sub claim

JessicaJHee:
## Description

Sets the default OIDC default resolver to resolve by the `sub` claim

- This sub claim is more stable and secure compared parameters used by existing resolvers that are based on mutable values (like user name and email) 
- This PR makes it so that the OIDC sign in flow is more secure by default
- If the user wants to preserve the previous default resolver `emailLocalPartMatchingUserEntityName`, they can still declaratively set this in the `app-config` like: 
```
        signIn:
          resolvers:
            - resolver: emailLocalPartMatchingUserEntityName
```

### Implementation
To accommodate multiple OIDC providers, I created the `oidcSubClaimMatchingIdPUserId` resolver (to be used only in the code, unavailable to customers) that tries both Ping Identity and Keycloak providers. This is done since only 1 resolver is allowed to be declared statically [here](https://github.com/redhat-developer/rhdh/blob/16276135ae56a64b3a79d96f7c7fee7d684715c8/packages/backend/src/modules/authProvidersModule.ts#L186). 

TODO: 

- [x] Update auth e2e tests 

- Fixes [RHIDP-6768](https://issues.redhat.com/browse/RHIDP-6768)

## PR acceptance criteria

- Test that sign in with OIDC works with no resolver declared (will use sub claim resolver by default)
- Test that sign in with OIDC still works with other resolvers when explicitly declared

### DIFF
--- a/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts
+++ b/e2e-tests/playwright/e2e/authProviders/rhsso-76-provider.spec.ts
@@ -147,7 +147,7 @@ for (const version of ["RHBK"]) {
       await common.signOut();
     });
 
-    test(`${version} - default resolver for RHSSO tests is set to oidcSubClaimMatchingKeycloakUserId, user_1 and user_2 should both authenticate`, async () => {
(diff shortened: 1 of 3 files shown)


albarbaro:
Tests are passing locally for me as well

PR #2723, opened 2025-04-09
do not merge: remove all dynamic plugins wrappers

christoph-jerolimov:
## Description

# Do not merge, this is an experiment if we can build an rhdh image without any dynamic plugin wrapper...

## PR acceptance criteria


gashcrumb:
Nice!  Probably this change could include updating [this file](https://github.com/redhat-developer/rhdh/blob/main/dynamic-plugins.default.yaml) to point to the relevant image.

gashcrumb:
Just making some notes here if you don't mind, probably could also get rid of [these](https://github.com/redhat-developer/rhdh/blob/main/package.json#L15-L17) too and therefore [all this](https://github.com/redhat-developer/rhdh/blob/main/docker/Dockerfile#L58-L136) and [these lines](https://github.com/redhat-developer/rhdh/blob/main/docker/Dockerfile#L182-L188)

gashcrumb:
And before I forget [this too](https://github.com/redhat-developer/rhdh/blob/main/.rhdh/docker/Dockerfile#L79-L157) could also be removed and then [this](https://github.com/redhat-developer/rhdh/blob/main/.rhdh/docker/Dockerfile#L210-L216) wouldn't be needed either.  Maybe [this](https://github.com/redhat-developer/rhdh/blob/main/.rhdh/docker/Dockerfile#L220) isn't necessary also

PR #2786, opened 2025-04-22
fix(sidebar): fixing styles break when marketplace plugin drawers opens

its-mitesh-kumar:
## Description : Fixing styles break when marketplace plugin drawers opens 

- Fixes : https://issues.redhat.com/browse/RHIDP-6687

**Before :** 

https://github.com/user-attachments/assets/77882205-1e64-4088-a65c-abd069b21102

**After:** 

https://github.com/user-attachments/assets/d592da1b-fb8f-4bce-83b0-f0456de1a35d


packages/app/src/components/Root/Root.tsx
```diff
@@ -98,6 +98,16 @@ const useStyles = makeStyles()({
   },
   sidebarItem: {
     textDecorationLine: 'none',
+    '& .MuiTypography-root': {
+      color: '#ffffff !important',
+    },
+    '& svg': {
+      color: '#ffffff !important',
+    },
```
christoph-jerolimov, line 106: Sorry, but I'm really skeptical that this is the right approach because this completely ignores the theme. Can you please test it with adjusted sidebar colors.

I guess it would be good to brainstorm the underlying issue with some others from the team.

First: Who sets this extra styles when the drawer is open.

Is it possible override/remove just these styles? Is it maybe an rhdh theme issue?

christoph-jerolimov (changes requested):

packages/app/src/components/Root/Root.tsx
its-mitesh-kumar: @christoph-jerolimov Let me explain the issue in detail:

**Issue**
When opening the `Marketplace Drawer`, which renders components like `<LinkButton />` and `<PluginPackageTable />`, the styles applied to the Backstage Sidebar items (BackstageSidebarItem-root) were unexpectedly overridden.

**Root Cause**
As Emotion injects styles into the DOM in the order components are mounted.
Since the Marketplace Drawer is mounted after the Sidebar and contains components like <LinkButton /> and <PluginPackageTable /> (which internally use <MuiButtonBase />), Emotion injects MUI styles (like .MuiButton-root, .MuiButton-text) after the sidebar styles (e.g., .BackstageSidebarItem-root).
Due to the later injection of styles from Marketplace Drawer, the sidebar styles get overridden unintentionally.

**Solutions Tried**
1. **Scoped Emotion Cache** (CacheProvider with prepend: true)
→ This helped to some extent but didn’t completely solve the issue.

2. **Scoped ThemeProvider with custom overrides for MuiButtonBase**
→ Again, this only partially resolved the styling conflict.

3. **Shadow DOM isolation**
→ Not implemented as it requires introducing a new library and comes with extra overhead.

Final Workaround
Since the above solutions were not sufficient, I opted for direct style overrides using higher specificity and !important to reliably enforce the Sidebar styles . 

Since in both `light` and `dark` theme , color should be white only , that's why ignored the theme . 

Let me know if we should approach it differently .

PR #2742, opened 2025-04-10
[release-1.6] chore(theme): cherry pick theme update

ciiay:
## Description

Please explain the changes you made here.

## Which issue(s) does this PR fix

Manual cherry pick #2741 , closes [RHIDP-6603](https://issues.redhat.com/browse/RHIDP-6603)

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

## How to test changes / Special notes to the reviewer
Test on RHDH local
image

Test on cluster:
image

PR #2799, opened 2025-04-24
fix(self-service): fixing crash issue

its-mitesh-kumar:
## Description 
I noticed that we have multiple versions of @backstage/plugin-catalog-react installed. Here's what I found using:
```
find . -iname plugin-catalog-react | xargs -I {}  sh -c 'echo {}; jq .version {}/package.json'
```
Results:
```
./node_modules/@backstage/plugin-scaffolder/node_modules/@backstage/plugin-catalog-react
"1.16.0"
./node_modules/@backstage/plugin-catalog-react
"1.15.2"
./node_modules/@backstage/plugin-user-settings/node_modules/@backstage/plugin-catalog-react
"1.16.0".
and so on .....
```
To resolve the mismatch, I tried forcing resolution using:
```
"resolutions": {
  "@backstage/plugin-catalog-react": "1.16.0"
}
```
as well as
```
"resolutions": {
  "@backstage/plugin-catalog-react": "1.15.2"
}
```
Both worked. But since we’re using Backstage 1.36.1, we should stick with @backstage/plugin-catalog-react@1.15.2, which aligns with [Backstage's version mapping](https://github.com/backstage/backstage/blob/v1.36.1/plugins/catalog-react/package.json).

- Fixes : https://issues.redhat.com/browse/RHIDP-7173

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
   },
   "resolutions": {
     "@types/react": "18.3.20",
+    "@backstage/plugin-catalog-react": "1.16.0",
     "@types/react-dom": "18.3.6",
     "@aws-sdk/util-utf8-browser": "npm:@smithy/util-utf8@~2",
     "@kubernetes/client-node@npm:0.20.0/jsonpath-plus": "^10.3.0",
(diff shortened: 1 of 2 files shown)


christoph-jerolimov:
@its-mitesh-kumar shared that the issue is resolved in main branch with PR #2742 

For now I think we can close this.

/close

PR #2808, opened 2025-04-25
chore(docs): add script to generate versions.md in dynamic plugins doc

kadel:
## Description

- add script to generate versions.md in dynamic plugins doc
- adds github workflow that runs every time `backstage.json`, `packages/app/package.json` or `packages/backend/package.json` is updated. Workflow opens PR with updated versions.md file

- Fixes https://issues.redhat.com/browse/RHIDP-7158

## PR acceptance criteria


nickboldt:
Two thoughts (but otherwise +1):
* could this PR include the updated `./docs/dynamic-plugins/versions.md` as well?
* rather than excluding the old 0.x and 1.0 branches, why don't we just delete them entirely? they're long since EOL.

kadel:
@nickboldt 

> Two thoughts (but otherwise +1):
> 
> * could this PR include the updated `./docs/dynamic-plugins/versions.md` as well?

good point, will do that. Not sure why I did not do that at from start.

> * rather than excluding the old 0.x and 1.0 branches, why don't we just delete them entirely? they're long since EOL.

I think that it still makes sense to keep the old branches, there is really no harm in it, they might be useful for some historical references 

> 
> Also copilot complained:
> 
> ```
> The use of 'console.fail' is not a standard method in Node.js. Consider using 'console.error' followed by terminating the process (e.g., process.exit(1)) to handle the error.
> console.fail(`Unable to find create-app version in ${createAppVersion}`)
> ```

His has bee already fixed a while back

docs/dynamic-plugins/versions.md
```diff
@@ -1,132 +1,324 @@
-# Version Compatibility Matrix
 
-## RHDH 1.4 (pre-release, versions can change for final release)
+## RHDH next (pre-release, versions can change for final release)
```
nickboldt, line 2: I trust in the `release-1.y` branches we'll see a real version here, not `next` ? Also given we have https://github.com/redhat-developer/rhdh/blob/main/package.json#L3 could (should?) this instead be derived from that file?

kadel: > I trust in the release-1.y branches we'll see a real version here, not next ?

I intentionally did not want to do that to make it clear that this is not released yet. That is why I choose to show next instead of version.

> Also given we have https://github.com/redhat-developer/rhdh/blob/main/package.json#L3 could we (should we?) use the versions derived from that file?

It has a full version (including z stream). I did not want to go to that level. But I can start parsing this instead of branch name if that is for some reason better.

docs/dynamic-plugins/versions.md
```diff
@@ -1,132 +1,324 @@
...
-## RHDH 1.2
 
-Based on Backstage 1.26.5
-@backstage/create-app 0.5.14
+If you want to check versions of other packages, you can check the
+[`package.json`](https://github.com/redhat-developer/rhdh/blob/release-1.3/packages/backend/package.json) in the
+[`backend`](https://github.com/redhat-developer/rhdh/tree/release-1.3/packages/backend) package
+in the `release-1.3` branch of the [RHDH repository](https://github.com/redhat-developer/rhdh/tree/release-1.3).
+
+## RHDH 1.1
```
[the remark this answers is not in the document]
kadel: Unless we have some source where I can pull programmatically  which versions are EOL than I would prefer to list them all here. I don't want to manually maintain the list of active and EOL releases here.

nickboldt (approved):
LGTM - nice to see the old 1.0-1.2 stuff has been purged.

PR #2816, opened 2025-04-28
fix(install-script): improve Keycloak installation script

AndrienkoAleksandr:
## Description

- Fixes #?

## PR acceptance criteria


AndrienkoAleksandr:
@albarbaro , what do you think does this pr make sense?

albarbaro:
It depends on where this script is used. Surely, deploying using the operator would be the preferred way. Is this script used as part of some automation?

AndrienkoAleksandr:
> It depends on where this script is used. Surely, deploying using the operator would be the preferred way. Is this script used as part of some automation?

No, I don't think so. I think it was created for dev purposes, @Zaperex Am I right?

Zaperex:
Ah yes, it was made to quickly setup a keycloak instance for testing the keycloak plugins.

PR #2827, opened 2025-04-30
chore(e2e): rhidp 6618 e2e auditor log

gustavolira:
## Description

- Fixes #?

## PR acceptance criteria


e2e-tests/playwright/e2e/audit-log/rbac-test-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    status,
+    "permission", // plugin name (RBAC)
+    "medium", // expected severity
+    filterWords,
+    process.env.NAME_SPACE_RBAC,
+  );
+}
+
+/* ───────────────────────── BUILD RBAC REST API HELPER ──────────────────── */
+export async function buildRbacApi(page: Page): Promise<RhdhRbacApi> {
```
dzemanov, line 118: I would suggest do not extract these 2 lines used once to separate file or maybe better, incorporate into `rbac-api.ts`.

e2e-tests/playwright/e2e/audit-log/scaffold.spec.ts
```diff
@@ -4,76 +4,42 @@ import { UIhelper } from "../../utils/ui-helper";
...
-
-    await LogUtils.validateLogEvent(
-      "ScaffolderTaskListFetch",
-      "user:development/guest requested for the list of scaffolder tasks",
-      "GET",
-      "/api/scaffolder/v2/tasks?createdBy=user%3Adevelopment%2Fguest",
-      baseURL!,
-      "scaffolder",
-    );
+    const isComponentIsAlreadyRegistered =
```
dzemanov, line 28:
```suggestion
    const isComponentAlreadyRegistered =
```

e2e-tests/playwright/e2e/audit-log/scaffold.spec.ts
```diff
@@ -4,76 +4,42 @@ import { UIhelper } from "../../utils/ui-helper";
...
     await common.loginAsGuest();
-    await uiHelper.clickLink({ ariaLabel: "Self-service" });
+    await page.goto("/create");
   });
 
-  test("Should fetch logs for ScaffolderParameterSchemaFetch event and validate log structure and values", async ({
-    baseURL,
-    page,
-  }) => {
+  test("Should fetch logs for 'entity-mutate or location-mutate' event and validate log structure and values", async () => {
```
dzemanov, line 26: I would suggest do not merge these events and test separately, since this way we will validate either one or second event. We should validate both. Can we introduce this spec test case specific component to make sure all cases will always happen?

e2e-tests/playwright/e2e/audit-log/auditor-rbac.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        { method: "GET", url: s.url },
+        s.meta,
+      );
+    });
+  }
+
+  /* --------------------------------------------------------------------- */
+  /*  ROLE WRITE                                                           */
+  /* --------------------------------------------------------------------- */
+  (["create", "update", "delete"] as const).forEach((action) => {
```
dzemanov, line 68: It is great that you merged these similar test cases. I would however suggest avoiding if `testCase1`, if `testCase2` in the test execution as it actually reduces clarity. Each test should be isolated and stand on its own. I would suggest adding similar structure as was done later in this file, with `call` and `url` over which you iterate.

e2e-tests/playwright/e2e/audit-log/auditor-rbac.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        { method: "GET", url: s.url },
+        s.meta,
+      );
+    });
+  }
+
+  /* --------------------------------------------------------------------- */
+  /*  POLICY WRITE                                                         */
+  /* --------------------------------------------------------------------- */
+  (["create", "update", "delete"] as const).forEach((action) => {
```
dzemanov, line 141: The same as above. Please avoid if `testCase1`, if `testCase2` inside the test execution.

e2e-tests/playwright/e2e/audit-log/rbac-test-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  },
+  condition: {
+    collection: "/api/permission/roles/conditions",
+    item: (id: number | string) => `/api/permission/roles/conditions/${id}`,
+  },
+};
+
+/* ───────────────────────────── ERROR-MESSAGE HELPER ────────────────────── */
+// Backend verbs differ from our logical action names:
+export const BACKEND_VERB: Record<
```
dzemanov, line 33: A nit pick: I would avoid having this global and move it directly to `buildNotAllowedError` as it is used only there and is for figuring out what verb to use in the error message.

e2e-tests/playwright/e2e/audit-log/rbac-test-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+> = {
+  create: "add",
+  update: "edit",
+  delete: "delete",
+};
+
+/**
+ * Build the expected NotAllowedError message exactly as the backend returns.
+ * For role operations the backend omits the entityRef after "role:".
+ * For policy operations ele inclui "policy role:<entityRef>".
```
dzemanov, line 45: Please translate to english `ele inclui`

dzemanov (changes requested):
Thank you for refactoring the repeated test cases and fixing skip of scaffolder audit logs, looks great. I have left a couple of suggestions.

e2e-tests/playwright/e2e/audit-log/scaffold.spec.ts
gustavolira: @dzemanov  The issue is that if the test is being run for the first time, it will be an insert operation, resulting in the creation of an entity-mutate. However, if we run the test again, it will become a location-mutate. I've attempted to implement it the way you suggested, but it's practically impossible to guarantee consistent behavior. If there were an option to delete the template, I could delete it before running the test, which would allow me to test each scenario individually

e2e-tests/playwright/e2e/audit-log/scaffold.spec.ts
```diff
@@ -4,76 +4,44 @@ import { UIhelper } from "../../utils/ui-helper";
 import { LogUtils } from "./log-utils";
 import { CatalogImport } from "../../support/pages/catalog-import";
 
-test.describe.skip("Audit Log check for Catalog Plugin", () => {
+test.describe("Audit Log check for Catalog Plugin", () => {
```
dzemanov, line 7:
```suggestion
test.describe("Audit Log check for Scaffolder Plugin", () => {
```

e2e-tests/playwright/e2e/audit-log/scaffold.spec.ts
```diff
@@ -4,76 +4,44 @@ import { UIhelper } from "../../utils/ui-helper";
...
     await common.loginAsGuest();
-    await uiHelper.clickLink({ ariaLabel: "Self-service" });
+    await page.goto("/create");
   });
 
-  test("Should fetch logs for ScaffolderParameterSchemaFetch event and validate log structure and values", async ({
-    baseURL,
-    page,
-  }) => {
+  test("Should fetch logs for entity-mutate event and validate log structure and values", async () => {
```
dzemanov, line 26: I wonder if this shouldn't be in Catalog audit spec? Also, if you have problems with reliably checking `entity-mutate`, you can unregister the component after the test runs.

dzemanov (changes requested):
Thank you for the changes. I left 2 more questions and then I think we are good to go from my side. For the scaffolder tests, should we maybe update them to the new auditor version and just skip them till we can use `severityLogLevelMappings`, instead of deleting them? Or there is not much value in testing fetch events?

e2e-tests/playwright/e2e/audit-log/auditor-catalog.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { test } from "@playwright/test";
+import { Common } from "../../utils/common";
+import { UIhelper } from "../../utils/ui-helper";
+import { LogUtils } from "./log-utils";
+import { CatalogImport } from "../../support/pages/catalog-import";
+import { Log } from "./logs";
+
+test.describe("Audit Log check for Scaffold Plugin", () => {
```
dzemanov, line 8:
```suggestion
test.describe("Audit Log check for Catalog Plugin", () => {
```

e2e-tests/playwright/e2e/audit-log/auditor-catalog.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    const logLine = await LogUtils.getPodLogsWithRetry(
+      [expectedEvent.eventId, expectedEvent.method, expectedEvent.url],
+      process.env.NAME_SPACE || "showcase-ci-nightly",
+    );
+
+    const actualLog = JSON.parse(logLine);
+    LogUtils.validateLog(actualLog, expectedLog);
+  });
```
dzemanov, line 62: Can you use here already existing method `LogUtils.validateLogEvent`?

e2e-tests/playwright/e2e/audit-log/rbac-test-utils.ts
dzemanov: @gustavolira what do you think about this? Does it work for you to move this to `rbac-api.ts`, since it is used for example also [here](https://github.com/redhat-developer/rhdh/blob/main/e2e-tests/playwright/e2e/plugins/rbac/rbac.spec.ts#L470). Or do not extract at all?

dzemanov (changes requested):

gustavolira:
> Thank you for the changes. I left 2 more questions and then I think we are good to go from my side. For the scaffolder tests, should we maybe update them to the new auditor version and just skip them till we can use `severityLogLevelMappings`, instead of deleting them? Or there is not much value in testing fetch events?

@dzemanov  In practice these logs add very little audit value because they’re tied to a read-only action. During a typical scaffold run the backend issues dozens of entity-fetch calls, so enabling them just floods the log with low-severity noise without providing extra security or compliance insight.

For that reason I’d rather drop the test instead of keeping a skipped placeholder – we’re already validating all write-path events

dzemanov (approved):
Looks great! Thank you for all the changes.

.ibm/pipelines/utils.sh
```diff
@@ -1098,3 +1098,25 @@ detect_ocp_and_set_env_var() {
...
+# Helper function for case conversion
+to_lowercase() {
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    # macOS - using tr
+    echo "$1" | tr '[:upper:]' '[:lower:]'
+  else
+    # Linux - using bash parameter expansion
+    echo "${1,,}"
+  fi
+}
```
zdrapela, line 1122: For the MacOS (zsh) support, I'd rather simplify the condition.

From: `if [[ -z "${IS_OPENSHIFT}" || "${IS_OPENSHIFT,,}" == "false" ]]; then`
To: `if [[ -z "${IS_OPENSHIFT}" || "${IS_OPENSHIFT}" == "false" ]]; then`

As you can see at [L1097](https://github.com/redhat-developer/rhdh/pull/2827/files#diff-5700e984b56368043ef9a0c1ab37e906444544c1fb8e47beb8189366fa0602c9L1097), the function that sets the `IS_OPENSHIFT` env. variable uses lower case, and it will not change throughout the test. It will not cause any issues.

PR #2839, opened 2025-05-05
fix(homepage): fixing correct title of DOM to fix Routes title in Top Visited and Recent Visited Card

its-mitesh-kumar:
## Description 

On the first load, the customized app.title from app-config.yaml (e.g., "Mitesh Kumar's Developer Hub") was not reflected in the document title immediately. This caused incorrect titles like "Red Hat Developer Hub" to be captured in the "Top Visited" and "Recently Visited" cards. The root cause was that the title was initially rendered from the static index.html file using:
```
<title><%= config.getOptionalString('app.title') ?? 'Backstage' %></title>
```
Since this is a build-time value and config.getOptionalString('app.title') is not available until runtime, the browser briefly showed the fallback title ('Red Hat Developer Hub'), and this was picked up prematurely by the visit tracking logic.

- Fixes :  https://issues.redhat.com/browse/RHIDP-7009

Before Fix : 

https://github.com/user-attachments/assets/194dbc65-8403-4813-8a2e-c661d5bb71b3

After Fix 

https://github.com/user-attachments/assets/2edabae8-8288-4e42-b0b4-e0894c0c61c4

### DIFF
--- a/packages/app/src/components/AppBase/AppBase.tsx
+++ b/packages/app/src/components/AppBase/AppBase.tsx
@@ -1,8 +1,10 @@
 import React, { useContext } from 'react';
+import { Helmet } from 'react-helmet';
 import { Route } from 'react-router-dom';
 
 import { FlatRoutes } from '@backstage/core-app-api';
 import { AlertDisplay, OAuthRequestDialog } from '@backstage/core-components';
+import { configApiRef, useApi } from '@backstage/core-plugin-api';
 import { ApiExplorerPage } from '@backstage/plugin-api-docs';
 import {
   CatalogEntityPage,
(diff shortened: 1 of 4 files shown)


christoph-jerolimov (changes requested):
Sorry I don't know a better solution yet, but this adds home page specific code here which doesn't look nice to me.

packages/app/src/components/AppBase/AppBase.tsx
```diff
@@ -31,6 +32,19 @@ import ConfigUpdater from '../Root/ConfigUpdater';
 import { SearchPage } from '../search/SearchPage';
 import { settingsPage } from '../UserSettings/SettingsPages';
 
+export const TitleSetter = () => {
+  const config = useApi(configApiRef);
+  const titleFromConfig = config.getOptionalString('app.title');
+
+  React.useEffect(() => {
+    if (titleFromConfig) {
+      document.title = `Welcome back! | ${titleFromConfig}`;
```
christoph-jerolimov, line 41: Welcome Back is a title just set on the home page, right? I don't think we should hard code that value in the AppBase.

christoph-jerolimov (changes requested):
Sorry I don't know a better solution yet, but this adds home page specific code here which doesn't look nice to me.

PR #2844, opened 2025-05-06
chore: update global-header wrapper version

debsmita1:
## Description

Updates the global-header wrapper version so that it displays the correct user name in the profile dropdown when the user logs in via OIDC auth

- Fixes https://issues.redhat.com/browse/RHIDP-7057

Screenshot:

Screenshot 2025-05-06 at 10 41 09 PM

## PR acceptance criteria

### DIFF
--- a/catalog-entities/marketplace/packages/red-hat-developer-hub-backstage-plugin-global-header.yaml
+++ b/catalog-entities/marketplace/packages/red-hat-developer-hub-backstage-plugin-global-header.yaml
@@ -18,7 +18,7 @@ metadata:
 spec:
   packageName: "@red-hat-developer-hub/backstage-plugin-global-header"
   dynamicArtifact: ./dynamic-plugins/dist/red-hat-developer-hub-backstage-plugin-global-header
-  version: 1.5.0
+  version: 1.6.0
   backstage:
     role: frontend-plugin
     supportedVersions: 1.36.1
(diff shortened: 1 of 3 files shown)


debsmita1:
> Hi @debsmita1 , is the screenshot right? The reporter wanted to display same name in the header as in the user entity card, which is "RHDH QE DN" in this case.

Yea, I initially thought it should match the User entity name in the "Backstage Identity" card
But later changed it to match the name in the "Profile" card

@christoph-jerolimov @ShiranHi  any opinions ?

ShiranHi:
> Yea, I initially thought it should match the User entity name in the "Backstage Identity" card But later changed it to match the name in the "Profile" card
> 
> @christoph-jerolimov @ShiranHi any opinions ?

I think it should use the more user-friendly display name (e.g. "Jane Doe" instead of "default/jdoe"). It seems the "Backstage Entity" card shows it this way ([example](https://demo.backstage.io/settings/general)).

But I wanted to note that:

1. We added the user name to the home page header, do we know which one we used there? It might be worth using the same one for consistency.

2. I think it would be helpful to ask this question in the ticket to gain a clearer understanding of the reporter's request. Based on the screenshot shared in the Jira ticket, it seems the better username display comes from the "Backstage Identity".

debsmita1:
@ciiay I spoke to the UX and we came to a conclusion that we follow the same approach as the dynamic-home-page(https://github.com/redhat-developer/rhdh-plugins/pull/686). I will create a separate ticket to address that
This fix needs to be backported all the way to 1.5

ciiay (approved):
/lgtm

Thank you 🤝 

image

debsmita1:
Manual cherrypick https://github.com/redhat-developer/rhdh/pull/2875

PR #2851, opened 2025-05-08
fix: update app-config.example.yaml

coreydaley:
## Description

Changing the `app.baseUrl` to get rid of the error:

```
app:start: ⚠️   Conflict between app baseUrl and backend baseUrl:
app:start: 
app:start:     app.baseUrl:     http://localhost:7007
app:start:     backend.baseUrl: http://localhost:7007
app:start: 
app:start:     Must have unique hostname and/or ports.
app:start: 
app:start:     This can be resolved by changing app.baseUrl and backend.baseUrl to point to their respective local development ports.
```

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/app-config.example.yaml
+++ b/app-config.example.yaml
@@ -1,6 +1,6 @@
 app:
   title: RHDH
-  baseUrl: http://localhost:7007
+  baseUrl: http://localhost:3000
 
 organization:
   name: My Org


coreydaley:
/assign @04kash 
It looks like you changed this line to use port 7007 in https://github.com/redhat-developer/rhdh/commit/9144c6cb4f494471141bd41a6cf072a6f39397ea and it throws the error listed in the description now. It looks like the app-config.yaml is correct still though https://github.com/redhat-developer/rhdh/blob/main/app-config.yaml

04kash:
I assume you're running `yarn dev` with both the frontend and backend on `http://localhost:7007`, which is causing the error above. As outlined in [these steps](https://github.com/redhat-developer/rhdh/blob/main/docs/index.md#running-locally-with-the-optional-plugins), this requires setting app.baseUrl to `http://localhost:3000` and adding a backend.cors.origin.

The app.baseUrl and backend.baseUrl were set to `http://localhost:7007` to be able to use the `yarn start` command which I had updated to `turbo run build --filter=app && NODE_OPTIONS=--no-node-snapshot turbo run start --filter=backend`. However, it looks like it was later modified to
`NODE_OPTIONS=--no-node-snapshot turbo run start --filter=backend` in [this commit](https://github.com/redhat-developer/rhdh/commit/829a0fb0fdf9061bb98ae043acb0e61cad66b8e1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R10) by @schultzp2020 (this no longer requires both app.baseUrl and backend.baseUrl to be the same), but the README wasn’t updated to reflect that.

We have two options (for the `yarn start` command):

1. Update the start script to `turbo run build --filter=app && turbo run start --filter=backend` as suggested by @davidfestal in [this comment](https://github.com/redhat-developer/rhdh/pull/1271#discussion_r1677687332). (This would require setting both the app.baseUrl and backend.baseUrl to be `http://localhost:7007` when using `yarn start`)

2. Keep the current start script (`NODE_OPTIONS=--no-node-snapshot turbo run start --filter=backend`). In this case, since only the backend is running, the frontend base URL doesn’t matter I believe—we can reset app.baseUrl to `http://localhost:3000`, and update the documentation [here](https://github.com/redhat-developer/rhdh/blob/main/docs/index.md#running-locally-with-the-optional-plugins) accordingly. Would also make sense to rename `start` to `start-backend` IMO so that its clear that this script is to only start the backend.

PR #2915, opened 2025-05-26
chore: flpath-2318 - add orchestrator plugin deployment to helm deployment

y-first:
## Description

Adds orchestrator plugin to Helm chart deployment. Also updates the Helm chart version to a version that contains the orchestrator (will be changed after relevant [PR)](https://github.com/redhat-developer/rhdh/pull/2828/files) is merged.

- Fixes [FLPATH-2318](https://issues.redhat.com//browse/FLPATH-2318)

## PR acceptance criteria

### DIFF
--- a/.ibm/pipelines/utils.sh
+++ b/.ibm/pipelines/utils.sh
@@ -846,6 +846,18 @@ cluster_setup_k8s_helm() {
   # install_crunchy_postgres_k8s_operator # Works with K8s but disabled in values file
 }
 
+install_orchestrator_infra_chart() {
+  ORCH_INFRA_NS="orchestrator-infra"
+  configure_namespace ${ORCH_INFRA_NS}
+
+  echo "Deploying orchestrator-infra chart"
+  cd "${DIR}"
+  helm upgrade -i orch-infra -n "${ORCH_INFRA_NS}" \
+    "oci://quay.io/rhdh/orchestrator-infra-chart" --version "${CHART_VERSION}" \
+    --set serverlessLogicOperator.subscription.spec.installPlanApproval=Automatic \
+    --set serverlessOperator.subscription.spec.installPlanApproval=Automatic
+}
+
 # Helper function to get common helm set parameters
 get_image_helm_set_params() {
   local params=""
@@ -866,7 +878,7 @@ perform_helm_install() {
   local release_name=$1
   local namespace=$2
   local value_file=$3
-
+  
   helm upgrade -i "${release_name}" -n "${namespace}" \
     "${HELM_CHART_URL}" --version "${CHART_VERSION}" \
     -f "${DIR}/value_files/${value_file}" \
@@ -878,6 +890,10 @@ base_deployment() {
   configure_namespace ${NAME_SPACE}
 
   deploy_redis_cache "${NAME_SPACE}"
+
+  install_orchestrator_infra_chart
+
+  cd "${DIR}"
   local rhdh_base_url="https://${RELEASE_NAME}-developer-hub-${NAME_SPACE}.${K8S_CLUSTER_ROUTER_BASE}"
   apply_yaml_files "${DIR}" "${NAME_SPACE}" "${rhdh_base_url}"
   echo "Deploying image from repository: ${QUAY_REPO}, TAG_NAME: ${TAG_NAME}, in NAME_SPACE: ${NAME_SPACE}"
@@ -989,7 +1005,8 @@ initiate_sanity_plugin_checks_deployment() {
     "${HELM_CHART_URL}" --version "${CHART_VERSION}" \
     -f "/tmp/${HELM_CHART_SANITY_PLUGINS_MERGED_VALUE_FILE_NAME}" \
     --set global.clusterRouterBase="${K8S_CLUSTER_ROUTER_BASE}" \
-    $(get_image_helm_set_params)
+    $(get_image_helm_set_params)  \
+    --set orchestrator.enabled=true
 }
 
 check_and_test() {
--- a/.ibm/pipelines/value_files/values_showcase.yaml
+++ b/.ibm/pipelines/value_files/values_showcase.yaml
@@ -315,3 +315,24 @@ upstream:
       - name: http-metrics
         port: 9464
         targetPort: 9464
+
+orchestrator:
+  enabled: true
+  serverlessLogicOperator:
+    enabled: true
+  serverlessOperator:
+    enabled: true
+  sonataflowPlatform:
+    monitoring:
+      enabled: true
+    eventing:
+      broker:
+        name: ""
+        namespace: ""
+    resources:
+      requests:
+        memory: "64Mi"
+        cpu: "250m"
+      limits:
+        memory: "1Gi"
+        cpu: "500m"


.ibm/pipelines/utils.sh
```diff
@@ -918,6 +957,7 @@ initiate_upgrade_deployments() {
     --set global.clusterRouterBase="${K8S_CLUSTER_ROUTER_BASE}" \
     --set upstream.backstage.image.repository="${QUAY_REPO}" \
     --set upstream.backstage.image.tag="${TAG_NAME}" \
+    --set orchestrator.enabled=true \
```
subhashkhileri: @y-first is there a upgrade test for orchestrator? if no can we remove this?

.ibm/pipelines/utils.sh
```diff
@@ -880,6 +916,8 @@ initiate_upgrade_base_deployments() {
 
   deploy_redis_cache "${NAME_SPACE}"
 
+  install_orchestrator_infra_chart
```
subhashkhileri: https://github.com/redhat-developer/rhdh/pull/2915/files#r2142522353

.ibm/pipelines/utils.sh
```diff
@@ -840,11 +840,46 @@ cluster_setup_k8s_helm() {
   # install_crunchy_postgres_k8s_operator # Works with K8s but disabled in values file
 }
 
+# Helper function to install orchestrator infra chart. Only applicable for chart versions 4.0.0 + 
+install_orchestrator_infra_chart() {
```
subhashkhileri, line 849: @y-first does it require helm install only for orch-infra? my understanding was it will be part of rhdh chart.

y-first: It is part of RHDH chart, but the orchestrator-infra chart needs to be installed before RHDH chart is installed. It contains dependencies for orchestrator (Serverless Logic and Serverless Operators)

.ibm/pipelines/utils.sh
y-first: there is no upgrade test currently for orchestrator.

.ibm/pipelines/utils.sh
```diff
@@ -865,6 +897,7 @@ perform_helm_install() {
   local release_name=$1
   local namespace=$2
   local value_file=$3
+  local additional_values=$4
```
subhashkhileri: additional_values isn't being used please remove.

y-first: Added below

.ibm/pipelines/utils.sh
```diff
@@ -846,6 +846,38 @@ cluster_setup_k8s_helm() {
   # install_crunchy_postgres_k8s_operator # Works with K8s but disabled in values file
 }
 
+install_orchestrator_infra_chart() {
+  ORCH_INFRA_NS="orchestrator-infra"
+  configure_namespace ${ORCH_INFRA_NS}
+
+  helm repo add redhat-developer https://redhat-developer.github.io/rhdh-chart
```
rm3l: I think the RHDH Chart is installed via its OCI artifact (as the goal here is to test the chart that is packaged via Konflux) - see https://github.com/redhat-developer/rhdh/blob/main/.ibm/pipelines/env_variables.sh#L16.
So I guess it would make sense to also use the orchestrator-infra chart OCI artifact too: https://quay.io/repository/rhdh/orchestrator-infra-chart?tab=tags

.ibm/pipelines/utils.sh
```diff
@@ -846,6 +846,38 @@ cluster_setup_k8s_helm() {
...
+
+  echo "Deploying orchestrator-infra chart"
+  cd "${DIR}"
+  helm upgrade -i orch-infra -n "${ORCH_INFRA_NS}" \
+    "redhat-developer/redhat-developer-hub-orchestrator-infra"
```
rm3l: Maybe you can just set the InstallModes as `Automatic` by appending this to your `helm upgrade -i` command:

```
--set serverlessLogicOperator.subscription.spec.installPlanApproval=Automatic \
--set serverlessOperator.subscription.spec.installPlanApproval=Automatic
```

This way, you won't need to search for the InstallPlan and you won't need to manually approve it with `oc patch installplan`.
Unless this is intentional of course, to mimic what the cluster admin would need to do..

rm3l (commented):
A few comments below..

rm3l:
@y-first Is this work related to this RHIDP Feature: https://issues.redhat.com/browse/RHIDP-6233 ?

y-first:
@rm3l Yes, it is. the equivalent epic on the Orchestrator QE side is [FLPATH-2280](https://issues.redhat.com/browse/FLPATH-2280).

.ibm/pipelines/utils.sh
y-first: This is meant to mimic what the cluster admin would do, because using the Manual mode is to make sure that the operator doesn't upgrade Openshift Serverless Logic and Openshift Serverless versions without permission. In this case because the cluster is ephemeral, there is no issue.

.ibm/pipelines/utils.sh
y-first: @rm3l I need to specify the chart version, do you know an easy way to pull the most recent version?

subhashkhileri: it will always be the new version, AFAIK

subhashkhileri: you can use in similar way how rhdh is deployed with latest chart

y-first: the rhdh chart gets the latest version by curling from here:
`curl -sS https://raw.githubusercontent.com/rhdh-bot/openshift-helm-charts/refs/heads/rhdh-1-rhel-9/installation/README.md | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'`

y-first: Is it safe to assume that the RHDH chart OCI artifact and the orchestrator-infra chart OCI artifact will always share the same version number?

.ibm/pipelines/utils.sh
y-first: Added.

.ibm/pipelines/utils.sh
rm3l: > the rhdh chart gets the latest version by curling from here:
> `curl -sS https://raw.githubusercontent.com/rhdh-bot/openshift-helm-charts/refs/heads/rhdh-1-rhel-9/installation/README.md | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'`

🤔 The rhdh-bot/openshift-helm-charts repo is deprecated, so I guess this command should be updated accordingly.

> Is it safe to assume that the RHDH chart OCI artifact and the orchestrator-infra chart OCI artifact will always share the same version number?

I think so at this time; they are published via the same pipeline run using the same chart version: https://gitlab.cee.redhat.com/rhidp/rhdh/-/blob/rhdh-1-rhel-9/build/helm/publish_task.yaml

y-first:
> /approve
> 
> Approved after looking at the [job logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-developer_rhdh/2915/pull-ci-redhat-developer-rhdh-main-e2e-tests/1942589383408357376/artifacts/e2e-tests/redhat-developer-rhdh/build-log.txt), but I'll leave it to @subhashkhileri and the QE folks to further validate this.

@subhashkhileri Do you have any additional comments?

PR #2937, opened 2025-06-03
Add oidcLdapUuidMatchingAnnotation sign in resolver

JessicaJHee:
## Description

The existing sub claim OIDC resolver is not compatible with the LDAP catalog plugin because the sub claim is an OIDC specific field and does not exist in LDAP. An alternative ID that is also immutable and secure that we can use in this case is UUID. 

The UUID is meant to be unique and cannot be changed.

For OpenLDAP (see [docs](https://datatracker.ietf.org/doc/html/rfc4530#section-2.4)):
"An entry's UUID is immutable." ... "In particular, servers MUST ensure that each generated UUID is unique in space and time." 

Same thing with Microsoft AD (see [docs](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers#sids-and-globally-unique-identifiers)):
"searching for any object by ObjectGUID might be the most reliable way of finding the object you want to locate. The values of other object properties can change, but the ObjectGUID property never changes. When an object is assigned a GUID, it keeps that value for life." 

- Fixes https://issues.redhat.com/browse/[RHIDP-7543](https://issues.redhat.com/browse/RHIDP-7543)

## PR acceptance criteria

Set up LDAP instance with OIDC auth in Keycloak (via user federation) 

You would need to create a new client scope and mapper in Keycloak to make this work. 

1. `Client scopes` > `Create client scope`: name it `ldap_uuid`
2. In the `ldap_uuid` scope, go to `Mapper` tab:
3. Click `Add mapper`  > `Configure a new mapper`  > `User Attribute`
4. Fill out like this: 
 image
    
5. In: `Clients` > `Client details` > `Client scopes` > click `Add client scope` and add the `ldap_uuid` scope we just created 
    
image    

* Make sure this scope is included by default
* You should now be able to see the `uuid` field under the `Evaluate` tab


JessicaJHee:
The existing e2e tests have passed when run locally - tests for this resolver can be added along with the LDAP e2e tests after, as they are not yet implemented.

PR #2977, opened 2025-06-12
feat: add config to allow for auxiliary auth providers without identity resolution

JessicaJHee:
## Problem/ Background
For instance, a primary IdP like Keycloak is used for sign-in and user provisioning, and GitHub auth is needed to authenticate requests for GitHub-based plugins. In this case, we should allow both auth providers to be configured. If the GitHub catalog provider is not configured, users signing in via GitHub will encounter the error with user not found in catalog.

However, if we enable the GitHub org catalog provider on top of the existing one (e.g. Keycloak), there is a risk of user duplication or conflicts due to attempts to ingest the same user from multiple sources. This is not good practice.

## Implementation 
This PR introduces a `disableIdentityResolution` config.
- False by default
- When set to true, a backstage identity will not be issued when the user signs in with an auxiliary sign-in provider (see [upstream code](https://github.com/backstage/backstage/blob/df766a420e2c84fe6e4debcd00d7f5fb3479eddd/plugins/auth-node/src/oauth/createOAuthRouteHandlers.ts#L350-L357))
- If you try to sign in with a provider with disableIdentityResolution enabled, you will get:
`Login failed; caused by Error: The GitHub provider is not configured to support sign-in`
- Permissions are based on the user entity, so this additional auth provider will not affect that.

- Fixes [RHIDP-7798](https://issues.redhat.com/browse/RHIDP-7798)

## PR acceptance criteria

1. Set up user sign in and user/group provisioning with a primary provider (like Keycloak + OIDC)
2. Set up GtiHub auth and set: `auth.providers.github.development.disableIdentityResolution: true`
3. Verify that you can log in with GitHub without a matching user in the catalog and that no Backstage identity is issued from signing in with GitHub


JessicaJHee:
~~Waiting for [this PR](https://github.com/redhat-developer/rhdh/pull/2993) to be merged to pull in update e2e auth tests~~

PatAKnight (approved):
/LGTM

PR #3109, opened 2025-07-11
Add RH registry to default .npmrc in image to support orchestrator pl…

Fortune-Ndlovu:
…ugin installs

## Description

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -252,6 +252,9 @@ RUN ln -s /opt/app-root/src/packages/app/dist/index.html /opt/app-root/src/packa
 # Fix for https://issues.redhat.com/browse/RHIDP-728
 RUN mkdir -p /opt/app-root/src/.npm; chown -R 1001:1001 /opt/app-root/src/.npm
 
+# Ensure orchestrator plugins can be resolved from the Red Hat NPM registry without needing a .npmrc Secret
+RUN npm config set @redhat:registry https://npm.registry.redhat.com
+
 # The fix-permissions script is important when operating in environments that dynamically use a random UID at runtime, such as OpenShift.
 # The upstream backstage image does not account for this and it causes the container to fail at runtime.
 # suppress warnings about dereferencing symlinks


docker/Dockerfile
rm3l, line 1: The downstream Dockerfile should be updated as well, IMO: https://github.com/redhat-developer/rhdh/blob/main/.rhdh/docker/Dockerfile

docker/Dockerfile
```diff
@@ -252,6 +252,9 @@ RUN ln -s /opt/app-root/src/packages/app/dist/index.html /opt/app-root/src/packa
 # Fix for https://issues.redhat.com/browse/RHIDP-728
 RUN mkdir -p /opt/app-root/src/.npm; chown -R 1001:1001 /opt/app-root/src/.npm
 
+# Ensure orchestrator plugins can be resolved from the Red Hat NPM registry without needing a .npmrc Secret
+RUN npm config set @redhat:registry https://npm.registry.redhat.com
```
rm3l: There is a layer where the npm config is edited: https://github.com/redhat-developer/rhdh/blob/c2d79cdb6d99791dd2b4bac20764333929e40ac0/docker/Dockerfile#L194-L199
I'd suggest moving this there.

rm3l: @nickboldt @kim-tsao Any concerns if we do this (in the downstream RHDH image too)? We think it would make sense to be added by default to the RHDH image, to help us fix https://issues.redhat.com/browse/RHDHBUGS-1464

docker/Dockerfile
Fortune-Ndlovu: I agree, and I have updated the downstream Dockerfile.

docker/Dockerfile
Fortune-Ndlovu: Thanks @rm3l, makes sense to centralize all npm config in one place.

kim-tsao: @rm3l yes, I think it'll be good to keep it consistent

.rhdh/docker/Dockerfile
```diff
@@ -215,6 +215,7 @@ RUN \
   find . -type f -name .npmrc -exec rm -Rf {} \; && \
   # reset npm config to the default registry and absolute path to .pem file
   npm config set registry=https://registry.npmjs.org/ && \
+  npm config set @redhat:registry https://npm.registry.redhat.com && \
```
rm3l: Performed some testing locally against the image from this PR, and, as I suspected, this unfortunately does not solve the issue. If a user adds their custom npmrc secret, it overrides the RH registry added here. So the Orchestrator plugins won't be fetched out of the box:

```
[...]

======= Installing dynamic plugin @redhat/backstage-plugin-orchestrator@1.6.0
        ==> Grabbing package archive through `npm pack`
Traceback (most recent call last):
  File "/opt/app-root/src/install-dynamic-plugins.py", line 523, in <module>
    main()
  File "/opt/app-root/src/install-dynamic-plugins.py", line 439, in main
    raise InstallException(f'Error while installing plugin { package } with \'npm pack\' : ' + completed.stderr.decode('utf-8'))
InstallException: Error while installing plugin @redhat/backstage-plugin-orchestrator@1.6.0 with 'npm pack' : npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@redhat%2fbackstage-plugin-orchestrator - Not found
npm error 404
npm error 404  '@redhat/backstage-plugin-orchestrator@1.6.0' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
npm error Log files were not written due to an error writing to the directory: /opt/app-root/src/.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal

======= Removed lock file: /dynamic-plugins-root/install-dynamic-plugins.lock
```

So we need a different approach here, I guess.

/hold

Fortune-Ndlovu: I've just tested the updated image:
`quay.io/rhdh-community/rhdh:pr-3109`
and using the helm upgrade with:
```bash
--set upstream.backstage.image.repository=quay.io/rhdh-community/rhdh \
--set upstream.backstage.image.tag=pr-3109
```
and confirmed that the pod still fails during the `install-dynamic-plugins` init container with the same error:
```bash
======= Installing dynamic plugin @redhat/backstage-plugin-orchestrator-backend-dynamic@1.6.0
        ==> Grabbing package archive through `npm pack`
Traceback (most recent call last):
  File "/opt/app-root/src/install-dynamic-plugins.py", line 523, in <module>
    main()
  File "/opt/app-root/src/install-dynamic-plugins.py", line 439, in main
    raise InstallException(f'Error while installing plugin { package } with \'npm pack\' : ' + completed.stderr.decode('utf-8'))
InstallException: Error while installing plugin @redhat/backstage-plugin-orchestrator-backend-dynamic@1.6.0 with 'npm pack' : npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@redhat%2fbackstage-plugin-orchestrator-backend-dynamic - Not found
npm error 404
npm error 404  '@redhat/backstage-plugin-orchestrator-backend-dynamic@1.6.0' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
npm error Log files were not written due to an error writing to the directory: /opt/app-root/src/.npm/_logs
npm error You can rerun the command with `--loglevel=verbose` to see the logs in your terminal

======= Removed lock file: /dynamic-plugins-root/install-dynamic-plugins.lock
```
So even with the corrected syntax:
`npm config set @redhat:registry=https://npm.registry.redhat.com`
...it still does not work when a custom `.npmrc` secret is mounted, because at runtime the Helm chart mounts a custom `.npmrc` secret (via `NPM_CONFIG_USERCONFIG=/opt/app-root/src/.npmrc.dynamic-plugins`). so when [`NPM_CONFIG_USERCONFIG`](https://github.com/redhat-developer/rhdh-chart/blob/main/charts/backstage/values.yaml#L210) is set, `npm` only uses that file, ignoring any configuration stored globally in the image. So, since the mounted `.npmrc.dynamic-plugins` does not contain the `@redhat:registry` entry, `npm` falls back to the default public registry `https://registry.npmjs.org/` for the orchestrator plugins, hence our 404. so the failure is due to the runtime-mounted `.npmrc` fully replacing the preconfigured registry settings, not mgerging them, we need a merge strategy at runtime.

Fortune-Ndlovu:
closing this pr, because this solution was not feasible and the best solution was to directly update the helm chart to directly install from the tarball URL, bypassing the default npm registry and avoiding .npmrc overrides.

PR #3120, opened 2025-07-14
chore(init): skip including dynamic packages if file not exists

dzemanov:
## Description

Skip including dynamic packages if file does not exist, as it will be created the first time by extensions with this config:
```
plugins: []
```

- Fixes https://issues.redhat.com/browse/RHIDP-8103

## PR acceptance criteria


docker/install-dynamic-plugins.py
```diff
@@ -294,7 +294,8 @@ def main():
         print('\n======= Including dynamic plugins from', include, flush=True)
 
         if not os.path.isfile(include):
-            raise InstallException(f"File {include} does not exist")
+            print(f"File {include} does not exist, skipping including dynamic packages from {include}", flush=True)
```
kadel:
```suggestion
            print(f"WARNING: File {include} does not exist, skipping including dynamic packages from {include}", flush=True)
```

kadel (changes requested):
When someone makes a typo in a filename, it is easy to miss that message. I would prefix it with `WARNING` to make it more visible. 

otherwise LGTM

kadel (approved):

PR #3146, opened 2025-07-17
fix(adoptionInsights): improve test stability and reduce flakiness in Adoption Insights tests

HusneShabbir:
---

Resolves:
https://issues.redhat.com/browse/RHDHBUGS-1879

---

## PR Description

This PR addresses the flakiness observed in the Playwright tests for the Adoption Insights plugin. The following key changes have been implemented to improve test stability and reliability:

### Core Changes:

* **Explicit API Call Waiting:**
    * Previously, the tests relied on `waitUntilApiCallSucceeds` in some places, which might not have covered all necessary API calls for panel data to be fully loaded and reflected.
    * Now, a new `waitForPanelApiCalls` helper has been introduced in `TestHelper`. This helper explicitly waits for *all* relevant Adoption Insights API endpoints (`active_users`, `total_users`, `top_templates`, `top_catalog_entities`, `top_plugins`, `top_techdocs`, `top_searches`) to be called, ensuring that the panels are populated with the latest data before assertions are made. This significantly reduces the chances of tests failing due to stale or incomplete data.
    * The `waitUntilApiCallSucceeds` is now also called within `waitForPanelApiCalls` to ensure the final data flush after all individual type calls.

* **Improved Panel Data Population (`populateMissingPanelData`):**
    * The `populateMissingPanelData` function has been enhanced to be more robust. It now includes a check to register a sample template if no templates are initially visible, ensuring that the "Top 3 templates" panel always has data to display. This prevents scenarios where tests might fail because there's no data to assert against.
    * Navigation for techdocs population now explicitly uses `page.goto('/docs')` for more reliable navigation to the TechDocs homepage.

* **Click and Verify Text Enhancements (`clickAndVerifyText`):**
    * A `newpage.waitForTimeout(5000)` has been added after verifying text in a new tab. This provides a buffer to account for potential data flush intervals, especially when interactions trigger updates that are not immediately reflected.
    * The new tab (`newpage`) is now explicitly closed with `await newpage.close()` after interactions, preventing potential resource leaks and ensuring a cleaner test environment for subsequent actions.

These modifications aim to make the Adoption Insights Playwright tests more resilient to timing issues and data loading variations, leading to a more consistent and reliable CI/CD pipeline.


e2e-tests/playwright/e2e/plugins/adoption-Insights/adoption-Insights.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { test, expect } from '@playwright/test';
 import { Common } from "../../../utils/common";
 import { UIhelper } from "../../../utils/ui-helper"; 
-import { TestHelper } from "../../../support/pages/adoption-Insights"; 
+import { TestHelper } from "../../../support/pages/adoption-insights";
```
zdrapela: The linter in the pre-commit hook also requires the files to be renamed:
- adoption-Insights/adoption-Insights.spec.ts -> adoption-insights/adoption-insights.spec.ts
- adoption-Insights.ts -> adoption-insights.ts

e2e-tests/playwright/e2e/plugins/adoption-Insights/adoption-Insights.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { test, expect } from '@playwright/test';
 import { Common } from "../../../utils/common";
 import { UIhelper } from "../../../utils/ui-helper"; 
-import { TestHelper } from "../../../support/pages/adoption-Insights"; 
+import { TestHelper } from "../../../support/pages/adoption-insights"; 
 
 // TODO: replace skip with serial https://issues.redhat.com/browse/RHDHBUGS-1879
 test.describe.skip("Test Adoption Insights", () => {
```
zdrapela: You have to re-enable the test to verify it in the CI
```suggestion
test.describe.serial("Test Adoption Insights", () => {
```

zdrapela (changes requested):
Thank you for the improvements. Please make sure the pre-commit hook (.husky/pre-commit) is kicking in, I can still see some errors. It should be by default, but you can trigger it manually, when you `cd e2e-tests` and then fix all the errors in you files that will be shown in `yarn lint:fix` and `yarn prettier:fix`

e2e-tests/playwright/support/pages/adoption-insights.ts
```diff
@@ -0,0 +1,183 @@
...
+  async clickAndVerifyText(
+    firstEntry: Locator,
+    expectedText: string,
+  ): Promise<void> {
+    const [newpage] = await Promise.all([
+      this.page.waitForEvent("popup"),
+      firstEntry.locator("a").click(),
+    ]);
+    // Wait for the expected API call to succeed
+    await this.waitUntilApiCallSucceeds(newpage);
```
karthikjeeyar, line 133: shouldn't this be added to above Promise.all array?

HusneShabbir: The reason `await this.waitUntilApiCallSucceeds(newpage)` isn’t included inside the Promise.all array is because it depends on the popup page (newpage) being fully initialized first.
We first wait for the popup to open (this.page.waitForEvent("popup")), then after that we start waiting for the API call on the new page. Including it inside Promise.all could cause it to run before newpage is available, which would cause issues.

karthikjeeyar: Got it. Thanks for clarifying.

zdrapela (approved):

PR #3147, opened 2025-07-17
chore(e2e): quick start e2e

teknaS47:
## Description

E2E tests for quick start plugin

- Fixes [RHIDP-7991](https://issues.redhat.com/browse/RHIDP-7991)

## PR acceptance criteria

### DIFF
--- a/e2e-tests/playwright/e2e/plugins/quick-start.spec.ts
+++ b/e2e-tests/playwright/e2e/plugins/quick-start.spec.ts
@@ -1,0 +1,46 @@
+import { expect, test } from "@playwright/test";
+import { Common } from "../../utils/common";
+import { UIhelper } from "../../utils/ui-helper";
+
+test.describe("Test Quick Start plugin", () => {
+  let uiHelper: UIhelper;
+  let common: Common;
+
+  test.beforeEach(async ({ page }) => {
+    common = new Common(page);
+    // await common.loginAsKeycloakUser();
+    await common.loginAsGuest();
+
+    uiHelper = new UIhelper(page);
+  });
+
+  test("Access Quick start from Global Header", async ({ page }) => {
+    await uiHelper.clickButtonByLabel("Help");
+    await uiHelper.clickByDataTestId("quickstart-button");
+    await uiHelper.clickButtonByLabel("Help");
+    await uiHelper.clickByDataTestId("quickstart-button");
+    await uiHelper.verifyText("Not started");
+    await uiHelper.verifyText("Let's get you started with Developer Hub");
+    await uiHelper.verifyText("We'll guide you through a few quick steps");
+    await uiHelper.clickButtonByText("Set up authentication");
+    await uiHelper.verifyButtonURL(
+      "Learn more",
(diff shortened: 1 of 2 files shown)


e2e-tests/playwright/e2e/plugins/quick-start.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    await common.loginAsGuest();
+
+    uiHelper = new UIhelper(page);
+  });
+
+  test("Access Quick start from Global Header", async ({ page }) => {
+    await uiHelper.clickButtonByLabel("Help");
+    await uiHelper.clickByDataTestId("quickstart-button");
+    await uiHelper.clickButtonByLabel("Help");
+    await uiHelper.clickByDataTestId("quickstart-button");
```
divyanshiGupta: @teknaS47 why are these steps repeated?

divyanshiGupta (changes requested):

e2e-tests/playwright/e2e/plugins/quick-start.spec.ts
teknaS47: To check if we can toggle the quick start panel with the button. As it starts with the quick start panel being visible, the first part hides it, and the second reopens it.

divyanshiGupta: Okay, but its not clear what is happening here. To make it more clear, lets first check if the panel is open then use `quickstart-button` to close it and then open it again.

divyanshiGupta:
@teknaS47 did you forgot to push the changes?

e2e-tests/playwright/e2e/plugins/quick-start.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  test.beforeEach(async ({ page }) => {
+    common = new Common(page);
+    // await common.loginAsKeycloakUser();
+    await common.loginAsGuest();
+
+    uiHelper = new UIhelper(page);
+  });
+
+  test("Access Quick start from Global Header", async ({ page }) => {
+    await uiHelper.verifyText("Not started");
```
divyanshiGupta: Instead of using `Not started` (changes when steps are completed to show the progress) to check if the drawer is open we should check for something that will remain constant like `Hide` button. Ideally we should add a data-testid on the quickstart component and then check it here but for now lets go with `Hide`.

divyanshiGupta: Also Quickstart is automatically opened only for the first time user opens RHDH or its not closed explicitly after opening it. So this test should be checking if the quickstart is open then use quickstart-button to close it and then open it otherwise if its not open use quickstart-button to open it and then procees with other tests. With this approach the test will not fail in case the quickstart is not open.

divyanshiGupta:
@teknaS47 can you check why the CI is red?

HusneShabbir:
Due to an update from OpenShift CI, all pre-submit jobs have been set to always run, even though they were marked as optional. This is causing all nightly jobs to trigger on every PR. subhashkhileri  is working on a fix.

divyanshiGupta (approved):
/approve

PR #3149, opened 2025-07-17
feat(ci): add workflow to check marketplace catalog entities

AndrienkoAleksandr:
## Description

Add workflow to check marketplace catalog entities. We have a lot of contributions with broken plugin/packages relations.

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/.github/workflows/scripts/check-marketplace-catalog-entities.py
+++ b/.github/workflows/scripts/check-marketplace-catalog-entities.py
@@ -1,0 +1,96 @@
+import os
+import sys
+import yaml
+
+def find_discrepancies():
+    packages_path = './catalog-entities/marketplace/packages'
+    plugins_path = './catalog-entities/marketplace/plugins'
+    warnings_found = False
+
+    plugin_names = set()
+    for filename in os.listdir(plugins_path):
+        if filename.endswith('.yaml'):
+            filepath = os.path.join(plugins_path, filename)
+            with open(filepath, 'r') as f:
+                try:
+                    # Load all documents from a single file
+                    all_data = yaml.safe_load_all(f)
+                    for data in all_data:
+                        if data and 'metadata' in data and 'name' in data['metadata']:
+                            plugin_names.add(data['metadata']['name'])
+                except yaml.YAMLError as e:
+                    print(f"Error reading {filepath}: {e}")
+                    warnings_found = True
+
+    for filename in os.listdir(packages_path):
+        if filename.endswith('.yaml'):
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/check-marketplace-catalog-entities.py
```diff
@@ -0,0 +1,96 @@
...
+            with open(os.path.join(packages_path, filename), 'r') as f:
+                try:
+                    data = yaml.safe_load(f)
+                    if data and 'spec' in data and 'partOf' in data['spec']:
+                        for plugin in data['spec']['partOf']:
+                            found = False
+                            for p_name in plugin_names:
+                                if plugin in p_name:
+                                    found = True
+                                    break
```
dzemanov, line 36: Did not check if this actually works, but I think it can be simplified to sth like this:
```suggestion
                        for plugin_ref in data['spec']['partOf']:
                            found = any(plugin_ref in plugin_name for plugin_name in plugin_names)
```

I think we want to do == instead of in. Because if we use in, you could have True, that kubernetes exists although we would e.g have only kubernetes-backend plugin. But those cases are rare.

The partOf relation is created like this:
```
          const pluginRef = parseEntityRef(pluginName, {
            defaultKind: MarketplaceKind.Plugin,
            defaultNamespace: packageEntity.metadata.namespace,
          });
```
So you could have also e.g  namespace-name/plugin-name, for which == doesn't work.

.github/workflows/scripts/check-marketplace-catalog-entities.py
```diff
@@ -0,0 +1,96 @@
...
+                            if not found:
+                                print(f"Warning: Package {filename}  contains unknown plugin with name '{plugin}'. Find correct plugin name defined in the yaml section \"metadata.name\".\n")
```
dzemanov, line 38:
```suggestion
                                print(f"Warning: Package {filename} contains unknown plugin with name '{plugin}'. Find correct plugin name defined in the yaml section \"metadata.name\".\n")
```

dzemanov (changes requested):
I think it would be also good to add check for if plugin doesn't contain packages and that package doesn't contain partOf. Not sure if we want to outright fail for that, but without it, relations are not set correctly and installation is not working properly.

dzemanov:
@AndrienkoAleksandr are you still working on this one? Can you create jira to track this? Or do we want to close this since the script will move to overlays and instead create the task to make it work with overlays? According to [RHIDP-8210](https://issues.redhat.com/browse/RHIDP-8210) the move is for 1.8 @nickboldt?

PR #3186, opened 2025-07-23
chore(release): add next-1.y tags from commits on the release-** branches (RHIDP-8256)

nickboldt:
### What does this PR do?

chore(release): add next-1.y tags from commits on the release-** branches ([RHIDP-8256](https://issues.redhat.com//browse/RHIDP-8256))

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### Screenshot/screencast of this PR
N/A

### What issues does this PR fix or reference?
N/A (or see commit message above for issue number)

### How to test this PR?
N/A

### PR Checklist

As the author of this Pull Request I made sure that:

- [x] Code produced is complete
- [ ] Code builds without errors
- [ ] Tests are covering the bugfix
- [ ] Relevant user documentation updated
- [ ] Relevant contributing documentation updated

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

### DIFF
--- a/.github/workflows/next-build-image.yaml
+++ b/.github/workflows/next-build-image.yaml
@@ -27,6 +27,8 @@ on:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
       - '[0-9]+.[0-9]+'
+    branches:
+      - 'release-**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/next-build-image.yaml
```diff
@@ -27,6 +27,8 @@ on:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
       - '[0-9]+.[0-9]+'
+    branches:
+      - 'release-**'
```
rm3l, line 31: Right now, the image from `main` is built daily (or on demand), not on every push; I guess because it takes almost an hour to build the images?
Would it make sense to do the same thing for the release branches?

nickboldt: meaning you want a daily(?) cron to build from all(?) the release-1.y branches? Or just the current release-1.7 branch, which we would somehow bump to 1.8 every quarter (manually or via generated PR) ?

I was thinking that since the stable branches don't get updated as often we could see if the load is reasonable to build on every commit. We could also set the build process to cancel any running jobs if a fresh commit comes in. 

seems like we can only run crons on the default branch https://github.com/orgs/community/discussions/26374 but we could maybe set up a matrix to run on BOTH main and release-1.y (just the 1 oe 2 latest ones) in parallel or series, if anything has changed since the last cron run

rm3l: > seems like we can only run crons on the default branch [github.com/orgs/community/discussions/26374](https://github.com/orgs/community/discussions/26374) but we could maybe set up a matrix to run on BOTH main and release-1.y (just the 1 oe 2 latest ones) in parallel or series

Yeah, I was thinking of a matrix indeed. We used to do something like this in the operator repo: https://github.com/redhat-developer/rhdh-operator/blob/750c45327dfa4574184bd1f13726c27eb293feaf/.github/workflows/nightly.yaml#L31-L33

But that's fine, I think, as the release branches are not updated as often as main. But yes, I think it would make sense to cancel in-progress builds, but it should already be the case, I guess: https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/next-build-image.yaml#L31C1-L33C27

PR #3222, opened 2025-07-28
feat: add GitHub Actions workflow for auto-building e2e-runner on new release branches

gustavolira:
## Description

Automate E2E Runner Image Build and Push to Quay via GitHub Actions

- Fixes https://issues.redhat.com/browse/RHIDP-6297

## PR acceptance criteria


.github/workflows/auto-release-branch-e2e.yaml
```diff
@@ -0,0 +1,108 @@
+name: Auto Build e2e-runner for New Release Branch
+
+on:
+  create:
+  push:
+    branches:
+      - 'release-1.**'
```
nickboldt: do you want this to trigger on ONLY pushes to the branch for the ibm dockerfile? Or on all files in the branch?

```suggestion
      - 'release-1.**'
   paths:
      - '.ibm/images/Dockerfile'
```

.github/workflows/auto-release-branch-e2e.yaml
```diff
@@ -0,0 +1,108 @@
+name: Auto Build e2e-runner for New Release Branch
+
+on:
+  create:
```
nickboldt: Does this need to be qualified so it doesn't fire for ALL topic branches and tags too?

```suggestion
  create:
    branches:
      - 'release-1.**'
```

.github/workflows/auto-release-branch-e2e.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+            # Check if tag already exists in Quay.io
+            echo "Checking if tag $BRANCH_NAME exists in Quay.io..."
+            QUAY_TAG_EXISTS=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh-e2e-runner/tag/" | jq -r --arg tag "$BRANCH_NAME" '.tags[]? | select(.name == $tag) | .name' 2>/dev/null || echo "")
+
+            if [[ -z "$QUAY_TAG_EXISTS" ]]; then
+              echo "✅ New release branch detected: $BRANCH_NAME (no Quay tag exists)"
+              echo "is_new_release=true" >> $GITHUB_OUTPUT
+            else
+              echo "⚠️ Release branch $BRANCH_NAME already has Quay tag, skipping build"
```
nickboldt: what about the case where we already have the floating `release-1.7` tag in quay, but you need respin it?

At https://quay.io/repository/rhdh-community/rhdh-e2e-runner?tab=tags I see
```
release-1.7-c139f560
release-1.7-79395d62
release-1.7
```

So you've clearly wanted more than one 1.7 build. ;) 

How would you trigger the respin if you're only searching for the `BRANCH_NAME` and not the `BRANCH_NAME-SHA` tag?

.github/workflows/auto-release-branch-e2e.yaml
gustavolira: The purpose of this workflow is to build the E2E runner only for release-1.* branches, and only when the branch is initially created.
Subsequent commits to the branch should not trigger a rebuild of the E2E runner.

We already have a separate workflow that builds the E2E runner whenever changes are made to the .ibm/Dockerfile.

.github/workflows/auto-release-branch-e2e.yaml
gustavolira: It's not necessary, the E2E runner is just responsible for building an image used to run end-to-end tests against our pipeline, and it's not related to RHDH itself.

nickboldt:
Instead of a whole new workflow to run on branch creation that duplicates https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/push-e2e-runner.yaml#L7-L9

why not just add this to the existing push-e2e-runner?

```
on
  create:
    branches:
      - 'release-1.**'
```

gustavolira:
> Instead of a whole new workflow to run on branch creation that duplicates https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/push-e2e-runner.yaml#L7-L9
> 
> why not just add this to the existing push-e2e-runner?
> 
> ```
> on
>   create:
>     branches:
>       - 'release-1.**'
> ```

I believe keeping separate workflows is better here because:
Tag existence check: This workflow checks if the tag already exists in Quay.io before building, avoiding unnecessary builds
Different triggers: create is specific for new branch creation, while push is for code changes
Specific logic: This has release branch detection logic that would complicate the existing workflow
The existing push-e2e-runner.yaml handles code changes, while this handles automatic builds for new release branches. Different responsibilities, different workflows.

schultzp2020:
> > Instead of a whole new workflow to run on branch creation that duplicates https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/push-e2e-runner.yaml#L7-L9
> > why not just add this to the existing push-e2e-runner?
> > ```
> > on
> >   create:
> >     branches:
> >       - 'release-1.**'
> > ```
> 
> I believe keeping separate workflows is better here because: Tag existence check: This workflow checks if the tag already exists in Quay.io before building, avoiding unnecessary builds Different triggers: create is specific for new branch creation, while push is for code changes Specific logic: This has release branch detection logic that would complicate the existing workflow The existing push-e2e-runner.yaml handles code changes, while this handles automatic builds for new release branches. Different responsibilities, different workflows.

I can go either way, but isn't it generally easier to have one workflow as long as the GitHub context stays the same?

.github/workflows/push-e2e-runner.yaml
```diff
@@ -57,20 +60,40 @@ jobs:
...
 
+      - name: Check if tag already exists in Quay.io (for release branches)
+        if: github.event_name == 'create' && startsWith(github.ref_name, 'release-1.')
+        run: |
+          BRANCH_NAME="${{ github.ref_name }}"
+          echo "Checking if tag $BRANCH_NAME exists in Quay.io..."
+          
+          QUAY_TAG_EXISTS=$(curl -s "https://quay.io/api/v1/repository/rhdh-community/rhdh-e2e-runner/tag/" | jq -r --arg tag "$BRANCH_NAME" '.tags[]? | select(.name == $tag) | .name' 2>/dev/null || echo "")
+          
+          if [[ -n "$QUAY_TAG_EXISTS" ]]; then
```
nickboldt, line 71: this will prevent respinning on subsequent pushes to the branch. You don't need this check.

gustavolira:
> > > Instead of a whole new workflow to run on branch creation that duplicates https://github.com/redhat-developer/rhdh/blob/main/.github/workflows/push-e2e-runner.yaml#L7-L9
> > > why not just add this to the existing push-e2e-runner?
> > > ```
> > > on
> > >   create:
> > >     branches:
> > >       - 'release-1.**'
> > > ```
> > 
> > 
> > I believe keeping separate workflows is better here because: Tag existence check: This workflow checks if the tag already exists in Quay.io before building, avoiding unnecessary builds Different triggers: create is specific for new branch creation, while push is for code changes Specific logic: This has release branch detection logic that would complicate the existing workflow The existing push-e2e-runner.yaml handles code changes, while this handles automatic builds for new release branches. Different responsibilities, different workflows.
> 
> I can go either way, but isn't it generally easier to have one workflow as long as the GitHub context stays the same?

I'm using one workflow now

PR #3242, opened 2025-07-30
chore(ci): switch PR workflow from pull_request to pull_request_target for secrets

subhashkhileri:
## Description

modify .github/workflows/pr.yaml to pull_request_target in order to access secrets for sealight integration for unit tests.

- Fixes #?

## PR acceptance criteria


.github/workflows/pr.yaml
```diff
@@ -26,12 +30,33 @@ concurrency:
...
+      is_internal: ${{ steps.check.outputs.is_internal }}
+    steps:
+      - name: Check if internal PR
+        id: check
+        run: |
+          if [[ "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
+            echo "is_internal=true" >> $GITHUB_OUTPUT
+            echo "✓ Internal PR - will run with PR code"
+          else
+            echo "is_internal=false" >> $GITHUB_OUTPUT
```
nickboldt: where are these variables being used in the script? below this line I only see `authorize` being called.

.github/workflows/pr.yaml
```diff
@@ -26,12 +30,33 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  authorize:
+    # The 'external' environment is configured with the maintainers team as required reviewers.
+    # All the subsequent jobs in this workflow 'need' this job, which will require manual approval for PRs coming from external forks.
+    # see list of approvers in OWNERS file
+    environment:
+      ${{ github.event.pull_request.head.repo.full_name == github.repository && 'internal' || 'external' }}
```
nickboldt, line 38: Do we need to include a list of people who can override/who don't need to be authorized, like we did in https://github.com/redhat-developer/rhdh-operator/blob/main/.github/workflows/pr.yaml#L18 ?

.github/workflows/pr.yaml
Zaperex: Good point, they aren't being used anywhere so I guess they're safe to remove.

.github/workflows/pr.yaml
Zaperex: We could, but that would be a giant list to maintain since this repo is used by alot more people compared to the `rhdh-operator` repo.

.github/workflows/pr.yaml
```diff
@@ -26,12 +30,33 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    outputs:
+      is_internal: ${{ steps.check.outputs.is_internal }}
```
Zaperex:
```suggestion
```

.github/workflows/pr.yaml
Zaperex:
```suggestion
```

.github/workflows/pr.yaml
```diff
@@ -26,12 +30,33 @@ concurrency:
...
+    steps:
+      - name: Check if internal PR
+        id: check
+        run: |
+          if [[ "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
+            echo "is_internal=true" >> $GITHUB_OUTPUT
```
Zaperex:
```suggestion
```

.github/workflows/pr.yaml
nickboldt: or just the top 5 major contributors? doesn't have to be everyone, just the managing partners, so to speak ;)

schultzp2020: I can go either way, although I am fine with it as is.

Zaperex: That sounds good, which 5 would that be?

Zaperex: I'll merge this for now so it unblocks @subhashkhileri since you guys seem to be fine with it for now. Need to force merge due to the required `Build on Node.js 22` and `Test on Node.js 22` jobs were modified here and can't trigger.

PR #3260, opened 2025-08-04
fix(app): add mui5 workaround to the shell

gashcrumb:
This change moves the statically applied classname generator code from each individual wrapper to the application shell to ensure that all plugins using MUI5 have consistent class names applied to those components by the theme provider.

## Description

- Fixes [RHDHBUGS-986](https://issues.redhat.com/browse/RHDHBUGS-986)

## PR acceptance criteria

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -21,6 +21,7 @@
   "major": {
     "dependencyDashboardApproval": true
   },
+  "reviewers": ["team:rhdh-security-new"],
   "packageRules": [    
     {
       "matchCategories": [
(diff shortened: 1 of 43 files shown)


dzemanov:
@gashcrumb does this mean that the plugins don't have to include the workaround in their code now? Not sure with what version of global header you have tested, it contains the workaround [here](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/global-header/plugins/global-header/src/index.ts#L18-#L22).

gashcrumb:
@dzemanov right, so with this change plugins would not need to include that workaround.  If the plugin includes code like you've linked, it won't be a problem as it includes a check if the CSS class prefix has already been added.

gashcrumb:
@dzemanov awesome, thanks for checking that!  I'll have a look.

/hold

gashcrumb:
it's looking like this will require a cli change instead, in which case these changes won't be necessary.

PR #3274, opened 2025-08-07
fix(extensions): remove verified badge of servicenow tile

ciiay:
## Description

Address [RHDH 1.7 Test Day Feedback/issues sheet row 12 issue](https://docs.google.com/spreadsheets/d/14O4IpU3NomiqorkS8Pd8FD7trLm18ZSQg2C9L1C8O2A/edit?gid=88072381#gid=88072381).

- Fixes [RHDHBUGS-1937](https://issues.redhat.com/browse/RHDHBUGS-1937)

## PR acceptance criteria

### DIFF
--- a/catalog-entities/marketplace/plugins/servicenow.yaml
+++ b/catalog-entities/marketplace/plugins/servicenow.yaml
@@ -13,7 +13,8 @@ metadata:
     operational awareness.
 
   annotations:
-    extensions.backstage.io/pre-installed: 'false'
+    # temporarily set to 'true' to remove the 'custom plugin' badge, change this to 'false' when RHDHBUGS-1945 gets resolved
+    extensions.backstage.io/pre-installed: 'true'
   tags:
     - itsm
     - incident-management
--- a/catalog-entities/marketplace/plugins/servicenow.yaml
+++ b/catalog-entities/marketplace/plugins/servicenow.yaml
@@ -14,7 +14,6 @@ metadata:
 
   annotations:
     extensions.backstage.io/pre-installed: 'false'
-    extensions.backstage.io/verified-by: Red Hat
   tags:
     - itsm
     - incident-management
--- a/catalog-entities/marketplace/plugins/servicenow.yaml
+++ b/catalog-entities/marketplace/plugins/servicenow.yaml
@@ -13,8 +13,7 @@ metadata:
     operational awareness.
 
   annotations:
-    # temporarily set to 'true' to remove the 'custom plugin' badge, change this to 'false' when RHDHBUGS-1945 gets resolved
-    extensions.backstage.io/pre-installed: 'true'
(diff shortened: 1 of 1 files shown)


catalog-entities/marketplace/plugins/servicenow.yaml
```diff
@@ -14,7 +14,6 @@ metadata:
 
   annotations:
     extensions.backstage.io/pre-installed: 'false'
```
dzemanov:
```suggestion
    extensions.backstage.io/pre-installed: 'true'
```

dzemanov (changes requested):
We are adding badge `Custom plugin` for all plugins that are not pre-installed. Even if you do not specify this annotation, marketplace catalog processor will add this annotation set to false. We will need to change the behavior of showing Custom badge on pre-installed set to false now that we are moving away from wrappers. But for now, to remove Custom plugin badge, since ServiceNow should have `no badge`, I would go with this quick fix.

Screenshot 2025-08-08 at 9 20 43

dzemanov:
Created issue for this: [RHDHBUGS-1945](https://issues.redhat.com/browse/RHDHBUGS-1945)

christoph-jerolimov (approved):
Lgtm, also if the comment isn't correct.

The annotation means "extensions yaml is pre-installed" not that the plugin itself is installed or enabled.

We need to document this fact here https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/marketplace/docs/catalog/plugins.md

catalog-entities/marketplace/plugins/servicenow.yaml
```diff
@@ -13,8 +13,8 @@ metadata:
     operational awareness.
 
   annotations:
-    extensions.backstage.io/pre-installed: 'false'
-    extensions.backstage.io/verified-by: Red Hat
+    # temporarily set to 'true' to remove the 'custom plugin' badge, change this to 'false' when RHDHBUGS-1945 gets resolved
+    extensions.backstage.io/pre-installed: 'true'
```
christoph-jerolimov: Hi, same as in #3275

I understand that this annotation confused people.

The annotation means "extensions yaml is pre-installed" not that the plugin itself is installed or enabled.

We need to document this fact here https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/marketplace/docs/catalog/plugins.md

Can you change the comment so that it doesn't sounds like we have a bug here:

```suggestion
    extensions.backstage.io/pre-installed: 'true' # this means the plugin yaml is preinstalled, not the plugin itself
```

christoph-jerolimov (changes requested):

christoph-jerolimov (approved):
Thanks @ciiay 👍 

/lgtm
/approve

PR #3307, opened 2025-08-20
spike: set the same tag expiry for ALL floating tags (not just the first one published RHIDP-8270

polasudo:
## Description

This pull request refactors the way image expiry labels are set in the `next-build-image.yaml` GitHub Actions workflow. Instead of hardcoding expiry durations in multiple places, the expiry value is now computed once based on the branch name and reused throughout the workflow. This improves maintainability and reduces the risk of inconsistencies.

**Workflow expiry label improvements:**

* Compute the `expires_after` value once based on the branch name (`next` or `next-*` gets `14d`, others get `183d`) and store it in the environment for later use. [[1]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9R95-R102) [[2]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9R187-R194)
* Update all occurrences of the `quay.expires-after` image label to use the computed `${{ env.EXPIRES_AFTER }}` variable instead of hardcoded values. [[1]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9L110-R118) [[2]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9L126-R134) [[3]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9L207-R223) [[4]](diffhunk://#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9L221-R237)

**Manifest creation improvements:**

* Update the manifest creation step to annotate each image tag with the correct expiry value using the `${EXPIRES_AFTER}` environment variable, ensuring Quay applies the correct expiry per tag.

- Fixes #?
[RHIDP-8270](https://issues.redhat.com/browse/RHIDP-8270)
## PR acceptance criteria


.github/workflows/next-build-image.yaml
```diff
@@ -176,6 +211,26 @@ jobs:
             echo "REF_NAME=$ref_name" >> $GITHUB_ENV
             echo "REF_NAME_SHORT=${ref_name_short}" >> $GITHUB_ENV
 
+            # compute expiry once and reuse
+            if [[ "$ref_name_short" == "next" ]] || [[ "$ref_name_short" == next-* ]]; then
+              expires_after="14d"  # Short expiry for next builds
+            elif [[ "$ref_name_short" =~ ^[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
```
nickboldt: this could be simpler ... just need this single if block and in all other cases fall back to 14d expiry.

.github/workflows/next-build-image.yaml
```diff
@@ -176,6 +211,26 @@ jobs:
...
+              expires_after="183d" # Long expiry for official releases (1.2, 1.2.3, etc.)
+            else
+              expires_after="14d"  # Short expiry for feature branches and other temporary builds
+            fi
+            echo "EXPIRES_AFTER=$expires_after" >> $GITHUB_ENV
+
+
```
nickboldt: lots of empty lines here.

.github/workflows/next-build-image.yaml
```diff
@@ -244,3 +313,208 @@ jobs:
           if: ${{ env.REF_NAME_SHORT != 'next' && !startsWith(env.REF_NAME_SHORT,'next-') }}
           run: |
             docker buildx imagetools inspect ${{ env.REGISTRY }}/${{ env.REGISTRY_IMAGE }}:${{ steps.meta-release.outputs.version }}
+
+        - name: Wait for manifest propagation
+          if: env.HAS_QUAY_AUTH == 'true'
+          run: |
+            echo "Waiting for Quay to propagate multi-arch manifests..."
+            sleep 15
```
nickboldt: only 15 seconds is enough? should this be a for loop with timeout and checks every few seconds?

what if wafter 15 seconds it hasn't propagagted due to some performance issue at quay or an outage?

.github/workflows/next-build-image.yaml
```diff
@@ -244,3 +313,208 @@ jobs:
...
+            REF_NAME: ${{ env.REF_NAME }}
+            REF_NAME_SHORT: ${{ env.REF_NAME_SHORT }}
+            SHORT_SHA: ${{ env.SHORT_SHA }}
+            EXPIRES_AFTER: ${{ env.EXPIRES_AFTER }}
+            QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+            QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+            QUAY_OAUTH_TOKEN: ${{ secrets.QUAY_OAUTH_TOKEN }}
+          run: |
+            set -e
+            echo "Managing image lifecycle: setting expiry and cleaning up per-arch images..."
```
nickboldt:
```suggestion
            echo "Set multi-arch image expiry and remove single-arch images..."
```

.github/workflows/next-build-image.yaml
```diff
@@ -107,7 +136,9 @@ jobs:
           imageTags: |
             type=raw,value=${{ env.REF_NAME }}-${{ env.PLATFORM_ARCH }}
             type=raw,value=${{ env.REF_NAME }}-${{ env.SHORT_SHA }}-${{ env.PLATFORM_ARCH }}
-          imageLabels: quay.expires-after=14d
+          imageLabels: |
+            quay.expires-after=${{ env.EXPIRES_AFTER }}
+            org.opencontainers.image.expires=${{ env.EXPIRES_AFTER }}
```
Zaperex: Would it make sense to just combine the `build-next` and `build-release` steps into one step? The only thing that was different was the `imageLabels`, but with your change they both reference `EXPIRES_AFTER` and we won't need to have the `if` check anymore. (tried selecting the entire step for both but github UI doesn't allow that it seems).

.github/workflows/next-build-image.yaml
```diff
@@ -123,7 +154,9 @@ jobs:
           imageTags: |
             type=raw,value=${{ env.REF_NAME }}-${{ env.PLATFORM_ARCH }}
             type=raw,value=${{ env.REF_NAME }}-${{ env.SHORT_SHA }}-${{ env.PLATFORM_ARCH }}
-          imageLabels: quay.expires-after=183d
+          imageLabels: |
+            quay.expires-after=${{ env.EXPIRES_AFTER }}
+            org.opencontainers.image.expires=${{ env.EXPIRES_AFTER }}
```
Zaperex, line 136: Same as https://github.com/redhat-developer/rhdh/pull/3307/files#r2310754480

.github/workflows/next-build-image.yaml
```diff
@@ -92,6 +94,33 @@ jobs:
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
           echo "PLATFORM_ARCH=${platform#*/}" >> $GITHUB_ENV
 
+          # compute expiry once and reuse
+          if [[ "$ref_name_short" == "next" ]] || [[ "$ref_name_short" == next-* ]]; then
+            expires_after="14d"  # Short expiry for next builds
+          elif [[ "$ref_name_short" =~ ^[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
```
Zaperex: Nick's comment (https://github.com/redhat-developer/rhdh/pull/3307/files#r2310518099) can also apply here

.github/workflows/next-build-image.yaml
```diff
@@ -176,6 +211,26 @@ jobs:
...
+
+
+
+        - name: Check Quay credentials
+          run: |
+            if [ "${{ env.HAS_QUAY_AUTH }}" != "true" ]; then
+              echo "::error::Missing QUAY_USERNAME or QUAY_TOKEN secrets"
+              exit 1
+            fi
+            echo "✓ Quay credentials are available"
```
Zaperex: Is this check necessary? Don't we already check this earlier?
https://github.com/redhat-developer/rhdh/pull/3307/files#diff-d6d011dd073e8d52787aa77626c58ab508c37815abc215304fa310e5ee563ca9R109-R115

polasudo: since there are 2 paralel runs we need those checks for both, in that case we need it twice

Zaperex: I mean in this case, they're sequential since the `merge` job `needs` the `build-image` job. And if the quay credentials check in `build-image` passes, then it would imply the check succeeds in the `merge` job (unless the secrets changed somehow between jobs which is highly unlikely)

polasudo: yeah i get it it will be deleted in next commit

nickboldt:
merged; manually triggered in https://github.com/redhat-developer/rhdh/actions/runs/17744786774

nickboldt:
cherrypick failed so manually pushed to 1.7 as https://github.com/redhat-developer/rhdh/commit/e5e5628c5a534feda8beefe86a8be0591003ca0c

PR #3352, opened 2025-09-02
feat: allow tag omission in OCI references for dynamic plugins

Zaperex:
## Description

- Adds support for dynamic plugin version inheritance for OCI packages via a `{{inherit}}` tag.
- Script also now throws an error if duplicate plugins are found across config files in the `includes` field. 
  - It also throws errors if there are duplicate plugins in the `dynamic-plugins.yaml`.
  - Essentially any duplicates found in the same "level" will throw an error. 

- Fixes [RHIDP-8215](https://issues.redhat.com/browse/RHIDP-8215)?

## PR acceptance criteria


docker/install-dynamic-plugins.py
```diff
@@ -40,15 +40,16 @@
...
 #
 # It expects the `dynamic-plugins.yaml` file to be present in the current directory and
 # to contain the list of plugins to install along with their optional configuration.
 #
 # The `dynamic-plugins.yaml` file must contain:
 #   - a `plugins` list of objects with the following properties:
-#     - `package`: the NPM package to install (either a package name or a path to a local package)
-#     - `integrity`: a string containing the integrity hash of the package (optional if package is local, as integrity check is not checked for local packages)
+#     - `package`: the NPM or OCI package to install (either a package name or a path to a local package)
+#       - For OCI packages, the tag or digest can be omitted if the `{{inherit}}` tag is set (requires the included configuration to contain a valid tag or digest to inherit from)
```
davidfestal:
```suggestion
#       - For OCI packages, the tag or digest can be replaced by the `{{inherit}}` tag (requires the included configuration to contain a valid tag or digest to inherit from)
```

docker/install-dynamic-plugins.py
```diff
@@ -100,21 +101,79 @@ def maybeMergeConfig(config, globalConfig):
...
+def mergePlugin(plugin: dict, allPlugins: dict, dynamicPluginsFile: str, level: int):
     package = plugin['package']
+    version = None # Used to track OCI package versions for version inference
     if not isinstance(package, str):
         raise InstallException(f"content of the \'plugins.package\' field must be a string in {dynamicPluginsFile}")
 
+    if package.startswith('oci://'):
+        # Check if package is in the expected format: 
+        # Tag format: `oci://<registry>:<tag>!<path>`
+        # SHA digest format: `oci://<registry>@sha<algo>:<digest>!<path>`
```
davidfestal: Do we also support the case when both a digest and a tag are provided ?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -1,22 +1,25 @@
...
 
-Plugins are defined in the `plugins` array in the `dynamic-plugin-config.yaml` file. Each plugin is defined as an object with the following properties:
+Plugins are defined in the `plugins` array in the `dynamic-plugins.yaml` file. Each plugin is defined as an object with the following properties:
 
-- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path.
+- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path. For OCI packages, the tag or digest can be omitted if the `{{inherit}}` tag is set (requires the included configuration to contain a valid tag or digest to inherit from)
```
davidfestal:
```suggestion
- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path. For OCI packages, the tag or digest can be replaced by the `{{inherit}}` tag (requires the included configuration to contain a valid tag or digest to inherit from)
```

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -1,22 +1,25 @@
...
 - `disabled`: A boolean value that determines whether the plugin is enabled or disabled.
 - `integrity`: The integrity hash of the package. This is required for `tgz` archives and npm packages.
-- `pluginConfig`: The configuration for the plugin. For backend plugins this is optional and can be used to pass configuration to the plugin. For frontend this is required, see [Frontend Plugin Wiring](frontend-plugin-wiring.md) for more information on how to configure bindings and routes. This is fragment o the `app-config.yaml` file. Anything that is added to this object will be merged with the main app-config.yaml file.
+- `inheritVersion`: A boolean value that determines whether the plugin (OCI package only) attempts to inherit it's version from.
```
davidfestal: We don't use `inheritVersion` anymore right ?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -62,14 +65,43 @@ plugins:
...
+
+For example, if we have an included dynamic plugin file (`dynamic-plugins.example.yaml`) with `v0.0.2` of our plugin which might be updated to match the current RHDH version:
+
+```yaml
+plugins:
+  - disabled: true
+    package: oci://quay.io/example/image:v0.0.2!backstage-plugin-myplugin
+```
+
+and a default `dynamic-plugins.yaml` file with the `{{inherit}}` tag using configurations for an older version that is still compatible:
```
davidfestal: why a *default* `dynamic-plugins.yaml` ?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -62,14 +65,43 @@ plugins:
...
+
+```yaml
+included:
+- dynamic-plugins.example.yaml
+plugins:
+  - disabled: false
+    package: oci://quay.io/example/image:{{inherit}}!backstage-plugin-myplugin
+    inheritVersion: true
```
davidfestal: I guess we don't need `inheritVersion` here.

docker/install-dynamic-plugins.py
```diff
@@ -100,21 +101,79 @@ def maybeMergeConfig(config, globalConfig):
...
+        version = tag_version if tag_version else digest_version
+        package = f"{registry}:!{path}"
+
     # if `package` already exists in `allPlugins`, then override its fields
     if package not in allPlugins:
+        if package.startswith('oci://'):
+            if inheritVersion is True:
+                # User might layer on additional configurations that provide a version so just print a warning for now
+                print(f"WARNING: {{{{inherit}}}} tag is set and there is currently no resolved tag or digest for {plugin['package']} in {dynamicPluginsFile}. This can be safely ignored if no errors are reported later.")
+                plugin["version"] = None
```
davidfestal: I don't what you mean by 
> User might layer on additional configurations that provide a version so just print a warning for now
Can you provide a bit more explanation ?

docker/install-dynamic-plugins.py
Zaperex: If you mean something like `oci://<registry>:<tag>@sha<algo>:<digest>!<path>`, my code currently does not support that. I'll add support for it.

docs/dynamic-plugins/installing-plugins.md
Zaperex: Ah shoot forgot to get rid of that line :sweat:

docs/dynamic-plugins/installing-plugins.md
Zaperex: Doesn't the script search for a `dynamic-plugins.yaml` file to load configurations from? 
https://github.com/redhat-developer/rhdh/blob/c1969a043208319250e03260fda19393d8ec812f/docker/install-dynamic-plugins.py#L264

docker/install-dynamic-plugins.py
Zaperex: ah I guess it's in case they put a plugin with `{{inherit}}` tag in the `includes` file which they then override with a proper version in the main `dynamic-plugins.yaml` file. Alternatively, we can just disallow `{{inherit}}` in the `includes` config files since it wouldn't make sense to do that in those?

docs/dynamic-plugins/installing-plugins.md
Zaperex: I guess I should just exclude the `default` word and just say `dynamic-plugins.yaml` since we mentioned that's the main file to add configs in.

docker/install-dynamic-plugins.py
Zaperex: @davidfestal it seems that the `OCIDownloader` does not support having both tag and digest `<tag>@<algo>:<digest>` at https://github.com/redhat-developer/rhdh/blob/c6734b1252eb7b5be1bbbc31b8f9cc2728ddfac2/docker/install-dynamic-plugins.py#L145

`skopeo` does not seem to support that format. For example:
```bash
$ skopeo inspect docker://quay.io/gashcrumb/example-root-http-middleware:latest@sha256:78105797a0323e930e17008194b220fd5a9
feef68f8e6a35c24dfc072161da04
FATA[0000] Error parsing image name "docker://quay.io/gashcrumb/example-root-http-middleware:latest@sha256:78105797a0323e930e17008194b220fd5a9feef68f8e6a35c24dfc072161da04": Docker references with both a tag and digest are currently not supported
```

Do we want to modify the `OCIDownloader` to support this format?

davidfestal: No, we can keep that for a next PR.

docker/install-dynamic-plugins.py
```diff
@@ -100,21 +102,88 @@ def maybeMergeConfig(config, globalConfig):
...
+        path = match.group(4) 
+        
+        # {{inherit}} tag indicates that the version should be inherited from the included configuration. Must NOT have a SHA digest included.
+        inheritVersion = (tag_version == "{{inherit}}" and digest_version == None)
+        package = f"{registry}:!{path}"
+
     # if `package` already exists in `allPlugins`, then override its fields
     if package not in allPlugins:
+        if package.startswith('oci://'):
+            if inheritVersion is True:
```
gashcrumb: Is there any way to reduce the nested `if` statements here and on line 165-166?  This kind of nesting tends to lead to difficult to follow logic, and if we make this kind of test more than once maybe it'd be more clear in intent to have a helper function that makes this test.

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -30,15 +32,15 @@ plugins:
     package: ./dynamic-plugins/dist/backstage-plugin-catalog-backend-module-github-dynamic
 ```
 
-While the plugin's default configuration comes from the `dynamic-plugins.default.yaml` file, you still have the option to replace it by incorporating a `pluginConfig` entry into the plugin configuration.
+While the plugin's default configuration comes from the `dynamic-plugins.default.yaml` file, you still have the option to override it by incorporating a `pluginConfig` entry into the plugin configuration.
```
davidfestal, line 35: replace is better here I guess, because there's not merging of the individual feeds between the default config and the overridden one. it's a all-or-nothing config overriding, e-hence the word "replace".

davidfestal (commented):
very small doc change proposal

docs/dynamic-plugins/installing-plugins.md
Zaperex: I personally feel like `override` fits better since `replace` tends to mean we're discarding the default config completely, but here we are just providing another config on top of the default configs.

davidfestal: As you like

Zaperex:
I would like to add some unit tests and the unique key for NPM, unless I should do this in a separate PR?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -1,22 +1,24 @@
...
+The placement of `dynamic-plugins.yaml` depends on the deployment method.
+For more information, see [Installing Dynamic Plugins with the Red Hat Developer Hub Operator](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.7/html/installing_and_viewing_plugins_in_red_hat_developer_hub/rhdh-installing-rhdh-plugins_title-plugins-rhdh-about#proc-config-dynamic-plugins-rhdh-operator_rhdh-installing-rhdh-plugins) or [Installing Dynamic Plugins Using the Helm Chart](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.7/html/installing_and_viewing_plugins_in_red_hat_developer_hub/rhdh-installing-rhdh-plugins_title-plugins-rhdh-about#con-install-dynamic-plugin-helm_rhdh-installing-rhdh-plugins).
```
christoph-jerolimov, line 6: You can reference the /latest/ doc here instead or?

docker/install-dynamic-plugins.py
davidfestal: @gashcrumb do you consider it resolved ?

docs/dynamic-plugins/installing-plugins.md
Zaperex: I guess we could but I prefer to stick with exact versions. Main concern with `latest` would be dead links since these doc guides titles don't always stay the same between versions.

nickboldt: yeah with the way CCS likes to reorg things, you can't assume a title will last more than 3 months. Also with the AEM migration all bets are off and all links might be changing.

PR #3410, opened 2025-09-12
feat(i18n): enhance translation-backend api response

debsmita1:
## Description

This PR addresses the comments here https://github.com/redhat-developer/rhdh/pull/3365#pullrequestreview-3211837185

- Logs the api failures in the console
- The merged translations are computed only once on the first request. Subsequent API requests return the cached result without re-reading the JSON files. This avoids unnecessary file system reads and JSON parsing on every request, improving performance.
- Ensures that only configured locales from app-config.yaml are included in the cached translations. Prevents downloading unnecessary languages.

- Fixes 
Part of https://github.com/redhat-developer/rhdh/pull/3365

## PR acceptance criteria


plugins/translation-backend/src/service/router.ts
```diff
@@ -15,43 +19,51 @@ export async function createRouter({
...
+          configuredLocales,
+        );
       }
 
-      if (Object.keys(mergedTranslations).length === 0) {
+      if (!cachedTranslations || Object.keys(cachedTranslations).length === 0) {
         res.status(404).json({
-          error: 'No valid translation overrides found in provided files',
+          error: 'No valid translation overrides found in the provided files',
         });
```
karthikjeeyar: I think we should differentiate the errors here for `file/resource not found` vs `empty data` after filtering by configured locales, using `404` for both these isn't semantically correct.

Suggestion is to return different error code:

404: When translation files are missing/invalid (actual resource problem)
200: When files exist but no translations match configured locales (successful but empty result {} )

PR #3411, opened 2025-09-12
feat(segment): update segment wrapper and add env vars

rohitkrai03:
## Description

- Updates the Segment wrapper with new version of the plugin.
- Updates dockerfile to include two new env vars for `RHDH_VERSION` and `BACKSTAGE_VERSION`.
- Using the new env vars in segment config.

- Fixes https://issues.redhat.com/browse/RHIDP-8618

## PR acceptance criteria

### DIFF
--- a/catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
+++ b/catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
@@ -84,6 +84,9 @@ spec:
           # highlight-start
           writeKey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
           maskIP: true # prevents IP addresses from being sent if true
+          testMode: false # prevents data from being sent if true
+          appVersion: ${RHDH_VERSION} # RHDH application version
+          backstageVersion: ${BACKSTAGE_VERSION} # Backstage version
           # highlight-end
     ```
 
--- a/catalog-entities/marketplace/packages/backstage-community-plugin-analytics-provider-segment.yaml
+++ b/catalog-entities/marketplace/packages/backstage-community-plugin-analytics-provider-segment.yaml
@@ -18,7 +18,7 @@ metadata:
 spec:
   packageName: "@backstage-community/plugin-analytics-provider-segment"
   dynamicArtifact: ./dynamic-plugins/dist/backstage-community-plugin-analytics-provider-segment
-  version: 1.16.0
+  version: 1.17.0
   backstage:
     role: frontend-plugin
     supportedVersions: 1.39.1
@@ -47,3 +47,5 @@ spec:
               writeKey: ${SEGMENT_WRITE_KEY}
               maskIP: true
               testMode: ${SEGMENT_TEST_MODE}
+              appVersion: ${RHDH_VERSION}
+              backstageVersion: ${BACKSTAGE_VERSION}
--- a/dynamic-plugins.default.yaml
+++ b/dynamic-plugins.default.yaml
@@ -1128,6 +1128,8 @@ plugins:
             writeKey: ${SEGMENT_WRITE_KEY}
             maskIP: true # prevents IP addresses from being sent if true
             testMode: ${SEGMENT_TEST_MODE} # prevents data from being sent if true
+            appVersion: ${RHDH_VERSION} # RHDH application version
+            backstageVersion: ${BACKSTAGE_VERSION} # Backstage version
 
   - package: ./dynamic-plugins/dist/backstage-community-plugin-catalog-backend-module-scaffolder-relation-processor-dynamic
     disabled: true
--- a/dynamic-plugins/wrappers/backstage-community-plugin-analytics-provider-segment/package.json
+++ b/dynamic-plugins/wrappers/backstage-community-plugin-analytics-provider-segment/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backstage-community-plugin-analytics-provider-segment",
-  "version": "1.16.0",
+  "version": "1.17.0",
   "main": "src/index.ts",
   "types": "src/index.ts",
   "license": "Apache-2.0",
@@ -28,7 +28,7 @@
     "export-dynamic:clean": "run export-dynamic --clean"
   },
   "dependencies": {
-    "@backstage-community/plugin-analytics-provider-segment": "1.16.0"
+    "@backstage-community/plugin-analytics-provider-segment": "1.17.0"
   },
   "devDependencies": {
     "@backstage/cli": "0.30.0",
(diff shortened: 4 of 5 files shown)


.rhdh/docker/Dockerfile
```diff
@@ -136,6 +136,14 @@ RUN \
   -e "s/(\"Last Commit\": \"(.+)\")/\1, \"Build Time\": \"$now\"/" && \
   cat packages/app/src/build-metadata.json; echo
 
+# Extract metadata values using grep/sed (no jq dependency) and create environment script
+RUN RHDH_VERSION=$(grep '"RHDH Version"' packages/app/src/build-metadata.json | sed 's/.*": "\([^"]*\)".*/\1/') && \
```
nickboldt: should use jq here since it's json text.

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
```diff
@@ -84,6 +84,9 @@ spec:
           # highlight-start
           writeKey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
           maskIP: true # prevents IP addresses from being sent if true
+          testMode: false # prevents data from being sent if true
+          appVersion: "1.0.0" # RHDH application version
+          backstageVersion: "1.39.1" # Backstage version
```
nickboldt: hardcoded instead of based on build metadata .json file?

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
```diff
@@ -84,6 +84,9 @@ spec:
           # highlight-start
           writeKey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
           maskIP: true # prevents IP addresses from being sent if true
+          testMode: false # prevents data from being sent if true
+          appVersion: "1.0.0" # RHDH application version
```
nickboldt: version 1.0?? the plugin is version 1.17 and the RHDH version is 1.8 -- is this being overwritten during a build process somewhere?

.rhdh/docker/Dockerfile
```diff
@@ -324,6 +336,8 @@ COPY --chown=1001:1001 $EXTERNAL_SOURCE_NESTED/packages/backend/src/instrumentat
 RUN chmod a=r ./instrumentation.js
 
 ENV NODE_OPTIONS="--no-node-snapshot"
-ENTRYPOINT ["node", "--require", "./instrumentation.js", "packages/backend", "--config", "app-config.yaml", "--config", "app-config.example.yaml", "--config", "app-config.example.production.yaml"]
+
+# Use bash -l -c to ensure a login shell is invoked, which sources /etc/profile.d/
+ENTRYPOINT ["bash", "-l", "-c", "node --require ./instrumentation.js packages/backend --config app-config.yaml --config app-config.example.yaml --config app-config.example.production.yaml"]
 
 # append Brew metadata here
```
nickboldt: we could also set these new env vars as container ENVs here: https://gitlab.cee.redhat.com/rhidp/rhdh/-/blob/rhdh-1-rhel-9/build/ci/sync-midstream.sh?ref_type=heads#L701-702

that way you can compute them with `skopeo inspect` rather than needing to fetch and run the container.

.rhdh/docker/Dockerfile
rohitkrai03: I tried using jq first but it was failing in CI with this error `No match for argument: jq
Error: Unable to find a match: jq`. Seems like it was not available in RH UBI images.

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
rohitkrai03: This is docs part of the marketplace manifest. We don't need to have the current version here.

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
rohitkrai03: This is docs part of the marketplace manifest. Just sample versions shown in the docs.

.rhdh/docker/Dockerfile
rohitkrai03: Yeah, that would also work but I don't have a lot of knowledge about that script. I can take a look if you want.

.rhdh/docker/Dockerfile
nickboldt: ah fair enough, I thought we had it installed.

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
nickboldt: well it's strange to set an example that doesn't reflect any sort of reality, but I guess YMMV.

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
nickboldt: so nothing will every read this and no one will ever wonder if it drifts out of date when we move to BS 1.42 in RHDH 1.8?

.rhdh/docker/Dockerfile
nickboldt: applied in https://gitlab.cee.redhat.com/rhidp/rhdh/-/commit/ca47ebea2b3817dc8a751a1639d6d087e07271f1

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
rohitkrai03: This is just documentation of how the plugin should be configured while installing. Even the writeKey value is just random here `writeKey: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`. The values could be different for different use cases. Its similar to the actual documentation for the plugin - https://github.com/backstage/community-plugins/tree/main/workspaces/analytics/plugins/analytics-provider-segment#version-tracking. I can update it to the current values but it doesn't need to be updated on every release.

.rhdh/docker/Dockerfile
rohitkrai03: So, now I can directly get the values using `skopeo inspect`?

catalog-entities/marketplace/plugins/analytics-provider-segment.yaml
rohitkrai03: Upon reconsideration, I have updated this to use ENV vars. Doesn't hurt to keep this updated with actual RHDH versions since the marketplace is primarily used in RHDH.

.rhdh/docker/Dockerfile
rohitkrai03: Updated.

nickboldt (approved):

PR #4154, opened 2026-02-02
chore(fix): ocp operator fix

teknaS47:
## Description

- Fixes [RHDHBUGS-2184](https://issues.redhat.com/browse/RHDHBUGS-2184)

## PR acceptance criteria


teknaS47:
Blocked by operator image rebuild, install-dynamic-plugins init container crash is unrelated to this PR.

.ci/pipelines/lib/orchestrator.sh
```diff
@@ -181,6 +181,167 @@ _orchestrator::apply_manifests() {
...
+  return 0
+}
+
+# Function: _orchestrator::create_knative_broker
+# Description: Create a Knative Eventing Broker in the given namespace.
+#   The SonataFlow Operator uses this broker to route workflow status
+#   events (via SinkBindings/Triggers) between workflows and data-index.
+# Arguments:
+#   $1 - namespace: Kubernetes namespace
+_orchestrator::create_knative_broker() {
```
sanketpathak, line 260: Do we need to create a Knative Broker when running with the Operator job?

teknaS47: No, it was added in an earlier iteration, and I removed it now.

.ci/pipelines/jobs/ocp-operator.sh
```diff
@@ -133,8 +129,8 @@ run_operator_runtime_config_change_tests() {
 }
 
 handle_ocp_operator() {
-  export NAME_SPACE="${NAME_SPACE:-showcase}"
-  export NAME_SPACE_RBAC="${NAME_SPACE_RBAC:-showcase-rbac}"
+  export NAME_SPACE="showcase-operator"
```
gustavolira: These were `${NAME_SPACE:-showcase}` / `${NAME_SPACE_RBAC:-...}` before — now they're hard-coded, while `NAME_SPACE_RUNTIME` on the next line still keeps its override. Was dropping the override intentional? Keeping `${NAME_SPACE:-showcase-operator}` would stay consistent and still let local-run.sh reuse the function with a custom namespace.

.ci/pipelines/lib/orchestrator.sh
```diff
@@ -268,10 +429,9 @@ _orchestrator::patch_workflow_postgres() {
   local database_name=${8:-}
   local database_schema=${9:-}
 
-  local service_ref="{\"name\": \"$svc_name\", \"namespace\": \"$svc_namespace\""
-  if [[ -n "$database_name" ]]; then
-    service_ref+=", \"databaseName\": \"$database_name\""
-  fi
+  local db="${database_name:-postgres}"
```
gustavolira: This used to omit `databaseName` entirely when empty; now it always defaults to `postgres`. The operator path uses inline yq instead, but `orchestrator::deploy_workflows` (Helm) still calls this — so Helm callers that pass an empty `database_name` now get `postgres` injected where the field was previously absent. Is that change to the Helm path intended?

.ci/pipelines/lib/orchestrator.sh
```diff
@@ -574,19 +724,229 @@ orchestrator::deploy_workflows_operator() {
...
+      export _SF_DI_URL="http://sonataflow-platform-data-index-service.${namespace}.svc.cluster.local"
+      export _SF_SERVICE_URL="http://${workflow}.${namespace}.svc.cluster.local"
+      yq eval -i '
+        del(.status) |
+        .spec.persistence.postgresql.secretRef.name = strenv(_SF_SECRET) |
+        .spec.persistence.postgresql.secretRef.userKey = strenv(_SF_UKEY) |
+        .spec.persistence.postgresql.secretRef.passwordKey = strenv(_SF_PKEY) |
+        .spec.persistence.postgresql.serviceRef.name = strenv(_SF_SVC) |
+        .spec.persistence.postgresql.serviceRef.namespace = strenv(_SF_NS) |
+        .spec.persistence.postgresql.serviceRef.databaseName = "postgres" |
```
gustavolira, line 784: Workflows get `databaseName: postgres` here while data-index/jobs-service use the dedicated `sonataflow` DB — and the old operator code patched workflows with `backstage_plugin_orchestrator`. Could we add a short comment on why workflows persist into `postgres` rather than their own DB? It's an easy thing to second-guess later.

.ci/pipelines/lib/orchestrator.sh
```diff
@@ -574,19 +724,229 @@ orchestrator::deploy_workflows_operator() {
...
+    fi
+    oc apply -f "${workflow_manifests}" -n "$namespace"
+
+    _orchestrator::wait_for_sonataflow_reconciliation "$namespace" "$workflow" 60
+    oc rollout status deployment/"$workflow" -n "$namespace" --timeout=600s
   done
 
   _orchestrator::wait_for_workflow_deployments "$namespace"
+
+  log::info "=== Workflow Service Diagnostics ==="
```
gustavolira: nit: This diagnostics block (plus the GraphQL dump, per-workflow env greps, and the 200-line log grep further down) runs on every operator deploy. Really handy for bringing this up — could we gate the verbose parts behind a debug flag (the repo already uses `ISRUNNINGLOCALDEBUG`) so steady-state CI logs stay readable?

.ci/pipelines/lib/orchestrator.sh
```diff
@@ -574,19 +724,229 @@ orchestrator::deploy_workflows_operator() {
...
+  local max_conn_attempts=12
+  local conn_ok=false
+
+  log::info "Verifying Backstage → data-index connectivity (via Service)..."
+  local node_probe
+  node_probe=$(
+    cat << 'PROBE'
+const http = require("http");
+const data = JSON.stringify({query:"{ ProcessDefinitions { id } }"});
+const req = http.request("http://sonataflow-platform-data-index-service/graphql",
```
gustavolira, line 892: nit: This hard-codes the data-index service DNS and assumes `node`/`http` exist in the `backstage-backend` container (same for the `/q/health` probe below). Fine today — a one-line comment noting the assumption would save a future debugging session if the base image or service names change.

.ci/pipelines/jobs/ocp-operator.sh
```diff
@@ -147,7 +143,7 @@ handle_ocp_operator() {
 
   cluster_setup_ocp_operator
 
-  prepare_operator
+  prepare_operator "3"
```
gustavolira: nit: The `"3"` here is a bare magic number — from the call site it's not obvious it's the operator-install retry count. Could we either name it (e.g. `local operator_install_retries=3`) or add a short comment? The other handlers call `prepare_operator` with no arg, so this difference is easy to miss.

.ci/pipelines/jobs/ocp-operator.sh
teknaS47: Restored the` ${NAME_SPACE:-...}` fallback for consistency

.ci/pipelines/lib/orchestrator.sh
teknaS47: Fixed, it now only includes databaseName when explicitly provided, same as before

teknaS47:
Fix already exists in the overlay repo, so closing this PR

PR #4217, opened 2026-02-09
fix: (e2e) debug and fix CI jobs

albarbaro:
## Description

- Fixes #?

## PR acceptance criteria

### DIFF
--- a/e2e-tests/local-test-setup.sh
+++ b/e2e-tests/local-test-setup.sh
@@ -120,18 +120,29 @@ if ! vault token lookup &>/dev/null; then
 fi
 
 log::info "Exporting secrets as environment variables..."
-# Export secrets safely without eval (avoids code injection risk)
-# Uses base64 encoding to safely handle special characters in values
-# Replaces -, . and / with _ in key names (env vars can only have alphanumeric and _)
-SECRETS_JSON=$(vault kv get -format=json -mount="kv" "selfservice/rhdh-qe/rhdh" | jq -r '.data.data')
-for key in $(echo "$SECRETS_JSON" | jq -r 'keys[]'); do
-    # Skip metadata keys
-    [[ "$key" == "secretsync/"* ]] && continue
-    # Get value and sanitize key name (put - at end for macOS tr compatibility)
-    value=$(echo "$SECRETS_JSON" | jq -r --arg k "$key" '.[$k]')
-    safe_key=$(echo "$key" | tr './-' '___')
-    export "$safe_key"="$value"
-done
+if [ -n "$ZSH_VERSION" ]; then
+    # --- ZSH / macOS FIX ---
+    # Fix: Vault flag order, printf to avoid control char interpretation, while loop for safety
+    SECRETS_JSON=$(vault kv get -mount="kv" -format=json "selfservice/rhdh-qe/rhdh" | jq -r '.data.data')
+    
+    while read -r key; do
+        [[ "$key" == "secretsync/"* ]] && continue
+        value=$(printf '%s' "$SECRETS_JSON" | jq -r --arg k "$key" '.[$k]')
+        safe_key=$(echo "$key" | tr './-' '___')
+        export "$safe_key"="$value"
+    done < <(printf '%s' "$SECRETS_JSON" | jq -r 'keys[]')
+
+elif [ -n "$BASH_VERSION" ]; then
+    # --- ORIGINAL BASH LOGIC ---
+    SECRETS_JSON=$(vault kv get -format=json -mount="kv" "selfservice/rhdh-qe/rhdh" | jq -r '.data.data')
+    
+    for key in $(echo "$SECRETS_JSON" | jq -r 'keys[]'); do
+        [[ "$key" == "secretsync/"* ]] && continue
+        value=$(echo "$SECRETS_JSON" | jq -r --arg k "$key" '.[$k]')
+        safe_key=$(echo "$key" | tr './-' '___')
+        export "$safe_key"="$value"
+    done
+fi
 
 log::section "Environment Ready"
 log::info "Available URLs:"
--- a/e2e-tests/local-test-setup.sh
+++ b/e2e-tests/local-test-setup.sh
@@ -120,29 +120,15 @@ if ! vault token lookup &>/dev/null; then
 fi
 
 log::info "Exporting secrets as environment variables..."
-if [ -n "$ZSH_VERSION" ]; then
-    # --- ZSH / macOS FIX ---
-    # Fix: Vault flag order, printf to avoid control char interpretation, while loop for safety
-    SECRETS_JSON=$(vault kv get -mount="kv" -format=json "selfservice/rhdh-qe/rhdh" | jq -r '.data.data')
-    
-    while read -r key; do
-        [[ "$key" == "secretsync/"* ]] && continue
-        value=$(printf '%s' "$SECRETS_JSON" | jq -r --arg k "$key" '.[$k]')
-        safe_key=$(echo "$key" | tr './-' '___')
-        export "$safe_key"="$value"
-    done < <(printf '%s' "$SECRETS_JSON" | jq -r 'keys[]')
+# --- ZSH FIX ---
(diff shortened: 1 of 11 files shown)


e2e-tests/playwright/e2e/plugins/adoption-insights/adoption-insights.spec.ts
gustavolira, line 63: let's try to avoid weak locator

e2e-tests/playwright.config.ts
```diff
@@ -84,7 +84,7 @@ export default defineConfig({
     },
     {
       name: PW_PROJECT.SHOWCASE,
-      dependencies: [PW_PROJECT.SMOKE_TEST],
+      //dependencies: [PW_PROJECT.SMOKE_TEST],
```
gustavolira: This removes the safety gate that prevents SHOWCASE tests from running when smoke tests fail. If this is just for debugging, make sure to revert before any merge — otherwise we lose the fail-fast behavior that catches broken environments early.

e2e-tests/playwright/utils/ui-helper.ts
```diff
@@ -1,6 +1,7 @@
 import { expect, Locator, Page } from "@playwright/test";
 import { UI_HELPER_ELEMENTS } from "../support/page-objects/global-obj";
 import { SEARCH_OBJECTS_COMPONENTS } from "../support/page-objects/page-obj";
+import { WAIT_OBJECTS } from "../support/page-objects/global-obj";
```
gustavolira: This import is from the same module as line 2 (`UI_HELPER_ELEMENTS`). Can we merge them into a single import?
```ts
import { UI_HELPER_ELEMENTS, WAIT_OBJECTS } from "../support/page-objects/global-obj";
```

e2e-tests/playwright/utils/ui-helper.ts
```diff
@@ -235,8 +236,18 @@ export class UIhelper {
       .click();
   }
 
+  async waitForLoad(timeout = 120000) {
```
gustavolira: This is an exact copy of `Common.waitForLoad()` in `common.ts:33-40`. Having the same logic in two places will drift over time. Since `Common` already holds a `UIhelper` instance, one option would be to keep `waitForLoad` only here in `UIhelper` and have `Common.waitForLoad()` delegate to `this.uiHelper.waitForLoad()`. Single source of truth.

e2e-tests/playwright/utils/ui-helper.ts
```diff
@@ -235,8 +236,18 @@ export class UIhelper {
...
+      await this.page.waitForSelector(item, {
+        state: "hidden",
+        timeout: timeout,
+      });
+    }
+  }
+
   async goToPageUrl(url: string, heading?: string) {
     await this.page.goto(url);
+    await this.waitForLoad()
```
gustavolira: Missing semicolon — this is likely what's causing the Prettier CI failure.

e2e-tests/local-test-setup.sh
```diff
@@ -120,18 +120,29 @@ if ! vault token lookup &>/dev/null; then
...
+    # --- ORIGINAL BASH LOGIC ---
+    SECRETS_JSON=$(vault kv get -format=json -mount="kv" "selfservice/rhdh-qe/rhdh" | jq -r '.data.data')
+    
+    for key in $(echo "$SECRETS_JSON" | jq -r 'keys[]'); do
+        [[ "$key" == "secretsync/"* ]] && continue
+        value=$(echo "$SECRETS_JSON" | jq -r --arg k "$key" '.[$k]')
+        safe_key=$(echo "$key" | tr './-' '___')
+        export "$safe_key"="$value"
+    done
+fi
```
gustavolira: There's no `else` fallback. If this runs under `sh`/`dash` (or any shell where neither `$ZSH_VERSION` nor `$BASH_VERSION` is set), the entire secrets export is silently skipped and tests will fail with confusing "variable not set" errors downstream.

Also worth noting: the shebang is `#!/bin/bash`, so under normal execution `$BASH_VERSION` is always set and the zsh branch is unreachable — it would only trigger if someone sources this file from a zsh session. Consider adding at least a warning:
```bash
else
    log::warn "Unsupported shell — secrets export may not work correctly"
fi
```
Also, the two blocks are nearly identical (only diff is vault flag order and `printf` vs `echo`). The `while read` + process substitution approach works in bash 4+ too, so we could unify them and avoid the duplication.

e2e-tests/playwright/support/pages/adoption-insights.ts
```diff
@@ -145,22 +145,53 @@ export class TestHelper {
     await newpage.close();
   }
 
+  private readonly defaultApiTimeoutMs = 20000;
```
gustavolira, line 148: Default timeout dropped from 60s to 20s. Are all existing callers of `waitUntilApiCallSucceeds` (those not yet migrated to `waitForApiCallAfterAction`) okay with this? If any path in CI genuinely takes >20s this will introduce new failures. Worth double-checking before merging.

e2e-tests/playwright/support/pages/adoption-insights.ts
```diff
@@ -145,22 +145,53 @@ export class TestHelper {
...
+   * miss the response and fail fast with a shorter timeout. Reduces flakiness and
+   * saves time vs waiting up to 60s after the action.
+   */
+  async waitForApiCallAfterAction(
+    page: Page,
+    action: () => Promise<void>,
+    urlPart: string = "/api/adoption-insights/events",
+    timeoutMs: number = 15000,
+  ): Promise<void> {
+    const [response] = await Promise.all([
```
gustavolira, line 184: Nice — the `Promise.all` pattern is the right way to avoid the race condition where the API response fires before `waitForResponse` is registered. Good improvement over the sequential approach.

e2e-tests/playwright.config.ts
```diff
@@ -40,7 +40,7 @@ const k8sSpecificConfig = {
 };
 
 export default defineConfig({
-  timeout: 90 * 1000,
+  timeout: 180 * 1000,
```
subhashkhileri: If we are increasing the timeouts, I’d suggest removing the retry completely (I’m in favor of that), or at least reducing it to 1. It’s currently set to 2, and that could make the overall execution time very long once tests start failing.

e2e-tests/playwright/e2e/plugins/adoption-insights/adoption-insights.spec.ts
albarbaro: This was not touched by the PR. I will open a separate Jira to have it fixed.

e2e-tests/playwright.config.ts
albarbaro: Already reverted, was for quick testing locally.

e2e-tests/playwright.config.ts
albarbaro: I would be in favour of removing the retries too, but looking at the [last test run](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-developer_rhdh/4217/pull-ci-redhat-developer-rhdh-main-e2e-ocp-helm/2021354631002591232/artifacts/e2e-ocp-helm/redhat-developer-rhdh-ocp-helm/artifacts/showcase/index.html#?q=s%3Aflaky), there are some test cases that are still relying on second attempts to pass. Those tests would probably need to be re-written with a proper retry logic before we remove the retries, otherwise they will make the job fail most probably.

e2e-tests/playwright/support/pages/adoption-insights.ts
albarbaro: Tests were exceeding the gloabl deadline becase some would wait for the whole 60 seconds before failing. The fix was run multiple times and did show any breakage in other tests.

e2e-tests/playwright/utils/ui-helper.ts
albarbaro: merged with the common.ts one.

e2e-tests/playwright.config.ts
albarbaro: Reduced to 120, let's see if it make it break again or not

PR #4469, opened 2026-03-26
feat(ci): parallelize deployments and readiness checks for showcase and showcase-rbac

gustavolira:
## Summary

- Run base and RBAC Helm deployments concurrently via background processes
- Overlap Backstage readiness waits (~5 min each) by running health checks in parallel, saving ~3-5 min
- Tests run sequentially (parallel Playwright caused OOMKilled — not enough memory for 6 concurrent browsers)
- Per-project output paths for test artifacts to avoid file conflicts between runs

## Design

| Phase | Execution | Why |
|---|---|---|
| Helm deployments | Parallel | Independent namespaces, no resource conflict |
| Readiness checks (HTTP polling) | Parallel | Lightweight, main time savings (~5 min) |
| Playwright tests | Sequential | Parallel caused OOMKilled (exit 137) — 2 Playwright x 3 workers exceeds CI memory |

## Changes

**Parallel deployments** (`utils.sh`):
- `initiate_deployments()` and `initiate_deployments_osd_gcp()` run base + RBAC concurrently with `&` + `wait`

**Parallel readiness + sequential tests** (`testing.sh`):
- New `testing::parallel_check_and_test()` pre-warms readiness checks in parallel, then calls `testing::check_and_test` sequentially for each deployment
- `testing::run_tests()` uses per-project output paths: `test-results-<subdir>/`, `junit-results-<subdir>.xml`, `playwright-report-<subdir>/`, per-project log files
- `testing::check_backstage_running()` uses per-project log file path to avoid race conditions

**Pod log collection** (`utils.sh`):
- `save_all_pod_logs()` passes pod logs directory as function parameter (not exported global)
- Uses unique `/tmp/pod_logs-<subdir>/` per artifacts subdir

**Callers** (`ocp-pull.sh`, `ocp-nightly.sh`):
- Both use `testing::parallel_check_and_test()`

## Test plan

- [ ] Verify OCP PR job runs deployments and readiness checks in parallel
- [ ] Verify Playwright tests run sequentially without OOMKilled
- [ ] Verify nightly subsequent test phases (runtime, sanity, localization) work correctly
- [ ] Verify OSD-GCP nightly deploys in parallel
- [ ] Verify artifacts match original layout (index.html, data/, trace/ at showcase/ root)
- [ ] Verify proper error reporting when one deployment fails

Resolves: [RHIDP-12296](https://issues.redhat.com/browse/RHIDP-12296)


.ci/pipelines/lib/testing.sh
```diff
@@ -110,19 +127,93 @@ testing::run_tests() {
...
+  local rbac_namespace=$6
+  local rbac_project=$7
+  local rbac_url=$8
+  local base_artifacts="${base_project}"
+  local rbac_artifacts="${rbac_project}"
+
+  # Phase 1: Wait for both deployments to be ready in parallel (lightweight HTTP polling)
+  log::section "Waiting for both deployments to be ready in parallel"
+
+  testing::check_backstage_running "${base_release}" "${base_namespace}" "${base_url}" "${base_artifacts}" &
```
subhashkhileri: **Bug: `$LOGFILE` race condition when running in background**

Both background readiness checks will use the shared global `$LOGFILE` (`test-log`). Inside `check_backstage_running()` (lines 279, 289), on crash-detection or timeout it writes:
```bash
common::save_artifact "${artifacts_subdir}" "/tmp/${LOGFILE}" || true
```

Since both background processes share the same `$LOGFILE`, they'll clobber each other's log if both hit the failure/timeout path.

`run_tests()` was correctly updated to use per-project logfiles — but `check_backstage_running()` was missed.

Suggest updating those lines to use a per-project path:
```bash
common::save_artifact "${artifacts_subdir}" "/tmp/${LOGFILE}-${artifacts_subdir}" || true
```

.ci/pipelines/utils.sh
```diff
@@ -618,13 +623,61 @@ rbac_deployment() {
...
+#   $1 - label: Human-readable label for logging (e.g. "deployments")
+#   $2 - base_func: Name of the base function to run
+#   $3 - base_arg: Argument for the base function
+#   $4 - rbac_func: Name of the RBAC function to run
+#   $5 - rbac_arg: Argument for the RBAC function
+# Returns: 0 if both succeed, 1 if either fails
+_run_parallel() {
+  local label=$1
+  local base_func=$2
+  local base_arg=$3
```
subhashkhileri: **Suggestion: Inline instead of abstracting**

This helper is called exactly twice (`initiate_deployments` and `initiate_deployments_osd_gcp`) and only supports single-argument functions. The generic naming/signature suggests reusability, but it's tightly constrained. Consider inlining the `&`/`wait` pattern directly in the two callers — it's ~6 lines and immediately readable:

```bash
initiate_deployments() {
  local base_artifacts_subdir=$1
  local rbac_artifacts_subdir=$2
  cd "${DIR}"
  base_deployment "${base_artifacts_subdir}" &
  local base_pid=$!
  rbac_deployment "${rbac_artifacts_subdir}" &
  local rbac_pid=$!
  local failed=0
  wait "${base_pid}" || { log::error "Base deployment failed"; failed=1; }
  wait "${rbac_pid}" || { log::error "RBAC deployment failed"; failed=1; }
  return "${failed}"
}
```

No indirection, no need to look up what `_run_parallel` does.

.ci/pipelines/lib/testing.sh
```diff
@@ -110,19 +127,93 @@ testing::run_tests() {
...
+  _run_check_and_test_phase "${base_release}" "${base_namespace}" "${base_project}" "${base_url}" "${base_artifacts}" "${base_ready}"
+  _run_check_and_test_phase "${rbac_release}" "${rbac_namespace}" "${rbac_project}" "${rbac_url}" "${rbac_artifacts}" "${rbac_ready}"
+  return 0
+}
+
+# Run test or mark failure for a single deployment after readiness check.
+# Args:
+#   $1-$5: release_name, namespace, playwright_project, url, artifacts_subdir
+#   $6: readiness check exit code (0 = ready)
+_run_check_and_test_phase() {
```
subhashkhileri: **Duplication with `testing::check_and_test()`**

This function duplicates the logic already in `testing::check_and_test()` (lines 307-341) — same pod display, same `SKIP_TESTS` check, same `run_tests`/`save_all_pod_logs` branching, same `mark_deploy_failed` on failure.

Consider reusing `check_and_test` instead. After the parallel readiness wait completes, `check_backstage_running` inside `check_and_test` will return immediately on the first curl (HTTP 200 in ~2 seconds since the deployment is already up). The re-check cost is negligible, and you avoid maintaining two copies of the same logic:

```bash
# In ocp-pull.sh — parallel readiness, then reuse existing check_and_test
testing::check_backstage_running "..." "..." "${url}" "..." &
local base_pid=$!
testing::check_backstage_running "..." "..." "${rbac_url}" "..." &
local rbac_pid=$!
wait "${base_pid}"; wait "${rbac_pid}"

testing::check_and_test "${RELEASE_NAME}" "${NAME_SPACE}" "${PW_PROJECT_SHOWCASE}" "${url}"
testing::check_and_test "${RELEASE_NAME_RBAC}" "${NAME_SPACE_RBAC}" "${PW_PROJECT_SHOWCASE_RBAC}" "${rbac_url}"
```

No new functions needed.

.ci/pipelines/utils.sh
```diff
@@ -133,11 +133,15 @@ save_all_pod_logs() {
   set +e
   local namespace=$1
   local artifacts_subdir="${2:-$namespace}"
-  rm -rf pod_logs && mkdir -p pod_logs
+
+  # Use a unique temp directory per artifacts_subdir to allow parallel pod log collection
+  export _POD_LOGS_DIR="/tmp/pod_logs-${artifacts_subdir}"
```
subhashkhileri: **Code smell: exported global to pass a directory path**

Using `export _POD_LOGS_DIR` + `unset` to communicate with `retrieve_pod_logs` is fragile. If `save_all_pod_logs` is ever called concurrently (which this PR is moving toward), the global gets clobbered between calls.

Consider passing it as a function parameter instead:
```bash
retrieve_pod_logs "$pod_name" "$container" "$namespace" "$pod_logs_dir"
```

This eliminates the global state, the `unset` cleanup, and the `${_POD_LOGS_DIR:-pod_logs}` fallback pattern in `retrieve_pod_logs`.

.ci/pipelines/lib/testing.sh
```diff
@@ -110,19 +127,93 @@ testing::run_tests() {
...
   return "$test_result"
 }
 
+# Run readiness checks in parallel, then execute tests sequentially.
+# Parallelizing the readiness waits (~5 min each) saves significant time while
+# running tests sequentially avoids OOM from concurrent Playwright instances.
+# Args: pairs of (release_name namespace playwright_project url) for base and RBAC
+#   $1-$4: base args
+#   $5-$8: RBAC args
+testing::parallel_check_and_test() {
```
subhashkhileri, line 151: **Nit: Hardcoded for exactly 2 deployments**

This function takes 8 positional args (4 per deployment) and is hardcoded to the base+RBAC pair. This is fine for now, but a brief comment like `# Hardcoded for the base + RBAC deployment pair` would clarify that the rigid 8-arg signature is intentional and not meant to be extended to N deployments.

subhashkhileri (commented):
Reviewed the parallelization changes. The overall design is sound — parallel readiness checks are a good win, and per-project artifact paths fix a real overwrite bug. A few suggestions below.

PR #4491, opened 2026-03-27
feat: add support for concurrent plugin installs

thepetk:
## Description

Tries to allow concurrency on the plugin installation phase. The dynamic plugin installer so far allowed only sequential installation.

According to my benchmarks it can reduce 2x or 3x the time needed to install the plugins for your RHDH instance. 

Key changes:

- Plugin installation now runs in a `ThreadPoolExecutor` which defaults to 5 workers, but that's configurable via `MAX_CONCURRENT_INSTALLS`.
- The OCI manifest inspection process is now pre-fetched and outside of the merge loop, as the merge loop has to be sequential. That allows us to run the `skopeo inspect` in parallel and use cached values in the sequential loop.
- For the above we use a shared `_oci_path_cache` dict avoids redundant inspects across plugins from the same image.
- A `threading.Lock` guards all concurrent writes to `plugin_path_by_hash`.
- As the previous logging flow would not work with concurrent installs I've added `PluginStatusDisplay` just to help the user read more easy what is happening.

- Fixes [RHIDP-3304](https://redhat.atlassian.net/browse/RHIDP-3304)

## PR acceptance criteria

I've tested this both locally but also on a custom image I've built. I'll need for sure some guidance if futhre testing is required from my side.

Note, that intentionally I didn't want to do a lot of structural changes, like removing the skopeo dependency cause I feel this set of changes is more straightforward and with less risk involved. Looking at the benchmarks I feel it worths the effort.

For sure according to my experiments, removing the skopeo dependency will make this process even faster, mostly because the skopeo copy needs to fetch everything while we're ok downloading only a part of each image.


thepetk:
> Image was [built](https://github.com/redhat-developer/rhdh/actions/runs/23651909761) and [published](https://github.com/redhat-developer/rhdh/actions/runs/23653593126) successfully. It is available at:
> 
> * [`quay.io/rhdh-community/rhdh:pr-4491`](https://quay.io/rhdh-community/rhdh:pr-4491)
> * [`quay.io/rhdh-community/rhdh:pr-4491-2de00eb5`](https://quay.io/rhdh-community/rhdh:pr-4491-2de00eb5)

FYI I've tried to run a test instance of the `ai-rolling-demo-gitops` using the `quay.io/rhdh-community/rhdh:pr-4491` image. The comparison between this one and `quay.io/rhdh/rhdh-hub-rhel9/rhdh:1.10-87` can be seen below. The suggested approach took 1.19 mins to end while the current one finished after 3.25 mins.

```
Init Containers:
  install-dynamic-plugins:
    Container ID: 
...
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Mon, 30 Mar 2026 11:10:30 +0100
      Finished:     Mon, 30 Mar 2026 11:11:49 +0100

Init Containers:
  install-dynamic-plugins:
    Container ID: 
...
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Mon, 30 Mar 2026 11:39:02 +0100
      Finished:     Mon, 30 Mar 2026 11:42:27 +0100
```

thepetk:
Adding a comment to keep it alive, just in case the team still wants to pursue this optimization :)

rm3l (commented):
@thepetk Looks like this is being superseded by https://github.com/redhat-developer/rhdh/pull/4574, which is rewriting the Python script to TypeScript with similar support for concurrent installs among other things. PTAL at that PR to compare the improvements in your environment.

PR #4574, opened 2026-04-13
feat(install-dynamic-plugins): port from Python to TypeScript/Node.js

gustavolira:
## Summary

Replaces the Python-based `install-dynamic-plugins.py` init-container script with a TypeScript/Node.js implementation, incorporating all improvements from the `install-dynamic-plugins-fast.py` POC (#4523): parallel OCI downloads, shared image cache, streaming SHA verification.

**Motivation**: The Python script was a longstanding duplication target for export overlays (see [rhdh-plugin-export-overlays#2231](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2231)). A Node.js implementation removes the Python dependency from the init-container runtime, enables reuse across RHDH core and overlays, and adopts the faster parallel architecture natively.

## What changed

- **New package** `scripts/install-dynamic-plugins/` — 18 TypeScript modules + 9 Jest test files (105 tests)
- **Bundled output**: single `dist/install-dynamic-plugins.cjs` (~412 KB), committed and verified fresh in CI (same pattern as `.yarn/releases/yarn-*.cjs`)
- **Containerfile** now copies the `.cjs` bundle instead of `.py`; wrapper shell script execs `node` instead of `python`
- **CI workflow** runs `npm run tsc && npm test` + bundle freshness check (replaces `pytest`)
- **Deleted**: `install-dynamic-plugins.py` (1288 LOC), `test_install-dynamic-plugins.py` (3065 LOC), `pytest.ini`

### Runtime contract is unchanged
Same `dynamic-plugins.yaml` input schema, same `app-config.dynamic-plugins.yaml` output, same `dynamic-plugin-config.hash` / `dynamic-plugin-image.hash` files, same lock-file behaviour, same `{{inherit}}` semantics and OCI path auto-detection.

### Resource-conscious concurrency
- `availableParallelism()` respects cgroup CPU limits (init containers often get 0.5 CPU)
- Default workers: `max(1, min(floor(cpus/2), 6))` — cap avoids exhausting registry/network
- NPM installs stay sequential (npm registry throttles parallel fetches)
- Override via `DYNAMIC_PLUGINS_WORKERS=<n>`

### Memory: streaming everywhere
- `node-tar` streams extraction — no full-archive read into RAM
- `node:crypto` pipeline for SHA integrity — chunks through the hash
- Typical 10-plugin run: 20–80 MB peak RSS (well below 512 Mi init-container limit)

### Security parity with Python
| Check | File |
|---|---|
| Path traversal in plugin path (`..`, absolute) | `tar-extract.ts` |
| Per-entry size cap (zip bomb, `MAX_ENTRY_SIZE`) | `tar-extract.ts`, `catalog-index.ts` |
| Sym/hardlink target must stay inside destination | `tar-extract.ts` |
| Reject device files / FIFOs / unknown entry types | `tar-extract.ts` |
| `package/` prefix enforced for NPM tarballs | `tar-extract.ts` |
| SRI integrity (`sha256` / `sha384` / `sha512`) | `integrity.ts` |
| Registry fallback `registry.access.redhat.com/rhdh` → `quay.io/rhdh` | `image-resolver.ts` |

## Test plan

- [x] `npm run tsc` passes (strict mode, `noUncheckedIndexedAccess`)
- [x] `npm test` — 105 Jest tests pass (npm-key, oci-key, integrity, tar-extract, merger, concurrency, lock-file, image-resolver, plugin-hash)
- [x] `npm run build` produces fresh `dist/install-dynamic-plugins.cjs`
- [ ] Container image builds successfully with new `.cjs` (CI will verify)
- [ ] Init-container flow validated on cluster (end-to-end catalog install + RHDH startup)
- [ ] Resource profile check: wall-clock ≤ `fast.py` baseline (~2:42 for full catalog)

## Compatibility

- Python 3.11 is **still installed** in the runtime image for techdocs/mkdocs
- `skopeo` is **still installed** for OCI inspection
- No new system dependencies required (Node.js 22 already runs the Backstage backend)
- `install-dynamic-plugins.sh` wrapper contract unchanged (`./install-dynamic-plugins.sh /dynamic-plugins-root`)

## Related

- Replaces the Python script referenced in [rhdh-plugin-export-overlays#2231](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2231)
- Incorporates improvements from #4523 (parallel downloads, shared image cache)


rostalan:
my browser cannot even load the 11,5k-line file 😅

gustavolira:
> my browser cannot even load the 11,5k-line file 😅

@rostalan No need to review that file, it's the auto-generated esbuild bundle of src/. Just pushed a commit marking it as linguist-generated so GitHub collapses it in the diff. Only src/ and __tests__/ need review; CI verifies the bundle is up-to-date on every push.

kadel:
One thing I noticed in the review — the error handling behavior changed from the Python version.

In Python, when a plugin fails to install, the script stops immediately. The config file is never written, nothing is cleaned up. It's messy but clearly "not finished".

In the new TS version, when a plugin fails, the script continues with the rest. It writes `app-config.dynamic-plugins.yaml` with configs from the successful plugins, runs cleanup, and then exits with code 1.

The exit code 1 prevents the main container from starting, so in normal Kubernetes/Compose flow this is fine. But the problem is that you end up with a state that looks "almost valid" — most plugins installed, config file written, cleanup done. If someone troubleshooting manually restarts just the main container (skipping the init), or if the deployment doesn't enforce init container success, they could get a partially working instance with missing plugins.

Also plugin dependencies could be an issue — a frontend plugin might be installed and configured while its backend dependency failed. The app would start but with broken UI.

I think the idea of collecting all errors and showing them at once is great (much better than fix-one-restart-fix-one-restart with 50 plugins). But maybe don't write the config file when there are errors? That way you get the diagnostic benefit without leaving a valid-looking state on disk.

gustavolira:
> One thing I noticed in the review — the error handling behavior changed from the Python version.
> 
> In Python, when a plugin fails to install, the script stops immediately. The config file is never written, nothing is cleaned up. It's messy but clearly "not finished".
> 
> In the new TS version, when a plugin fails, the script continues with the rest. It writes `app-config.dynamic-plugins.yaml` with configs from the successful plugins, runs cleanup, and then exits with code 1.
> 
> The exit code 1 prevents the main container from starting, so in normal Kubernetes/Compose flow this is fine. But the problem is that you end up with a state that looks "almost valid" — most plugins installed, config file written, cleanup done. If someone troubleshooting manually restarts just the main container (skipping the init), or if the deployment doesn't enforce init container success, they could get a partially working instance with missing plugins.
> 
> Also plugin dependencies could be an issue — a frontend plugin might be installed and configured while its backend dependency failed. The app would start but with broken UI.
> 
> I think the idea of collecting all errors and showing them at once is great (much better than fix-one-restart-fix-one-restart with 50 plugins). But maybe don't write the config file when there are errors? That way you get the diagnostic benefit without leaving a valid-looking state on disk.

@kadel tyvm for your review! Just pushed a fix that on any plugin failure we now skip both the `app-config.dynamic-plugins.yaml` write and `cleanupRemoved()`, so the prior-run state is preserved instead of an "almost valid" one. Still collect all errors into one summary so we keep the UX win over Python's fail-fast.

Skipped cleanup too (not just the write) because otherwise the previous run's config would remain on disk referencing plugin dirs we just deleted, worse than the original bug.

Added unit tests for the success and failure paths, and verified end-to-end with a bogus OCI ref: exit 1, no config written, prior sentinel preserved.

.github/workflows/pr.yaml
```diff
@@ -148,9 +148,23 @@ jobs:
         if: ${{ steps.check-image.outputs.is_skipped != 'true' }}
         run: yarn run test --continue --affected
 
-      - name: Run Python tests
+      - name: Test install-dynamic-plugins (TypeScript)
```
jonkoops: I think mentioning the technology involved doesn't really provide value here.
```suggestion
      - name: Test install-dynamic-plugins
```

.github/workflows/pr.yaml
```diff
@@ -148,9 +148,23 @@ jobs:
...
         if: ${{ steps.check-image.outputs.is_skipped != 'true' }}
-        run: pytest scripts/install-dynamic-plugins/test_install-dynamic-plugins.py -v
+        working-directory: scripts/install-dynamic-plugins
+        run: |
+          npm install --no-audit --no-fund
```
jonkoops: NPM has a dedicated command for installations on CI, see [`npm ci`](https://docs.npmjs.com/cli/v11/commands/npm-ci).

```suggestion
          npm ci
```

gashcrumb:
I took a moment to try this out locally by grabbing the dynamic-plugins.default.yaml from the latest catalog image and used this dynamic-plugins.yaml:

```
includes:
  - dynamic-plugins.default.yaml
plugins: []
```

Script worked fine but it didn't appear to me like it was much faster than the python script.  So I opted to time it's execution as compared to the python script, it seems to take about the same amount of time, at least in my environment:

```
/usr/bin/time -f 'elapsed %e s (user %U, sys %S)' node scripts/install-dynamic-plugins/dist/install-dynamic-plugins.cjs dynamic-plugins-root
elapsed 68.06 s (user 59.52, sys 9.47)

/usr/bin/time -f 'elapsed %e s (user %U, sys %S)' python scripts/install-dynamic-plugins/install-dynamic-plugins.py dynamic-plugins-root
elapsed 62.30 s (user 53.11, sys 6.12)
```

I suppose most of this is spent doing `npm unpack` as several of our defaults still point to wrappers.

gashcrumb:
One other observation, flipping between scripts reveals that the two scripts use different methods to evaluate differences in existing dynamic-plugins-root content.  I figured I'd do an additional test where I'd run each script with an already populated dynamic-plugins-root.  For both the python script and this javascript version I actually had to run each script a couple of times before getting to a point where the scripts would not download and reinstall an existing plugin.  I grabbed the execution time for this case once the contents of the dynamic-plugins-root folder settled:

```
/usr/bin/time -f 'elapsed %e s (user %U, sys %S)' node scripts/install-dynamic-plugins/dist/install-dynamic-plugins.cjs dynamic-plugins-root
elapsed 17.77 s (user 4.38, sys 1.64)

/usr/bin/time -f 'elapsed %e s (user %U, sys %S)' python scripts/install-dynamic-plugins/install-dynamic-plugins.py dynamic-plugins-root
elapsed 13.72 s (user 2.38, sys 0.85)
```

gustavolira:
> I took a moment to try this out locally by grabbing the dynamic-plugins.default.yaml from the latest catalog image and used this dynamic-plugins.yaml:
> 
> ```
> includes:
>   - dynamic-plugins.default.yaml
> plugins: []
> ```
> 
> Script worked fine but it didn't appear to me like it was much faster than the python script. So I opted to time it's execution as compared to the python script, it seems to take about the same amount of time, at least in my environment:
> 
> ```
> /usr/bin/time -f 'elapsed %e s (user %U, sys %S)' node scripts/install-dynamic-plugins/dist/install-dynamic-plugins.cjs dynamic-plugins-root
> elapsed 68.06 s (user 59.52, sys 9.47)
> 
> /usr/bin/time -f 'elapsed %e s (user %U, sys %S)' python scripts/install-dynamic-plugins/install-dynamic-plugins.py dynamic-plugins-root
> elapsed 62.30 s (user 53.11, sys 6.12)
> ```
> 
> I suppose most of this is spent doing `npm unpack` as several of our defaults still point to wrappers.

My test was using sanity-plugin check which loads every plugin that we are shipping

.github/workflows/pr.yaml
```diff
@@ -148,9 +148,23 @@ jobs:
...
+          npm run tsc
+          npm test
+
+      - name: Verify install-dynamic-plugins bundle is up to date
+        if: ${{ steps.check-image.outputs.is_skipped != 'true' }}
+        working-directory: scripts/install-dynamic-plugins
+        run: |
```
jonkoops, line 156: I see there is a build step to bundle all of this up into a single file. I assume that this is to reduce the size of the final image?

Would it not be acceptable to `npm install --production` to only install the production dependencies and then bring in the source + node_modules instead?

Note that Node.js also supports [single executable applications](https://nodejs.org/api/single-executable-applications.html), perhaps that would be fit for purpose here?

scripts/install-dynamic-plugins/esbuild.config.mjs
```diff
@@ -0,0 +1,15 @@
+import { build } from "esbuild";
+
+await build({
+  entryPoints: ["src/index.ts"],
+  bundle: true,
+  platform: "node",
+  target: "node22",
+  format: "cjs",
+  outfile: "dist/install-dynamic-plugins.cjs",
```
jonkoops: I would argue that we don't have to use CommonJS here. Node.js natively supports ESM these days, so that should be the target. There is also no need to add a `.mjs` extension, as it can detect the module type from source.

scripts/install-dynamic-plugins/.prettierrc.js
jonkoops, line 1: Would relying on the default behavior of Prettier not be preferred? Having an opinionated formatter saves us from having to think about this in the first place.

scripts/install-dynamic-plugins/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@rhdh/install-dynamic-plugins",
+  "version": "1.0.0",
+  "private": true,
+  "license": "Apache-2.0",
+  "description": "Install dynamic plugins into a RHDH instance from a dynamic-plugins.yaml config (OCI, NPM, local).",
+  "engines": {
+    "node": ">=22"
+  },
+  "main": "dist/install-dynamic-plugins.cjs",
```
jonkoops: There isn't really a `main` entrypoint, as this is not a library that actually ships with an importable programmatic interface (or at all). Using `bin` would be more appropriate if the purpose is to execute a command from the CLI.

scripts/install-dynamic-plugins/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "devDependencies": {
+    "@types/jest": "29.5.12",
+    "@types/node": "22.10.0",
+    "@types/tar": "6.1.13",
+    "esbuild": "0.25.0",
+    "jest": "29.7.0",
+    "prettier": "3.3.3",
+    "ts-jest": "29.4.4",
+    "typescript": "5.9.3"
+  }
```
jonkoops, line 33: Let's not pin dependencies to fixed version, as that might cause more dependencies to be pulled in for circular transitive dependencies with other ranges. Using a caret (`^`) version specifier would be the go-to here, using the lockfile to make things reproducible.

scripts/install-dynamic-plugins/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "dependencies": {
+    "tar": "7.5.13",
+    "yaml": "2.8.2"
+  },
+  "devDependencies": {
+    "@types/jest": "29.5.12",
+    "@types/node": "22.10.0",
+    "@types/tar": "6.1.13",
+    "esbuild": "0.25.0",
+    "jest": "29.7.0",
```
jonkoops: From a tooling perspective I would recommend using Vitest, as it works natively with TypeScript and ESM + mocking, and has a much more limited dependency tree + sibling dependencies (e.g. `@types/jest`, `ts-jest`) that have to be brought in.

scripts/install-dynamic-plugins/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "node": ">=22"
+  },
+  "main": "dist/install-dynamic-plugins.cjs",
+  "scripts": {
+    "build": "node esbuild.config.mjs",
+    "tsc": "tsc --noEmit",
+    "test": "jest",
+    "test:coverage": "jest --coverage",
+    "prettier:check": "prettier --check src __tests__",
+    "prettier:fix": "prettier --write src __tests__"
```
jonkoops: Doing a Prettier run over the whole directory is not an option?
```suggestion
    "prettier:check": "prettier --check src .",
    "prettier:fix": "prettier --write src ."
```

In general, I am wondering if this is not duplicated effort, ideally we'd have one single root dependency in the entire project that juggles the linting and formatting.

scripts/install-dynamic-plugins/tsconfig.json
jonkoops, line 1: Note that a lot of the config options provided here already have sensible defaults, especially in TypeScript 6. Might be worth it to just remove the options that already have the same or similar defaults.

jonkoops:
FYI I am doing my review in some batches as there is rather a lot to dig into here.

gustavolira:
> One other observation, flipping between scripts reveals that the two scripts use different methods to evaluate differences in existing dynamic-plugins-root content. I figured I'd do an additional test where I'd run each script with an already populated dynamic-plugins-root. For both the python script and this javascript version I actually had to run each script a couple of times before getting to a point where the scripts would not download and reinstall an existing plugin. I grabbed the execution time for this case once the contents of the dynamic-plugins-root folder settled:
> 
> ```
> /usr/bin/time -f 'elapsed %e s (user %U, sys %S)' node scripts/install-dynamic-plugins/dist/install-dynamic-plugins.cjs dynamic-plugins-root
> elapsed 17.77 s (user 4.38, sys 1.64)
> 
> /usr/bin/time -f 'elapsed %e s (user %U, sys %S)' python scripts/install-dynamic-plugins/install-dynamic-plugins.py dynamic-plugins-root
> elapsed 13.72 s (user 2.38, sys 0.85)
> ```

@gashcrumb  Thank you very much for the benchmark and explanation, very usefull.
Based on that I just pushed a new commit with some improvements.

New numbers from a local bench (30 OCI plugins, mock skopeo, populated dynamic-plugins-root, 5 runs averaged)

| Scenario                                                        | TS (wall / user / sys) | Python (wall / user / sys) | TS vs Python |
  |-----------------------------------------------------------------|------------------------|----------------------------|--------------|
  | Case 1 — 30 plugins, IfNotPresent                               | 0.08s / 0.05s / 0.01s  | 0.10s / 0.06s / 0.02s      | ~25% faster  |
  | Case 2 — 30 plugins, Always (different images)                  | 0.21s / 0.19s / 0.22s  | 0.76s / 0.21s / 0.33s      | ~3.6× faster |
  | Case 3 — 30 plugins, Always (same image — RHDH catalog pattern) | 0.10s / 0.06s / 0.03s  | 0.71s / 0.20s / 0.31s      | ~7× faster   |

The changes was basically
- Aligned the install hash with Python's format so cross-installs no longer trigger reinstalls (likely the cause of "had to run a couple of times before settling")
- Cached skopeo.exists() results so we don't fork skopeo once per plugin from the same image
- Added a synchronous pre-pass for definitely-no-op plugins so they bypass the worker pool entirely

.github/workflows/pr.yaml
gustavolira: The bundle is forced by the hermetic Konflux build — `cachi2` disables network during the image build, so `npm install` does not run on the builder. A pre-built single `.cjs` is the simplest way around it without adding another cachi2 prefetch config.

SEAs would be nice long-term but they are still stability tier 1 in Node 22 and need `postject` + a custom node binary.

The `.yarn/releases/yarn-*.cjs` pattern in this repo does the same thing for the same reason.

scripts/install-dynamic-plugins/esbuild.config.mjs
gustavolira: Kept CJS deliberately. Two reasons:

1. Init container startup on every pod restart, often on 0.5 CPU. CJS is measurably faster to cold-start than ESM for a single-file bundle.
2. `.cjs` matches what is already in the repo (`.yarn/releases/yarn-*.cjs`). Switching to `.mjs` means renames across the Containerfile, the wrapper script, `.gitignore`, `.dockerignore`, `.gitattributes`, and the CI freshness check.

The source is ESM-style already, esbuild just emits a CJS wrapper.

scripts/install-dynamic-plugins/tsconfig.json
gustavolira: updated

gustavolira:
@jonkoops tyvm for the review, addressed pretty much all of them, left inline comments on the two I didn't apply

jonkoops (approved):

gustavolira:
/review
--pr_reviewer.inline_code_comments=true
-i

jonkoops (approved):

PR #4576, opened 2026-04-13
chore: fix oci disable bug in install-dynamic-plugins.py

Zaperex:
## Description

Added a pre-merge step to disable annotation fetch for metadata for disabled plugins without an explicitly defined plugin path

- Fixes [RHDHBUGS-2533](https://redhat.atlassian.net/browse/RHDHBUGS-2533)

## PR acceptance criteria


scripts/install-dynamic-plugins/install-dynamic-plugins.py
```diff
@@ -1135,6 +1135,116 @@ def extract_catalog_index(catalog_index_image: str, catalog_index_mount: str, ca
...
+    # Track the source file of each OCI plugin entry with a defined path
+    defined_paths_per_plugin_registry = {}  # {registry: {path: source_file}}
+
+    def process_entry(plugin, level, source_file):
+        package = plugin.get('package', '')
+        if not isinstance(package, str) or not package.startswith(OCI_PROTOCOL_PREFIX):
+            return
+        match = re.match(OciPackageMerger.EXPECTED_OCI_PATTERN, package)
+        if not match:
+            raise InstallException(f"oci package \'{package}\' is not in the expected format \'{OCI_PROTOCOL_PREFIX}<registry>:<tag>\' or \'{OCI_PROTOCOL_PREFIX}<registry>@<algo>:<digest>\' (optionally followed by \'!<path>\') in {source_file} where <registry> may include a port (e.g. host:5000/path) and <algo> is one of {RECOGNIZED_ALGORITHMS}")
```
rm3l, line 1172: Wondering why we need to validate this.. IMO we should just skip any kind of validation for a package ref when `disabled` is `true`, no? Users may purposely want to disable the plugin definition completely (rather than commenting it out), even if the definition is invalid.

Zaperex: Hmm I guess that makes sense, but this is more for the general case for matching with entries that already exist in an `includes` file, which means we would need a consistent format for the package field to be able to properly match plugins to merge their `disabled` states.

rm3l: I see. How about skipping invalid entries that are `disabled` with just a warning? That way, if someone has an invalid disabled entry in their dynamic-plugins config but a valid enabled entry in an `includes` file, the valid one won't be disabled (as I guess there would be no matching here, but this is fine IMO). The warning on the invalid one would help users know something might be misconfigured. Would that make sense?
I was commenting on this specific package format validation error, but noticed there is also other `InstallException` raised in https://github.com/Zaperex/rhdh/blob/3178310b2e593c0b5ab187711dcc57c91626933a/scripts/install-dynamic-plugins/install-dynamic-plugins.py#L1204-L1209
These validations are all fine when the package is enabled, but should not block the init container on disabled plugins IMO.

Zaperex: Sure that makes sense

scripts/install-dynamic-plugins/install-dynamic-plugins.py
```diff
@@ -1135,6 +1135,133 @@ def extract_catalog_index(catalog_index_image: str, catalog_index_mount: str, ca
...
+    Non-OCI entries are passed through unchanged.
+    """
+    filtered = []
+    for plugin in plugins:
+        package = plugin.get('package', '')
+        if isinstance(package, str) and package.startswith(OCI_PROTOCOL_PREFIX):
+            match = re.match(OciPackageMerger.EXPECTED_OCI_PATTERN, package)
+            if match and match.group(1) in disabled_plugin_registries:
+                print(f'\n======= Disabling OCI plugin {package}', flush=True)
+                continue
```
rm3l, line 1260:
```suggestion
            if match and match.group(1) in disabled_plugin_registries:
                print(f'\n======= Disabling OCI plugin {package}', flush=True)
                continue
            if not match and plugin.get('disabled', False):
                print(f'\n======= Disabling invalid OCI plugin {package}', flush=True)
                continue
```

Tested the current changes, and when the regex doesn't match, the invalid plugin is not filtered out even if it is marked as disabled, and ends up crashing the init container.

<details>

<summary>Logs</summary>

```
======= Created lock file: /dynamic-plugins-root/install-dynamic-plugins.lock

======= Extracting catalog index from quay.io/rhdh/plugin-catalog-index:1.9
        ==> Copying catalog index image to local filesystem
        ==> Extracting catalog index layers
        ==> Extracting layer a9837d06aa6340c55fe90c8ba4f4310ec114d362a3b93da55a7ff75ac0d2d1b1
        ==> Successfully extracted dynamic-plugins.default.yaml from catalog index image
        ==> Extracting extensions catalog entities to /extensions
        ==> Successfully extracted extensions catalog entities from index image

======= Replacing dynamic-plugins.default.yaml with catalog index: /dynamic-plugins-root/.catalog-index-temp/dynamic-plugins.default.yaml

======= Including dynamic plugins from /dynamic-plugins-root/.catalog-index-temp/dynamic-plugins.default.yaml
WARNING: Skipping disabled OCI plugin with invalid format: 'oci://reg.example.com:fake_port/myplugin!my-plugin' in dynamic-plugins.yaml. Expected format: 'oci://<registry>:<tag>' or 'oci://<r
egistry>@<algo>:<digest>' (optionally followed by '!<path>') where <registry> may include a port (e.g. host:5000/path) and <algo> is one of ('sha512', 'sha384', 'sha256')

======= Disabling OCI plugin oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-community-plugin-3scale-backend:bs_1.45.3__3.10.0

[...]

======= Adding new dynamic plugin configuration for ./dynamic-plugins/dist/red-hat-developer-hub-backstage-plugin-bulk-import-backend-dynamic

======= Adding new dynamic plugin configuration for ./dynamic-plugins/dist/roadiehq-scaffolder-backend-module-http-request-dynamic
Traceback (most recent call last):
  File "/opt/app-root/src/install-dynamic-plugins.py", line 1430, in <module>
    main()
  File "/opt/app-root/src/install-dynamic-plugins.py", line 1384, in main
    merge_plugin(plugin, all_plugins, dynamic_plugins_file, level=1)
  File "/opt/app-root/src/install-dynamic-plugins.py", line 140, in merge_plugin
    return OciPackageMerger(plugin, dynamic_plugins_file, all_plugins).merge_plugin(level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/src/install-dynamic-plugins.py", line 589, in merge_plugin
    plugin_key, version, inherit_version, resolved_path = self.parse_plugin_key(package)
                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/src/install-dynamic-plugins.py", line 508, in parse_plugin_key
    raise InstallException(f"oci package \'{package}\' is not in the expected format \'{OCI_PROTOCOL_PREFIX}<registry>:<tag>\' or \'{OCI_PROTOCOL_PREFIX}<registry>@<algo>:<digest>\' (optionally followed by \'!<path>\') in {self.dynamic_plugins_file} where <registry> may include a port (e.g. host:5000/path) and <algo> is one of {RECOGNIZED_ALGORITHMS}")
InstallException: oci package 'oci://reg.example.com:fake_port/myplugin!my-plugin' is not in the expected format 'oci://<registry>:<tag>' or 'oci://<registry>@<algo>:<digest>' (optionally followed by '!<path>') in dynamic-plugins.yaml where <registry> may include a port (e.g. host:5000/path) and <algo> is one of ('sha512', 'sha384', 'sha256')

======= Cleaning up temporary catalog index directory
======= Removed lock file: /dynamic-plugins-root/install-dynamic-plugins.lock
```

</details>

rm3l (approved):
Looks good, thanks!

PR #4583, opened 2026-04-14
fix(e2e): fix bulk-import tests for 1.8 [AI /fix-e2e]

zdrapela:
## Summary
Fix bulk-import E2E test failures on release-1.8 nightly (`rhdh/rhdh-hub-rhel9:1.8`):
- Add optional chaining for `process.env.JOB_NAME` to prevent `TypeError` when undefined
- Replace `JOB_TYPE` presubmit check with `JOB_NAME` nightly check — tests only run on nightly OCP jobs
- Add `filterAndVerifyAddedRepo()` retry method to handle backend processing delays during repo ingestion
- Fix status text assertions to match actual UI (`"Added"` in Added repositories view)

## Test Results
- Verified against `quay.io/rhdh/rhdh-hub-rhel9:1.8` on OCP cluster
- All 13 bulk-import tests pass
- Stability: 3/3 passes
- Code quality: tsc, lint, prettier all pass
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-developer_rhdh/4583/pull-ci-redhat-developer-rhdh-release-1.8-e2e-ocp-helm-nightly/2044331931989970944/artifacts/e2e-ocp-helm-nightly/redhat-developer-rhdh-ocp-helm-nightly/artifacts/showcase-rbac-nightly/index.html#?q=bulk-import

## Related
- Prow nightly job: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-redhat-developer-rhdh-release-1.8-e2e-ocp-helm-nightly/2043902694631936000

### DIFF
--- a/e2e-tests/playwright/e2e/plugins/bulk-import.spec.ts
+++ b/e2e-tests/playwright/e2e/plugins/bulk-import.spec.ts
@@ -127,18 +127,13 @@ spec:
   });
 
   test('Verify that the two selected repositories are listed: one with the status "Already imported" and another with the status "WAIT_PR_APPROVAL."', async () => {
-    await bulkimport.filterAndVerifyAddedRepo(
-      catalogRepoDetails.name,
-      [catalogRepoDetails.url, "Added"],
-      uiHelper,
-      common,
-    );
-    await bulkimport.filterAndVerifyAddedRepo(
-      newRepoDetails.repoName,
-      ["Waiting for Approval"],
-      uiHelper,
-      common,
-    );
+    await bulkimport.filterAndVerifyAddedRepo(catalogRepoDetails.name, [
+      catalogRepoDetails.url,
+      "Added",
+    ]);
+    await bulkimport.filterAndVerifyAddedRepo(newRepoDetails.repoName, [
+      "Waiting for Approval",
+    ]);
   });
 
   test("Verify the Content of catalog-info.yaml in the PR is Correct", async () => {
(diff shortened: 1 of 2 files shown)


e2e-tests/playwright/support/pages/bulk-import.ts
```diff
@@ -58,4 +60,32 @@ export class BulkImport {
...
+   * that the row is visible with the expected status text. Retries the
+   * entire sequence to handle backend processing delays.
+   */
+  async filterAndVerifyAddedRepo(
+    repoName: string,
+    expectedCellTexts: (string | RegExp)[],
+    uiHelper: UIhelper,
+    common: Common,
+  ) {
+    await expect(async () => {
```
gustavolira, line 82: These `uiHelper` and `common` params are just wrappers around `this.page`. Could `BulkImport` own them itself (construct them in the constructor) so callers don't have to pass them on every call? Would drop four lines per call site and the dependency direction feels more natural — a page object owning its helpers rather than receiving them.

e2e-tests/playwright/e2e/plugins/bulk-import.spec.ts
```diff
@@ -213,15 +215,23 @@ spec:
...
-    await uiHelper.clickOnButtonInTableByUniqueText(
-      newRepoDetails.repoName,
-      "Refresh",
-    );
-    await uiHelper.verifyRowInTableByUniqueText(newRepoDetails.repoName, [
-      "Already imported",
-    ]);
+    // Verify that the status has changed to "Added" after merging the PR.
+    // Use retry to wait for the backend to process the merge.
+    await expect(async () => {
```
gustavolira: This is basically what `filterAndVerifyAddedRepo` does, just with a `Refresh` click between filter and verify. Since this PR just extracted that retry pattern into a helper, is it worth extending the helper (e.g. an optional `beforeAssert` step) or adding a sibling, so we don't keep the same pattern inline here?

zdrapela:
Closing in favor of https://github.com/redhat-developer/rhdh/pull/4606

PR #4594, opened 2026-04-15
chore: [main] include nodejs 22.22.2 header

josephca:
## Description

Update nodejs header to v22.22.2 from v22.22.0

- Fixes https://redhat.atlassian.net/browse/RHIDP-13100


nickboldt:
Main is moving to Node 24 in https://github.com/redhat-developer/rhdh/pull/4575 so we probably don't need this to be merged (and then have to rebase Hope's PR again)

Will close and cherrypick for 1.8 branch instead.

PR #4602, opened 2026-04-15
fix: replace-semantics for schedule duration subtrees (RHDHBUGS-2139)

gustavolira:
## Summary

Fixes [RHDHBUGS-2139](https://issues.redhat.com/browse/RHDHBUGS-2139): Keycloak catalog provider sync ran every `PT60M30S` (1h30s) instead of the user-configured 30s.

The dynamic-plugins loader deep-merges every subtree, which silently combines sibling keys of a Backstage `HumanDuration`. A default `schedule.frequency: { minutes: 60 }` plus a user-provided `schedule.frequency: { seconds: 30 }` collapsed into `{ minutes: 60, seconds: 30 }` — ISO-8601 `PT60M30S`.

## Fix

Patch `merge()` in `install-dynamic-plugins.py` so the three `HumanDuration` subtrees replace the destination dict instead of recursing:

- `schedule.frequency`
- `schedule.timeout`
- `schedule.initialDelay`

Whatever the most recent source provides is the absolute value — no leakage from a previous default. Also fixes the truncated prefix in the scalar-collision error message so it now shows the full config path.

Includes a TypeScript reference implementation + Jest tests in `dynamic-plugins/_utils/src/` to keep parity and prepare for the planned Python-to-TS migration of the installer script.

## Test plan

- [x] 12 new Jest tests pass (100% line coverage, 94.44% branch coverage on the merge module)
- [x] Covers each duration subtree individually (`frequency`, `timeout`, `initialDelay`)
- [x] Covers the full [RHDHBUGS-2139](https://redhat.atlassian.net/browse/RHDHBUGS-2139) repro path (`catalog.providers.keycloakOrg.default.schedule.frequency`)
- [x] Negative test: a `frequency` key outside a `schedule` subtree still deep-merges
- [x] Pre-existing Python tests still pass
- [ ] Manual verification: deploy with the patched script and confirm keycloak sync cadence matches the user config (reviewer task)

## Notes

- No behaviour change for plugins that do not define `schedule.*` subtrees.
- The TypeScript module is a reference implementation — it is not wired into the runtime yet. It exists to hold the intended semantics under test, and to anchor the future TS/Node migration of `install-dynamic-plugins.py`.

### DIFF
--- a/dynamic-plugins/_utils/src/merge-plugin-config.test.ts
+++ b/dynamic-plugins/_utils/src/merge-plugin-config.test.ts
@@ -1,0 +1,171 @@
+import { mergePluginConfig } from "./merge-plugin-config";
+
+describe("mergePluginConfig", () => {
+  it("copies scalar keys from source when destination is empty", () => {
+    const destination = {};
+    mergePluginConfig({ a: 1, b: "foo" }, destination);
+    expect(destination).toEqual({ a: 1, b: "foo" });
+  });
+
+  it("deep-merges non-duration sibling keys", () => {
+    const destination: Record<string, unknown> = { a: { b: 1 } };
+    mergePluginConfig({ a: { c: 2 } }, destination);
+    expect(destination).toEqual({ a: { b: 1, c: 2 } });
+  });
+
+  it("throws when a scalar key is defined with different values", () => {
+    const destination = { a: 1 };
+    expect(() => mergePluginConfig({ a: 2 }, destination)).toThrow(
+      "Config key 'a' defined differently for 2 dynamic plugins",
+    );
+  });
+
+  it("includes the full path in the collision error", () => {
+    const destination = { outer: { inner: { leaf: 1 } } };
+    expect(() =>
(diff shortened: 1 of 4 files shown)


gustavolira:
Closing in favor of surgical backports targeting the release branches where the bug actually reproduces:

- release-1.9: #4603
- release-1.8: #4604
- release-1.7: #4605

Rationale: on `main` the offending `dynamic-plugins.default.yaml` was already removed during the OCI migration, so the `PT60M30S` repro path does not exist on `main` today. This PR's merge-semantics patch would have been defensive-only code with no concrete failure mode to prevent, and it added non-trivial review surface (Python refactor, TS reference module, tests, CodeQL/SonarCloud interactions).

The three backport PRs each delete the seven-line `schedule:` block under `keycloakOrg.default`, letting the plugin's hardcoded `{ minutes: 30 }` fallback handle the unconfigured case and letting the user's value be honored verbatim when they do configure one.

PR #4638, opened 2026-04-20
feat(scripts): add yarn-lockfile-surgeon for minimum-version lockfile bumps

jonkoops:
`yarn-lockfile-surgeon` is a CLI tool for surgically bumping packages in Yarn Berry lockfiles. Unlike `yarn up`, which resolves dependency ranges to the latest matching version, this tool resolves to the **minimum satisfying version** — keeping lockfile diffs as small as possible when applying security patches on LTS branches.

The tool uses Yarn's own `@yarnpkg/parsers` and `@yarnpkg/core` to parse, manipulate, and serialize the lockfile. It walks transitive dependencies, adding new entries only when existing lockfile resolutions don't satisfy the updated ranges. Range aliases are collected to prevent `yarn install` from re-resolving covered ranges.

### DIFF
--- a/scripts/yarn-lockfile-surgeon/lib.ts
+++ b/scripts/yarn-lockfile-surgeon/lib.ts
@@ -1,0 +1,285 @@
+import { readFileSync, writeFileSync } from "node:fs";
+import { minSatisfying, satisfies } from "semver";
+import { parseSyml, stringifySyml } from "@yarnpkg/parsers";
+import { structUtils, Manifest } from "@yarnpkg/core";
+import type { Descriptor } from "@yarnpkg/core";
+import type { AbbreviatedVersion } from "package-json";
+
+// ---------------------------------------------------------------------------
+// Lockfile descriptor helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Splits a compound lockfile descriptor key into individual descriptor strings.
+ * Uses the same regex as Yarn's `Project.ts`.
+ */
+export function splitDescriptorKey(key: string): string[] {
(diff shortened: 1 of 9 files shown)


polasudo:
Nice One, but one thing I think is missing is the printout (conslo log or something) which would look like this "old version → new version" for each bumped package. The old version is right there in the lockfile    
  entry before it gets deleted, but it's never printed. I would like to see something like this eg: "@scope/pkg: 1.0.0 → 1.2.0" it would make it much easier to see what actually changed at 1st sight. Right now I would just need to run it and then go in and take a look if it changed everything it should.

jonkoops:
@polasudo Done, I added a bit nicer output so it is easier to see what the script does :+1:

scripts/yarn-lockfile-surgeon/lib.ts
```diff
@@ -0,0 +1,280 @@
...
+ * minimum satisfying versions, and writes the updated lockfile.
+ */
+export async function bumpLockfile(
+  lockfilePath: string,
+  targets: Descriptor[],
+  registry: RegistryFetcher,
+): Promise<void> {
+  const lock = parseSyml(
+    readFileSync(lockfilePath, "utf-8"),
+  );
```
polasudo, line 122: can we add here some consolle log if there is something wrong with the path ?

jonkoops: Added a more descriptive error, which should provide the user with all the information they need.

polasudo (approved):

PR #4641, opened 2026-04-20
fix(deps): [release-1.9] upgrade backstage packages to fix CVE-2026-24046

jonkoops:
Upgrades Backstage packages to their closest patched versions to address CVE-2026-24046 (GHSA-rq6q-wr2q-7pgp), a symlink path traversal vulnerability in Scaffolder actions.

| Package | Old | New |
|---|---|---|
| `@backstage/backend-defaults` | 0.13.1 | 0.13.2 |
| `@backstage/plugin-scaffolder-backend` | 3.0.2 (patch removed) | 3.0.2 (official fix) |
| `@backstage/plugin-scaffolder-node` | 0.12.1 | 0.12.3 |

`@backstage/plugin-scaffolder-backend@3.0.2` is already the official fix version per the advisory, so the redundant yarn patch was simply removed.

Replaces the previous `yarn patch:` based mitigation with the official fix versions. Lockfile changes were applied using `yarn-lockfile-surgeon` (#4638) to resolve transitive dependencies to their minimum satisfying versions, keeping the diff minimal.


yarn.lock
```diff
@@ -3087,6 +3003,28 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@backstage/backend-plugin-api@npm:^1.6.1, @backstage/backend-plugin-api@npm:^1.9.0":
+  version: 1.9.0
```
kim-tsao, line 3007: We have to lock this version down to 1.5.0 in RHDH 1.9 due to breaking changes.  See [comment](https://github.com/backstage/community-plugins/pull/6490#issuecomment-3682166411)

We also introduced a patch for @backstage/backend-plugin-api in this [PR](https://github.com/redhat-developer/rhdh/pull/4498) so we should try to resolve it to that patch version

kim-tsao (changes requested):

yarn.lock
alizard0: All backend-plugin-api are 1.5.0

PR #4656, opened 2026-04-21
fix(e2e): topology-rbac and kubernetes-rbac tests in k8s jobs

sanketpathak:
## Description

Fixing E2E Failures on K8s jobs in topology-rbac and kubernetes-rbac tests

https://redhat.atlassian.net/browse/RHDHBUGS-2817
- Fixes #?

## PR acceptance criteria

### DIFF
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -372,17 +372,13 @@ apply_yaml_files() {
   common::create_configmap_from_file "dynamic-global-header-config" "$project" \
     "dynamic-global-header-config.yaml" "$dir/resources/config_map/dynamic-global-header-config.yaml"
 
-  # Skip Topology resources for K8s deployments (AKS/EKS/GKE)
-  if [[ "$JOB_NAME" != *"aks"* && "$JOB_NAME" != *"eks"* && "$JOB_NAME" != *"gke"* ]]; then
-    # Create Deployment and Pipeline for Topology test.
-    oc apply -f "$dir/resources/topology_test/topology-test.yaml" --namespace="${project}"
-    if [[ -z "${IS_OPENSHIFT}" || "${IS_OPENSHIFT}" == "false" ]]; then
-      kubectl apply -f "$dir/resources/topology_test/topology-test-ingress.yaml" --namespace="${project}"
-    else
-      oc apply -f "$dir/resources/topology_test/topology-test-route.yaml" --namespace="${project}"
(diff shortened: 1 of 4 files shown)


.ci/pipelines/utils.sh
```diff
@@ -514,19 +517,29 @@ cluster_setup_ocp_operator() {
...
-  # operator::install_tekton
+  # Install Tekton Pipelines for K8s deployments (AKS/EKS/GKE) for topology tests to work
+  operator::install_tekton
   # operator::install_postgres_k8s # Works with K8s but disabled in values file
+
+  # Wait for Tekton Pipelines to be ready before proceeding
+  log::info "Waiting for Tekton Pipelines to be ready..."
+  k8s_wait::deployment "tekton-pipelines" "tekton-pipelines-webhook" 30 10 || return 1
+  k8s_wait::endpoint "tekton-pipelines-webhook" "tekton-pipelines" 1800 10 || return 1
+  k8s_wait::crd "pipelines.tekton.dev" 120 5 || return 1
```
jrichter1: these would make sense as part of `install_tekton` I think

sanketpathak: I did this to make it consistent with the pipeline setup in OCP https://github.com/redhat-developer/rhdh/blob/9045e7b27624ac59e01011dec71be9adc05187d8/.ci/pipelines/utils.sh#L507-L508

jrichter1: I don't see why waiting for the operator shouldn't be part of the respective install function in either case. Both are always called, then followed by the waits, which is also duplicating code.

If you want to be consistent, merge both install functions with the waits :)

sanketpathak: This is part of waitfor_tekton_pipelines

zdrapela: I agree with @jrichter1. I'd merge all the installations with the waits to keep it readable, consistent, and easy to understand.

The `waitfor_tekton_pipelines` was probably orphaned during refactorings. 

I recall that at some point, the installation was intentionally split from the waits, so that installation of multiple resources could be triggered, and then we could wait for all of them to be ready. But I don't see much value in that and it's making the scripts overly complicated.

zdrapela (approved):
Nice work! It's cool to see it passing at all the platforms 👍 LGTM

teknaS47 (approved):
Looks good.
/lgtm

PR #4707, opened 2026-04-28
feat(scorecard): add scorecard openssf to default packages tech-preview

alizard0:
It adds scorecard, scorecard-backend and scorecard-backend-module-openssf to default.packages.yaml to make them TP.

https://redhat.atlassian.net/browse/RHIDP-13317


dzemanov:
/hold
@alizard0 let's hold on this till we get clearance about if TP support is alright to be implemented for 1.10 (as part of [RHDHPLAN-1234](https://redhat.atlassian.net/browse/RHDHPLAN-1234) feature), as it is already feature freeze.

alizard0:
Closing the PR because the TP promotion isn't ready yet.
https://redhat.atlassian.net/browse/RHDHPLAN-1234

PR #4767, opened 2026-05-06
chore(deps): [release-1.9] patch vm2 to 3.11.0 or higher

alizard0:
https://redhat.atlassian.net/browse/RHIDP-13453


JessicaJHee (approved):
/lgtm

Verified `vm2` is not present in root, and in `dynamic-plugins` version `>= 3.11.0`

PR #4776, opened 2026-05-07
fix: move lightspeed plugins to disabled in `default.packages.yaml`

rm3l:
## Description

Lightspeed plugins are already enabled by default in the install methods, so they should not also be in the enabled list of `default.packages.yaml`.
Otherwise, since the current behavior in the install methods when lightspeed is disabled is to skip adding those plugins to the list of plugins, if they are enabled in the DPDPY, they will still be installed automatically by the install-dynamic-plugins script.

- Fixes https://redhat.atlassian.net/browse/RHDHBUGS-3092

## PR acceptance criteria


Jdubrick (approved):
lgtm

michael-valdron (approved):

PR #4798, opened 2026-05-12
feat(e2e): enable V8 coverage collection with Codecov e2e flag

gustavolira:
## Summary

Enable V8 JavaScript coverage collection for all E2E tests, with a dedicated `rhdh-e2e-frontend` Codecov flag to keep E2E coverage separate from unit test coverage.

- **Migrate 38 spec files** from `import { test } from "@playwright/test"` to `import { test } from "@support/coverage/test"` — the coverage module wraps the `page` fixture with `page.coverage.startJSCoverage()`/`stopJSCoverage()` and uses the existing `monocart-coverage-reports` reporter to merge raw V8 data into LCOV
- **Add `rhdh-e2e-frontend` flag to `codecov.yml`** — same source paths as the `rhdh` (unit test) flag, with `carryforward: true`
- **Enable coverage in CI by default** — `COLLECT_COVERAGE=true` is set in `testing::run_tests` before `yarn playwright test`
- **Clear stale coverage before each run** — `coverage/e2e/` and `coverage/e2e-raw/` are removed before Playwright runs, preventing stale lcov.info from a previous project being uploaded
- **Save and upload coverage** — after each test run, `coverage/e2e/lcov.info` is saved as a CI artifact and uploaded to Codecov with `--flag rhdh-e2e-frontend` when `CODECOV_TOKEN` is available; the Codecov CLI binary is verified with SHA256 checksum before execution
- **Re-export all `@playwright/test` types** from the coverage module via `export *` so spec files don't need split imports for types like `Page`, `BrowserContext`, etc.
- **Fix ESLint warnings in `auditor-rbac.spec.ts`** — refactored conditional test logic into data-driven test parameters to eliminate `playwright/no-conditional-in-test` and `playwright/no-conditional-expect` warnings

## How it works

```
1. COLLECT_COVERAGE=true → coverage fixture starts V8 JS coverage on page
2. Test exercises the RHDH UI
3. Fixture teardown stops JS coverage and writes raw V8 entries to coverage/e2e-raw/
4. Monocart reporter (onEnd) merges all raw JSONs → coverage/e2e/lcov.info
5. testing.sh uploads lcov.info to Codecov with --flag rhdh-e2e-frontend
```

## What this does NOT cover

- **CODECOV_TOKEN in Prow**: The token needs to be configured in the Prow job secrets. Without it, coverage is still collected and saved as an artifact but not uploaded.

## Files changed

| Area | Files | Description |
|------|-------|-------------|
| Codecov config | `codecov.yml` | Add `rhdh-e2e-frontend` flag with carryforward |
| CI pipeline | `.ci/pipelines/lib/testing.sh` | Set `COLLECT_COVERAGE`, clear stale coverage, save + upload with SHA256-verified Codecov CLI |
| Coverage module | `e2e-tests/playwright/support/coverage/test.ts` | Add `export *` for type re-exports |
| ESLint fix | `e2e-tests/playwright/e2e/audit-log/auditor-rbac.spec.ts` | Refactor conditionals into data-driven test params |
| Spec files | 38 files in `e2e-tests/playwright/e2e/` | Change import source to `@support/coverage/test` |

## Test plan

- [ ] `COLLECT_COVERAGE=true yarn playwright test --project=smoke-test` produces `coverage/e2e/lcov.info`
- [ ] Without `COLLECT_COVERAGE`, no coverage files are created (zero overhead)
- [ ] TypeScript compiles without errors (`npx tsc --noEmit`)
- [ ] Codecov dashboard shows `rhdh-e2e-frontend` flag separately from `rhdh` flag
- [ ] CI test run saves coverage artifact even without `CODECOV_TOKEN`
- [ ] ESLint reports zero warnings on `auditor-rbac.spec.ts`

Jira: [RHIDP-13243](https://redhat.atlassian.net/browse/RHIDP-13243)

### DIFF
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -82,6 +82,10 @@ testing::run_tests() {
   # the reporter merges raw JSON into lcov via monocart-coverage-reports.
   export COLLECT_COVERAGE="${COLLECT_COVERAGE:-false}"
 
+  # Remove stale coverage artifacts so a previous project's lcov.info
+  # is never mistakenly uploaded for the current run.
+  rm -rf "${e2e_tests_dir}/coverage/e2e" "${e2e_tests_dir}/coverage/e2e-raw"
+
   (
     set -e
     log::info "Using PR container image: ${TAG_NAME}"
@@ -108,7 +112,7 @@ testing::run_tests() {
   if [[ -f "${e2e_tests_dir}/coverage/e2e/lcov.info" ]]; then
     common::save_artifact "${artifacts_subdir}" "${e2e_tests_dir}/coverage/e2e/" "coverage" || true
     if [[ -n "${CODECOV_TOKEN:-}" ]]; then
-      log::info "Uploading E2E coverage to Codecov (flag: e2e)..."
+      log::info "Uploading E2E coverage to Codecov (flag: rhdh-e2e-frontend)..."
       local codecov_bin="/tmp/codecov"
       if [[ ! -x "$codecov_bin" ]]; then
         curl -sL -o "$codecov_bin" https://cli.codecov.io/latest/linux/codecov
@@ -127,7 +131,7 @@ testing::run_tests() {
         "$codecov_bin" upload-process \
           --token "${CODECOV_TOKEN}" \
           --file "${e2e_tests_dir}/coverage/e2e/lcov.info" \
-          --flag e2e \
+          --flag rhdh-e2e-frontend \
           --slug redhat-developer/rhdh \
           --fail-on-error || log::warn "Codecov E2E coverage upload failed (non-fatal)"
       fi
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -76,6 +76,11 @@ testing::run_tests() {
   Xvfb :99 &
   export DISPLAY=:99
 
+  # RHIDP-13243: Enable V8 coverage collection for E2E tests.
+  # The coverage fixture wraps page.coverage.startJSCoverage/stopJSCoverage
+  # and the reporter merges raw JSON into lcov via monocart-coverage-reports.
+  export COLLECT_COVERAGE="${COLLECT_COVERAGE:-true}"
+
   (
     set -e
     log::info "Using PR container image: ${TAG_NAME}"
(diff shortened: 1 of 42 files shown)


gustavolira:
/review
-i

.ci/pipelines/lib/testing.sh
```diff
@@ -76,6 +76,16 @@ testing::run_tests() {
   Xvfb :99 &
   export DISPLAY=:99
 
+  # RHIDP-13243: V8 coverage collection for E2E tests (opt-in).
+  # Set COLLECT_COVERAGE=true in the job config to enable. When enabled, the
+  # coverage fixture wraps page.coverage.startJSCoverage/stopJSCoverage and
+  # the reporter merges raw JSON into lcov via monocart-coverage-reports.
+  export COLLECT_COVERAGE="${COLLECT_COVERAGE:-false}"
```
subhashkhileri, line 83: `COLLECT_COVERAGE` defaults to `false` here — is there somewhere else in the job config that flips it on, or is coverage actually disabled by default right now?

e2e-tests/playwright/support/coverage/test.ts
```diff
@@ -21,6 +21,10 @@ import {
   type Page,
   type TestInfo,
 } from "@playwright/test";
+// Re-export all Playwright types and values so specs can replace
+// `from "@playwright/test"` with this module. The locally-defined `test`
+// and `expect` below shadow the star re-exports.
+export * from "@playwright/test";
```
subhashkhileri, line 27: Specs that call `setupBrowser` create their own `BrowserContext` and `Page`, so they bypass the `page` fixture override entirely — coverage never starts for those pages even when `COLLECT_COVERAGE=true`.

.ci/pipelines/lib/testing.sh
gustavolira: Good catch — `COLLECT_COVERAGE` defaults to `false` intentionally. Coverage is opt-in because it adds overhead (V8 JS profiler + writing raw JSON files per test). No CI job enables it yet; the plan is to turn it on in a dedicated nightly job once the infrastructure is validated end-to-end. The default ensures zero impact on existing CI runs.

e2e-tests/playwright/support/coverage/test.ts
gustavolira: Great catch — this was a real gap. Fixed in f66624d5: `setupBrowser()` now calls `startCoverageForPage(page)` before returning, and a new `teardownBrowser(page, testInfo)` helper flushes coverage via `stopCoverageForPage` then closes the page.

Updated 7 spec files to use `teardownBrowser` in `afterAll`. There are still ~8 specs with complex `afterAll` hooks (auth-providers, bulk-import, orchestrator-rbac) that start coverage now but don't flush it — those will be addressed as a follow-up to keep this PR focused.

subhashkhileri (approved):
/lgtm

PatAKnight (approved):
LGTM

PR #4802, opened 2026-05-13
feat(app-next): add auth plugin to frontend app

kadel:
## Description

The `@backstage/plugin-auth` plugin provides the OAuth2 consent page
`(/oauth2/authorize/:sessionId)` needed by the new Backstage frontend system. Without it, the app
 has no route to handle OAuth2 authorization consent flows — when a third-party tool (like
`backstage-cli`) initiates an OAuth2 session, the user cannot be redirected to a consent page to
approve or deny the request.

This was discovered while testing `backstage-cli actions execute`, which requires an
authenticated session via the OAuth2 CLI flow.

## PR acceptance criteria

### DIFF
--- a/packages/app-next/package.json
+++ b/packages/app-next/package.json
@@ -51,6 +51,7 @@
     "@backstage/integration-react": "1.2.16",
     "@backstage/plugin-app": "0.4.2",
     "@backstage/plugin-app-visualizer": "0.2.1",
+    "@backstage/plugin-auth": "0.1.6",
     "@backstage/plugin-auth-react": "0.1.25",
     "@backstage/plugin-catalog": "2.0.1",
     "@backstage/plugin-catalog-common": "1.1.8",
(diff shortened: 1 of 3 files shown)


gashcrumb:
There's a [dynamic plugin](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pkgs/container/rhdh-plugin-export-overlays%2Fbackstage-plugin-auth/862083482?tag=bs_1.49.4__0.1.6) for this that could be configured.  Is this a common enough use case that we should include this statically by default, or should we update the dynamic plugin config so that this is normally installed by default when we move to the NFS?

kadel:
> There's a [dynamic plugin](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pkgs/container/rhdh-plugin-export-overlays%2Fbackstage-plugin-auth/862083482?tag=bs_1.49.4__0.1.6) for this that could be configured. Is this a common enough use case that we should include this statically by default, or should we update the dynamic plugin config so that this is normally installed by default when we move to the NFS?

You are probably right, doing this via dynamic plugin might be a better. I will test it. 

I expect that this will be quite common as it is requirement for `backstage-cli` to work, and `backstage-cli` will be increasingly more important piece for agentic development. 
It should be definitely installed by default, but doing it via dynamic plugin is fine.

kadel:
https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2498/ updates backstage-auth dynamic plugin to work on both new and old frontend system. No changes on rhdh site required

PR #4804, opened 2026-05-13
chore: update yarn-lockfile script to fix some errors

kim-tsao:
## Description

While running `yarn-lockfile-surgeon` to update `fast-xml-builder` to v1.1.7, I ran into an error:

`➤ YN0001: │ Error: path-expression-matcher@^1.1.3 isn't supported by any available resolver`

which was indicative of a couple of problems:

1.  `path-expression-matcher@npm:^1.1.3` was incorrectly removed from the [yarn.lock ](https://github.com/redhat-developer/rhdh/blob/main/yarn.lock#L31996) leaving only the ^1.5.0 version available for resolution
2.  The yarn.lock generated in phase3 did not qualify the path-expression-matcher dependency with `npm` which caused issues when running `yarn install --mode=update-lockfile`

changes looked something like this:

```
fast-xml-builder@npm:1.1.7, fast-xml-builder@npm:^1.1.5":
  version: 1.1.7
  resolution: "fast-xml-builder@npm:1.1.7"
  dependencies:
    path-expression-matcher: "^1.1.3" instead of path-expression-matcher: "npm:^1.1.3"
```

Assisted-by: Cursor

- Fixes #?

## PR acceptance criteria


jonkoops:
Thanks @kim-tsao, I went ahead and pushed a couple of commits on top of the original implementation you had.

The main change is that instead of post-processing the entire lockfile to fix up bare semver ranges, we now qualify them with `npm:` right when registry metadata is fetched. That way `Manifest.load` creates properly qualified descriptors from the start and `buildLockEntry` outputs correct ranges without any fixup, removing the need for `qualifyNpmDescriptorsInLockfile` and `qualifyBareNpmDependencyRange` entirely.

I also replaced the hand-rolled protocol parsing in `lockfileDependencyNpmSelector` with inline `structUtils.parseRange` calls from `@yarnpkg/core`, which already knows how to decompose range strings. The function was small enough after that to just inline at its two call sites.

On the test side, the existing fixtures now use `npm:`-prefixed dependency values matching real Yarn lockfile format, so this class of bug would be caught going forward even without the dedicated test (which I removed since it duplicated existing coverage).

This is looking good to me, but I will ask another team member to review it since I pushed some changes as well.

jonkoops (approved):

PR #4809, opened 2026-05-14
refactor(e2e): refactor runtime tests into TypeScript

zdrapela:
## Goal

Consolidate `SHOWCASE_RUNTIME_DB` into `SHOWCASE_RUNTIME` — a single Playwright project that deploys RHDH from TypeScript (`runtime-deploy.ts`) and runs all runtime tests sequentially (`workers: 1`).

## What the runtime tests actually test

All 4 test files are infrastructure-level tests that only need a basic RHDH instance up and serving the UI:

| Test | What it validates |
|------|-------------------|
| `config-map.spec.ts` | ConfigMap `app.title` change propagates after deployment restart |
| `verify-schema-mode.spec.ts` | RHDH boots with `pluginDivisionMode: schema` on a restricted (`NOCREATEDB`) DB user |
| `verify-tls-config-with-external-rds.spec.ts` | RHDH connects to 4 AWS RDS PostgreSQL versions over TLS |
| `verify-tls-config-with-external-azure-db.spec.ts` | RHDH connects to 4 Azure DB PostgreSQL versions over TLS |

None exercise catalog, scaffolder, TechDocs, search, Kubernetes, ArgoCD, or any other domain-specific plugin. They only need guest auth and the core backend serving the UI.

### Single source of truth (`runtime-config.ts`)

All deployment configuration is generated from a single TypeScript module — no static YAML files. The same config produces both Helm values and Operator Backstage CR, ensuring the two install methods stay in sync.

- **Helm path:** generates values YAML as a temp file with only chart-default overrides + `--set` flags for image/cluster. Lightspeed disabled via `global.lightspeed.enabled: false`. PVC for dynamic-plugins-root persists plugins across restarts.
- **Operator path:** generates app-config ConfigMap, dynamic-plugins ConfigMap (`includes: [], plugins: []`), and Backstage CR programmatically. Lightspeed disabled via `flavours: []`.
- **CATALOG_INDEX_IMAGE:** opt-in override. Helm uses `global.catalogIndex.image.*` (matching CI's `helm::get_image_params()`); Operator pushes env var with `containers: ["install-dynamic-plugins"]`. When not set, chart/operator defaults apply.

### Deployment moved from shell to TypeScript

CI scripts (`ocp-nightly.sh`, `ocp-operator.sh`) are reduced to thin wrappers that set env vars and invoke Playwright. All deployment logic lives in `runtime-deploy.ts`:

- `ensureRuntimeDeployed()` — idempotent, called from the first test's `beforeAll`
- Supports both Helm (`helm upgrade -i`) and Operator (`Backstage CR`) install methods
- Discovers PostgreSQL service/secret for schema-mode env vars
- External DB tests (`postgres-config.ts`) reconfigure the running instance at runtime

### Shared utilities extracted

- `resolveInstallMethod()`, `base64Encode/Decode()` → `helper.ts`
- `patchAppConfig()`, `restartDeploymentWithRetry()`, `jsonPatchDeployment()` → `KubeClient`
- `AppConfigYaml` interface, `ImageRef` type, `imageRefToString()` → `runtime-config.ts`
- `getDeploymentName()` in schema-mode-setup now delegates to `getRhdhDeploymentName()`
- `run()`, `discoverRouterBase()` exported from `runtime-deploy.ts`

### Files deleted (replaced by TypeScript generation)

- `.ci/pipelines/resources/postgres-db/values-showcase-postgres.yaml` → generated by `generateHelmValuesYaml()`
- `.ci/pipelines/resources/postgres-db/rds-app-config.yaml` → generated by `generateAppConfigYaml()`
- `.ci/pipelines/resources/rhdh-operator/rhdh-start-runtime.yaml` → generated by `generateBackstageCR()`

Jira: https://redhat.atlassian.net/browse/RHIDP-9140
Jira: https://redhat.atlassian.net/browse/RHIDP-9141


e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-rds.spec.ts
```diff
@@ -41,51 +42,61 @@ test.describe("Verify TLS configuration with RDS PostgreSQL health check", () =>
...
+        "RDS environment variables not configured (RDS_DB_CERTIFICATES_PATH, RDS_USER, RDS_PASSWORD) — RDS tests are opt-in",
       );
+      return;
     }
 
     const kubeClient = new KubeClient();
 
+    // Prepare the deployment for external database tests: patch the app-config
+    // to use env var placeholders and clean up any schema-mode env var patches
+    await prepareForExternalDatabase(kubeClient, namespace, deploymentName);
```
gustavolira, line 65: Only `config-map.spec.ts` calls `ensureRuntimeDeployed()` — this spec (and the schema-mode one) just assume the deployment already exists, which only holds because Playwright discovers files alphabetically and `configuration-test/` sorts before `external-database/`. Run the RDS or schema-mode spec on its own (or with `--grep`) and `prepareForExternalDatabase` will fail on a missing ConfigMap. Since `ensureRuntimeDeployed()` is already idempotent, can we just call it in each runtime spec's `beforeAll` and drop the hidden ordering dependency the old `dependencies: [SHOWCASE_RUNTIME_DB]` used to make explicit?

e2e-tests/playwright/utils/kube-client.ts
```diff
@@ -91,29 +123,7 @@ export class KubeClient {
...
-        contexts: [
-          {
-            name: "default-context",
-            user: "ci-user",
-            cluster: "my-openshift-cluster",
-          },
-        ],
-        currentContext: "default-context",
-      });
+      this.kc.loadFromDefault();
```
gustavolira: This swaps the explicit `K8S_CLUSTER_URL`/`K8S_CLUSTER_TOKEN` config for `loadFromDefault()`, which changes how every test that builds a `KubeClient` authenticates to the cluster, not just runtime. It works if the job has a kubeconfig context (e.g. after `oc login`), but did we confirm that holds across all the jobs that use KubeClient — auth-providers and the EKS/GKE/AKS ones included? A missing context would now fail differently than before.

e2e-tests/playwright/utils/runtime-deploy.ts
```diff
@@ -0,0 +1,487 @@
...
+ * (helm) or operator-managed StatefulSet (operator). The deployment
+ * happens once in the first test file's beforeAll — subsequent specs
+ * reuse the existing deployment since the project runs with workers: 1.
+ *
+ * All deployment configuration is generated from `runtime-config.ts` —
+ * a single source of truth that produces Helm values YAML, Operator
+ * app-config, dynamic-plugins ConfigMaps, and the Backstage CR.
+ *
+ * Environment variables consumed:
+ *   K8S_CLUSTER_URL, K8S_CLUSTER_TOKEN  — cluster access
```
gustavolira: nit: with `KubeClient` now on `loadFromDefault()`, these two env vars aren't actually consumed anymore — worth dropping them from this list so the header stays accurate.

e2e-tests/playwright/utils/runtime-config.ts
```diff
@@ -0,0 +1,411 @@
...
+
+/**
+ * Generate the Backstage CR object for the operator path.
+ *
+ * The CR uses spec.deployment.patch to override the container image and
+ * spec.application for app-config, dynamic plugins, extra files, env vars,
+ * and route configuration.
+ */
+export function generateBackstageCR(config: RuntimeDeployConfig): BackstageCR {
+  const fullImage = `${config.image.registry}/${config.image.repository}:${config.image.tag}`;
```
gustavolira: This always uses `:` as the separator, so a digest-pinned main image (`...@sha256:...`) would come out as `repo:sha256:...`. You already handle this correctly for the catalog index via `ImageRef`/`imageRefToString` — reusing it here would keep the two consistent. Same limitation existed in the old YAML so it's not a regression, just a chance to close the gap.

e2e-tests/playwright/utils/postgres-config.ts
```diff
@@ -236,3 +234,129 @@ export async function clearDatabase(credentials: {
...
+    "POSTGRES_PORT",
+    "POSTGRES_DB",
+    "POSTGRES_USER",
+    "POSTGRES_PASSWORD",
+  ];
+
+  const removed = await kubeClient.removeContainerEnvVars(
+    deploymentName,
+    namespace,
+    "backstage-backend",
```
gustavolira: nit: `"backstage-backend"` is hard-coded here, in `ensurePostgresCredEnvVars`, and in the CR generator. You already centralized the deployment name in `getRhdhDeploymentName()` — a sibling constant for the container name would keep these from drifting if it's ever renamed.

e2e-tests/playwright/utils/postgres-config.ts
```diff
@@ -236,3 +234,129 @@ export async function clearDatabase(credentials: {
...
+ * they may come from Helm chart templates, schema-mode patches, or other sources),
+ * then adds fresh secretKeyRef entries pointing to the postgres-cred secret.
+ * This ensures the app-config ${POSTGRES_HOST} etc. placeholders resolve from postgres-cred.
+ */
+async function ensurePostgresCredEnvVars(
+  kubeClient: KubeClient,
+  deploymentName: string,
+  namespace: string,
+): Promise<void> {
+  const keys = [
```
gustavolira: nit: this key list overlaps heavily with `schemaModeVars` above. Could we pull a shared `POSTGRES_ENV_KEYS` constant and append the extra two (`PGSSLMODE`, `NODE_EXTRA_CA_CERTS`) where needed, so the two don't fall out of sync?

e2e-tests/playwright/utils/kube-client.ts
```diff
@@ -857,6 +828,219 @@ export class KubeClient {
...
+        `No app-config data key found in ConfigMap '${configMapName}'`,
+      );
+    }
+
+    const originalYaml = configMap.data[configKey];
+    const appConfig = yaml.load(originalYaml) as Record<string, unknown>;
+    mutator(appConfig);
+
+    const updatedYaml = yaml.dump(appConfig);
+    if (updatedYaml === yaml.dump(yaml.load(originalYaml))) {
```
gustavolira: nit: this re-parses and re-dumps `originalYaml` just for the no-op check. Capturing `const normalizedOriginal = yaml.dump(yaml.load(originalYaml))` once before the mutator runs reads a bit cleaner and avoids the second parse. Behaviour is correct as-is.

e2e-tests/playwright/utils/runtime-deploy.ts
```diff
@@ -0,0 +1,487 @@
...
+    return;
+  }
+
+  // Export schema-mode env vars
+  process.env.SCHEMA_MODE_PORT_FORWARD_NAMESPACE = namespace;
+  process.env.SCHEMA_MODE_PORT_FORWARD_RESOURCE = `svc/${svcName}`;
+  process.env.SCHEMA_MODE_DB_ADMIN_USER = "postgres";
+  process.env.SCHEMA_MODE_DB_ADMIN_PASSWORD = adminPassword;
+  process.env.SCHEMA_MODE_DB_PASSWORD =
+    process.env.SCHEMA_MODE_DB_PASSWORD || "test_password_123";
```
gustavolira: nit: `"test_password_123"` and `"bn_backstage"` are baked inline here. They're throwaway schema-mode values so it's low-stakes, but named constants near the top would document intent better than magic strings.

e2e-tests/playwright/utils/kube-client.ts
```diff
@@ -309,116 +326,21 @@ export class KubeClient {
   }
 
   async updateConfigMapTitle(
-    configMapName: string,
+    _configMapName: string,
```
gustavolira: nit: the first param is now ignored (renamed to `_configMapName`) since the app-config map is resolved dynamically. If `config-map.spec.ts` is the only caller, dropping the param entirely would be cleaner than keeping a permanently-unused one.

e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-rds.spec.ts
zdrapela: Nice catch! Let me do that

e2e-tests/playwright/utils/kube-client.ts
zdrapela: This change actually aligns the `kube-client` with how `auth-providers` use it in `RHDHDeployment`: https://github.com/zdrapela/rhdh/blob/476594f768bdabba8d0fb3c641ec785cb0b4953a/e2e-tests/playwright/utils/authentication-providers/rhdh-deployment.ts#L50, as well as `rhdh-e2e-test-utils` https://github.com/redhat-developer/rhdh-e2e-test-utils/blob/685727f09997b0b2da4ecc9c5f8abb55226592ef/src/utils/kubernetes-client.ts#L13
And no K8s job uses it. 
I believe this implementation is safe, as the CI script already relies on being authenticated in the default Kubeconfig, including the `locar-run.sh`.

e2e-tests/playwright/utils/runtime-deploy.ts
zdrapela: You're right!

e2e-tests/playwright/utils/runtime-config.ts
zdrapela: Okay, let's do it

e2e-tests/playwright/utils/postgres-config.ts
zdrapela: That's actually not the deployment name, but the name of the backend container. But it's true that it makes sense to have it stored in a variable, instead of being hardcoded in a text.

e2e-tests/playwright/utils/runtime-deploy.ts
zdrapela: True, and I also moved the `schema-mode` specific parts to its helper

e2e-tests/playwright/utils/kube-client.ts
zdrapela: True. In the end I found it better to even get rid of `updateConfigMapTitle` and instead just use `patchAppConfig` here

gustavolira (approved):

gustavolira (approved):

PR #4821, opened 2026-05-15
chore(deps): [release-1.8] patch protobufjs by upgrading @google-cloud/firestore

alizard0:
Upgraded @google-cloud/firestore to bump `protobufjs`
https://redhat.atlassian.net/browse/RHIDP-13301


kim-tsao (changes requested):
Do we want to bump up all versions of protobufjs to v7.5.8?

kim-tsao (approved):
/lgtm

PR #4864, opened 2026-05-20
test: add Layer 3 component tests mirroring UI E2E specs, plus Playwright tag system (RHIDP-13235)

gustavolira:
## Description

Implements **[RHIDP-13235](https://redhat.atlassian.net/browse/RHIDP-13235)** — additive Layer 3 component tests that mirror the behavior of existing UI-only E2E specs, plus a Playwright tag system. Part of the [RHIDP-11861](https://redhat.atlassian.net/browse/RHIDP-11861) test-layer work.

This is **not** a migration: every existing `.spec.ts` keeps all its tests. The goal is to cover the same UI behaviors at a faster, cluster-free layer (React Testing Library + `@backstage/test-utils`) and to make those layers identifiable via tags.

## Component tests added (behavior → component)

| Mirrors E2E spec | Component test |
|------------------|----------------|
| learning-path-page | `LearningPathsPage.test.tsx` |
| user-settings-info-card | `InfoCard.test.tsx` (extended) |
| settings | `GeneralPage.test.tsx` |
| sidebar | `CustomSidebarItem.test.tsx` |
| custom-theme | `useThemedConfig.test.tsx` |
| header-mount-points | `getMountPointData.test.ts` |

Each test carries an inline header noting the E2E spec it mirrors.

## Playwright tag system

- Tagged the four E2E specs that now have a sibling Layer 3 test with `@layer3-equivalent`, and `smoke-test` with `@smoke` (tags live in the `describe` title so `--grep` selects them).
- `.ci/pipelines/lib/testing.sh` honors an optional `PLAYWRIGHT_GREP` env var, passing `--grep` through to the run (no-op when unset).
- `e2e-tests/README.md` documents the tag glossary (`@layer3-equivalent`, `@smoke`, `@ga-plugin`, `@non-ga-plugin`) and how to filter locally and in CI.

## Scope boundaries

- `extensions` and `dynamic-home-page-customization` render from plugin-side code (rhdh-plugins), not from a component in this repo — out of scope here, proposed as a follow-up in that repo.
- `home-page-customization` is rendered dynamically via Scalprum; the reachable piece (mount-point resolution) is covered by `getMountPointData.test.ts`.
- Tooling: uses `@backstage/test-utils` (`renderInTestApp` / `TestApiProvider`), matching the existing pattern in `packages/app`.

## Testing

- [x] 6 component test suites pass (20 tests)
- [x] `yarn tsc`, lint and prettier pass (app + e2e-tests + .ci)
- [x] New tests contribute to the Codecov `rhdh` flag

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -185,6 +185,37 @@ npx playwright show-report .local-test/rhdh/.local-test/artifact_dir/showcase
 
 ---
 
+### Test tags
+
+Specs can carry `@tag` markers in their `test.describe` title. Playwright's
+`--grep` / `--grep-invert` filters select tests by tag, letting CI or a local
+run target a subset.
+
+| Tag                  | Meaning                                                                                   |
+| -------------------- | ----------------------------------------------------------------------------------------- |
+| `@layer3-equivalent` | A UI behavior in this spec also has a sibling Layer 3 component test (in `packages/app`). |
+| `@smoke`             | Fast, high-signal check suitable to run on every PR.                                      |
+| `@ga-plugin`         | Exercises a generally-available (GA) plugin.                                              |
+| `@non-ga-plugin`     | Exercises a tech-preview / dev-preview (non-GA) plugin.                                   |
+
+```bash
+# Run only smoke-tagged tests
+yarn playwright test --project=showcase --grep "@smoke"
+
+# Run everything except specs that already have a Layer 3 equivalent
+yarn playwright test --project=showcase --grep-invert "@layer3-equivalent"
+```
+
+In CI, set the `PLAYWRIGHT_GREP` environment variable (consumed by
+`.ci/pipelines/lib/testing.sh`) to apply the same filter to a job, e.g.
+`PLAYWRIGHT_GREP='@smoke'`.
+
+> **Note:** `@layer3-equivalent` marks overlap with Layer 3 coverage; it does
+> not remove the E2E spec. Whether to retire any duplicated E2E coverage is a
+> separate decision tracked elsewhere.
+
+---
+
 ### Configuration Options
 
 #### Job Types
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -187,9 +187,13 @@ npx playwright show-report .local-test/rhdh/.local-test/artifact_dir/showcase
 
 ### Test tags
 
-Specs can carry `@tag` markers in their `test.describe` title. Playwright's
-`--grep` / `--grep-invert` filters select tests by tag, letting CI or a local
-run target a subset.
+Specs declare `@tag` markers via Playwright's native
+[test tags](https://playwright.dev/docs/test-annotations#tag-tests) — the `tag`
+option on `test` / `test.describe` (e.g.
+`test.describe("Smoke test", { tag: "@smoke" }, () => { ... })`). This keeps the
+tag out of the test title, so names stay stable for historical reporting while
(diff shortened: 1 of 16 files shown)


e2e-tests/README.md
```diff
@@ -185,6 +185,37 @@ npx playwright show-report .local-test/rhdh/.local-test/artifact_dir/showcase
 
 ---
 
+### Test tags
```
zdrapela, line 188: I propose using test tags (Playwright feature, please see the docs: https://playwright.dev/docs/test-annotations#tag-tests) instead of embedding "tag-like" expressions in the description (legacy way of doing that).
That is a cleaner solution, and grep filtering will still work.
That way, we won't pollute the descriptions and the test names will remain the same, so they will match with historical data in ReportPortal, etc.

packages/app/src/components/learningPaths/LearningPathsPage.test.tsx
zdrapela, line 1: I see these tests target the old frontend, while the plan for 2.1 is to switch to the new frontend system (NFS) (https://redhat.atlassian.net/browse/RHDHPLAN-846). 
- This follows https://backstage.io/docs/plugins/testing/#utility-functions
- For NFS, https://backstage.io/docs/backend-system/building-plugins-and-modules/testing/ and https://backstage.io/docs/frontend-system/building-plugins/testing/ will apply. 

I just want to check if this is intentional to add the tests now, before migration, and you expect to migrate the tests to NFS afterward.

packages/app/src/components/learningPaths/LearningPathsPage.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      name: /Building Operators on OpenShift/,
+    });
+    expect(link).toHaveAttribute(
+      'href',
+      'https://developers.redhat.com/learn/openshift/operators',
+    );
+    expect(link).toHaveAttribute('target', '_blank');
+  });
+
+  it('shows an error report when no learning paths are returned', async () => {
```
zdrapela, line 89: Food for thought:
Do we really want to stop at mirroring only what E2E does? Sure, 100 % code coverage is not the holy grail, but if we extended the test a bit, we would have discovered an unreachable statement: https://github.com/gustavolira/rhdh/blob/e84d5e648defd2b0e630c3ce71b5c02c4c37a07c/packages/app/src/components/learningPaths/LearningPathsPage.tsx#L49-L58 And unit tests are cheap...

gustavolira: @zdrapela This PR is intentionally just an example/reference, it's part of a series showing tests at layer 1, 2 and 3. I also want to follow up with a PR demonstrating how to run e2e tests in CI without depending on a running OpenShift cluster, just by building the frontend/backend locally.

So I didn't go deep into the tests on purpose, these are meant as a starting point, not full coverage.

That said, it's great that you reviewed it and spotted more ways to evolve the test (and found that unreachable branch)

packages/app/src/components/learningPaths/LearningPathsPage.test.tsx
gustavolira: These components live only in the legacy packages/app and have no NFS equivalent yet. As a forward-compatible step I switched the component tests to `@backstage/frontend-test-utils`

zdrapela: Cool, thank you!

zdrapela (approved):
After the update, /lgtm

PR #4865, opened 2026-05-20
test(app): raise packages/app UI coverage with component and unit tests (RHIDP-13853)

gustavolira:
## Description

Implements **[RHIDP-13853](https://redhat.atlassian.net/browse/RHIDP-13853)** — raises `packages/app` UI coverage by adding Layer 3 component/unit tests for previously-untested modules that carried the most uncovered lines (identified via Codecov). Complements [RHIDP-13235](https://redhat.atlassian.net/browse/RHIDP-13235) (which only covers behaviors mirrored by existing UI E2E specs).

## Tests added (target → coverage)

| Module | Before | After (lines) |
|--------|--------|---------------|
| `api/LearningPathApiClient.ts` | 13% | 100% |
| `components/Root/MenuIcon.tsx` | 21% | 100% |
| `components/ErrorPages/errorButtons/GoBackButton.tsx` | 0% | 100% |
| `components/ErrorPages/errorButtons/ContactSupportButton.tsx` | 0% | 100% |
| `components/ErrorPages/ErrorPage.tsx` | 0% | 100% |
| `components/SignInPage/SignInPage.tsx` | 0% | 100% |
| `components/Root/ResizableDrawer.tsx` | 12% | ~85% |

What the tests assert:

- **LearningPathApiClient** — URL building (default + configured proxy path), bearer-token header presence/absence, non-ok error path.
- **MenuIcon** — empty / registered system icon / inline-SVG / URL / material-name branches.
- **GoBackButton** — navigates back when history exists, renders nothing otherwise.
- **ContactSupportButton** — prop > configured URL > default fallback.
- **ErrorPage** — status/message/info, contact-support action, stack trace, no go-back for non-404.
- **ResizableDrawer** — children render while open, resize listeners registered when resizable, resize drag reports the new width.
- **SignInPage** — proxied vs local provider selection, guest only in development, default-to-github when unset/unknown (core-components widgets stubbed to isolate the selection logic).

## Out of scope

- `Root.tsx` (~130 uncovered lines) is the largest remaining target but high-effort (full sidebar nav) — left as a follow-up.
- `DynamicRoot.tsx` reports 0% despite having a test (lazy-import coverage-instrumentation artifact) — separate investigation.
- Entrypoints, trivial icon exports, and `apis.ts` factory wiring — low value.

## Testing

- [x] 25 tests pass across 7 suites
- [x] `tsc`, lint and prettier pass
- [x] New tests contribute to the Codecov `rhdh` flag

### DIFF
--- a/packages/app/src/components/ErrorPages/errorButtons/GoBackButton.test.tsx
+++ b/packages/app/src/components/ErrorPages/errorButtons/GoBackButton.test.tsx
@@ -1,5 +1,3 @@
-import { MemoryRouter } from 'react-router-dom';
-
 import { render, screen } from '@testing-library/react';
 import { userEvent } from '@testing-library/user-event';
 
@@ -7,6 +5,7 @@ import { GoBackButton } from './GoBackButton';
 
 const mockNavigate = jest.fn();
 
+// useNavigate is mocked, so no router context (MemoryRouter) is needed.
 jest.mock('react-router-dom', () => ({
   ...jest.requireActual('react-router-dom'),
   useNavigate: () => mockNavigate,
@@ -22,15 +21,15 @@ const setHistoryLength = (length: number) =>
     configurable: true,
   });
 
-const renderButton = () =>
-  render(
-    <MemoryRouter>
-      <GoBackButton />
-    </MemoryRouter>,
-  );
+const renderButton = () => render(<GoBackButton />);
 
 describe('GoBackButton', () => {
-  afterEach(() => jest.clearAllMocks());
+  const originalHistoryLength = window.history.length;
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    setHistoryLength(originalHistoryLength);
+  });
 
   it('navigates back when there is history to go back to', async () => {
     setHistoryLength(3);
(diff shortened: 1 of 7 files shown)


packages/app/src/components/ErrorPages/errorButtons/GoBackButton.test.tsx
kadel, line 1: Do we really need to tests for button that just calls `navigate(-1);`?

This feels like writing tests only for sake of writing tests and increasing coverage numbers rather than providing actual value.

kadel: It was already written so we can keep it. But I don't think that it makes much sense to test functions like this. 
`ContactSupportButton` and `ErrorPage` are similar stories.

gustavolira: > Do we really need to tests for button that just calls `navigate(-1);`?
> 
> This feels like writing tests only for sake of writing tests and increasing coverage numbers rather than providing actual value.

The real value there is the conditional rendering, not the click itself

gustavolira: > It was already written so we can keep it. But I don't think that it makes much sense to test functions like this. `ContactSupportButton` and `ErrorPage` are similar stories.

Agree they're simple components, but `ContactSupportButton` has a 3-level fallback (prop, config, default URL) that can break silently, and `ErrorPage` has conditional logic worth covering. Not strictly required, but I think they add some value.

PR #4866, opened 2026-05-21
test(app): restore the skipped DynamicRoot test suite (RHIDP-3611)

gustavolira:
## Description

Restores the `DynamicRoot` component test suite, which was `describe.skip`'d during the Backstage 1.29 upgrade (commit `8f96a153`, Aug 2024) and left disabled. As a result `DynamicRoot.tsx` reported **0% coverage** despite having a 29-test suite. Addresses **[RHIDP-3611](https://redhat.atlassian.net/browse/RHIDP-3611)** (which was closed prematurely — the skip and TODO were never removed).

This is the highest-leverage coverage item found while analyzing Codecov: ~205 instrumented lines that already had tests but counted as 0%.

## Root cause (two post-1.29 drifts)

1. **Unbound external route.** Backstage 1.29 throws on unbound *required* external routes at `createApp`. The fixture's `fooPluginTarget` declares an external route that is only bound in the route-binding tests, so every other test threw. Marked the fixture's external route `optional: true`.
2. **Framework translation API + analytics registration.** `DynamicRoot` now always injects a `core.translation` API. The API-factory/analytics tests must exclude framework-injected APIs when counting dynamic APIs (now via a single `getDynamicApis` helper), and an analytics extension registers under `analyticsApiRef`, not the plugin's own id. Also added the missing `fooPluginAnalyticsApi` fixture export.

### Trade-off on the optional route

Marking the fixture route `optional: true` is the contained fix: the binding tests still bind it and verify the map, just for an optional route. The faithful-but-invasive alternative would be to add a `routeBindings` entry to every non-binding test's config so the route stays required and bound everywhere. The optional approach reflects that a dynamically loaded plugin's external route may legitimately be unbound; it does not paper over the 1.29 strictness (the binding tests still exercise binding).

## Result

- The suite is un-skipped; **all 29 tests run and pass**.
- `DynamicRoot.tsx` coverage: **0% → ~75%** (lines).
- No production code changed — test-only.

## Follow-up (out of scope here)

- `ScalprumRoot.tsx` remains 0%: this suite mocks `ScalprumProvider`, so `ScalprumRoot` is never exercised. Restoring its coverage is a separate task.

## Testing

- [x] 29/29 tests pass
- [x] `tsc`, lint and prettier pass

### DIFF
--- a/packages/app/src/components/DynamicRoot/DynamicRoot.test.tsx
+++ b/packages/app/src/components/DynamicRoot/DynamicRoot.test.tsx
@@ -5,6 +5,7 @@ import * as appDefaults from '@backstage/app-defaults';
 import { Entity } from '@backstage/catalog-model';
 import { AppRouteBinder, defaultConfigLoader } from '@backstage/core-app-api';
 import {
+  analyticsApiRef,
   createApiFactory,
   createApiRef,
   createExternalRouteRef,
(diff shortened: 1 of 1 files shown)


schultzp2020 (approved):
/lgtm

PR #4868, opened 2026-05-21
fix(ci): add operator health check before applying Backstage CR

zdrapela:
## Problem

The [GKE operator nightly job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-redhat-developer-rhdh-main-e2e-gke-operator-nightly/2057341085121777664) failed because the RHDH operator controller-manager pod disappeared after installation. When the Backstage CR was applied, there was no operator to reconcile it, so the deployment was never created. The 5-minute wait timed out with:

```
Timeout waiting for: Backstage deployment created by operator
Backstage deployment not created after 5 minutes
Checking operator logs...
No resources found in rhdh-operator namespace.
```

No E2E tests ran at all.

## Root Cause

There was no health check verifying the operator controller-manager pod was actually running before applying the Backstage CR. The CRD could become available while the operator pod was still starting, crashing, or being evicted.

## Changes

**`.ci/pipelines/install-methods/operator.sh`**:
- Added `k8s_wait::deployment` in `prepare_operator()` to verify the controller-manager pod is running and ready after CRD becomes available (5-minute timeout)
- Added `k8s_wait::deployment` in `deploy_rhdh_operator()` before applying the Backstage CR as a defense-in-depth check (2-minute timeout)
- Enhanced `_operator_debug_info()` to show operator pod status and namespace events

**`.ci/pipelines/jobs/gke-operator.sh`**:
- Aligned `prepare_operator` retry count to `"3"` (matching AKS and EKS jobs which already use 3 retries)

## Verification

Tested on the GKE cluster (`gke-us-central1-c-standard-ci-rhdh`) with a real RHDH operator deployment:
- **Positive case**: Correctly detects running operator pod (phase=Running, ready=True)
- **Negative case**: Correctly detects missing operator pod (scaled deployment to 0)
- **Recovery case**: Correctly detects operator recovery after scale-up

Initial implementation used `"controller-manager"` as the pod name pattern for `k8s_wait::deployment`, but cluster testing revealed the pod is named `rhdh-operator-<hash>`, not `*controller-manager*`. Fixed to use `${OPERATOR_MANAGER}` (= `rhdh-operator`) which correctly matches the pod name prefix.

### DIFF
--- a/.ci/pipelines/install-methods/operator.sh
+++ b/.ci/pipelines/install-methods/operator.sh
@@ -56,6 +56,18 @@ prepare_operator() {
 
   # Wait for Backstage CRD to be available after operator installation
   k8s_wait::crd "backstages.rhdh.redhat.com" 300 10 || return 1
+
+  # Verify the operator controller-manager pod is running and ready.
+  # The CRD can appear before the operator pod is fully ready, so this
+  # ensures the controller is actually available to reconcile Backstage CRs.
+  # Uses label selector to target the controller-manager pod precisely.
+  log::info "Waiting for operator controller-manager pod to be ready in namespace '${OPERATOR_MANAGER}'..."
+  if ! oc wait -n "${OPERATOR_MANAGER}" pod -l control-plane=controller-manager \
+    --for=condition=Ready --timeout=300s 2> /dev/null; then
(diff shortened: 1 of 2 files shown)


zdrapela:
not needed anymore

PR #4924, opened 2026-06-08
[release-1.10] fix(e2e): fix showcase-runtime init container crash and improve logging [/fix-e2e]

zdrapela:
Jira: https://redhat.atlassian.net/browse/RHDHBUGS-3333

## Problem

The `install-dynamic-plugins` init container in the `showcase-runtime` deployment crashes with exit code 1 in all showcase-runtime E2E tests (`periodic-ci-redhat-developer-rhdh-release-1.10-e2e-ocp-helm-nightly`).

**Root cause:** The RHDH 1.10 chart enables lightspeed by default, injecting plugins with `{{inherit}}` version syntax that requires `dynamic-plugins.default.yaml`. The custom `initContainers` override in `values-showcase-postgres.yaml` lacked the `CATALOG_INDEX_IMAGE` env var needed to extract this file from the catalog index OCI image. Without it, `{{inherit}}` cannot resolve, causing `InstallException`.

## Fix

1. **Remove the stale custom `initContainers` override** — the chart's default init container has all required env vars (`CATALOG_INDEX_IMAGE`, `CATALOG_ENTITIES_EXTRACT_DIR`, `MAX_ENTRY_SIZE`) and proper volume mounts.
2. **Add missing volumes** — `extensions-catalog` and `temp` emptyDir volumes needed by the chart's default init container and main container.
3. **Fix `kube-client.ts` init container log capture** — extract the failing container name from the error message instead of hardcoding `backstage-backend`, and include `initContainers` in the fallback container list.

## Verification

- Deployed `showcase-runtime` on OCP cluster with the fixed values file
- Both init containers (`install-dynamic-plugins`, `lightspeed-rag-init`) exit 0
- Pod reaches 2/2 Running with 0 restarts
- Deployment restart succeeds (the step that was failing in all 8 tests)
- Lightspeed remains enabled (chart default)

## Affected tests

- `verify-tls-config-with-external-azure-db.spec.ts` (4 tests)
- `verify-tls-config-with-external-rds.spec.ts` (4 tests)

### DIFF
--- a/.ci/pipelines/resources/postgres-db/values-showcase-postgres.yaml
+++ b/.ci/pipelines/resources/postgres-db/values-showcase-postgres.yaml
@@ -46,6 +46,10 @@ upstream:
     extraVolumeMounts:
       - mountPath: /opt/app-root/src/dynamic-plugins-root
         name: dynamic-plugins-root
+      - mountPath: /extensions
+        name: extensions-catalog
+      - mountPath: /tmp
+        name: temp
       - mountPath: /opt/app-root/src/postgres-crt.pem
         name: postgres-crt # inject certificate secret to Backstage cont.
         subPath: postgres-crt.pem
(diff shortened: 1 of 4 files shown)


Fortune-Ndlovu (approved):
/lgtm

PR #4937, opened 2026-06-09
feat: add enabled field in DP config

hopehadfield:
## Description

Adds the `enabled` field with backwards compatibility

See https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2577 for relevant change in DPDY generation 

- Fixes [RHIDP-11988D](https://redhat.atlassian.net/browse/RHIDP-11988)

## PR acceptance criteria

- [X] Unit Tests are updated and passing
- [X] Documentation is updated if necessary (requirement for new features)


scripts/install-dynamic-plugins/src/types.ts
```diff
@@ -86,3 +96,48 @@ export function effectivePullPolicy(plugin: { pullPolicy?: PullPolicy; package:
...
+
+  if (hasEnabled && hasDisabled) {
+    warn?.(
+      `WARNING: Plugin ${plugin.package} specifies both 'enabled' and 'disabled'. ` +
+        `The 'enabled' field takes precedence; please use only 'enabled'.`,
+    );
+    return !plugin.enabled;
+  }
+  if (hasEnabled) return !plugin.enabled;
+  if (hasDisabled) return plugin.disabled === true;
```
Zaperex, line 141: We should also add an explicit deprecation warning log if they use `disabled` instead of `enabled`

hopehadfield: Just noted in the epic description:

> "We are not strictly deprecating disabled yet (i.e., we won't spam logs with "deprecated" warnings), just supporting the alternative."

So actually I should remove the deprecated tag entirely

Zaperex (approved):
/lgtm

PR #4967, opened 2026-06-17
feat(e2e): cluster-free plugin sanity check for the catalog index (RHIDP-13508)

gustavolira:
## Summary

Plugin sanity checking for the RHDH catalog index, implementing [RHIDP-13508](https://redhat.atlassian.net/browse/RHIDP-13508).

Validates that every plugin the catalog index declares actually **loads in the real RHDH backend**, with no cluster deployment and no product image. Building on the cluster-free harness (#5005), Playwright boots `packages/backend` from source — the product's own Backstage line and `dynamicPluginsFeatureLoader` — and the spec compares the installed set against `/api/dynamic-plugins-info/loaded-plugins`, failing with the exact list of installed-but-not-loaded plugins.

The index is selected by `CATALOG_INDEX_IMAGE` (branch-matched nightly index by default; RC index via Gangway `--catalog-index-image`).

## How it works

- **`local-harness/catalog-index-refs.sh`** — extracts `dynamic-plugins.default.yaml` from the index image and lists every declared package (uncommented entries only), applying `plugin-sanity-excludes.txt`. Layers are walked **top-down**: OCI layers are ordered base-first, so an index rebuilt as an overlay would otherwise yield a stale copy of the file. A missing excludes file is fatal rather than silently disabling every exclusion.
- **`local-harness/populate-catalog-index.sh`** — wraps those refs into an install config (the index ships packages `disabled: true`, so a plain `includes` would install nothing) and installs via the shared `populate.sh`, which gained an optional config argument.
- **`playwright.plugin-sanity.config.ts`** — dedicated request-only config; no browser, no frontend dev server.

## Guarding against a green-but-empty run

The spec would otherwise only assert *installed ⊆ loaded*, which stays green when the install silently underran. So:

- `populate-catalog-index.sh` records the index image and resolved `oci://` package count in `dynamic-plugins-root/.catalog-index-refs`
- the spec asserts the installed count matches it, and that the recorded image matches `CATALOG_INDEX_IMAGE`
- `plugin-sanity` gets its own `globalSetup` requiring that breadcrumb, so a `dynamic-plugins-root` left over from the curated `populate.sh` no longer satisfies the guard

## Failure reporting

A broken plugin takes the whole backend down by design, so the culprit is made obvious: `testing::report_plugin_startup_failures` scans pod logs (current + previous containers, so CrashLoopBackOff is covered) and prints a `PLUGIN STARTUP FAILURES` block saved as an artifact; the cluster-free wrapper greps its own run log for the same signatures.

## CI integration

Runs in `handle_ocp_nightly` → `run_sanity_plugins_check` via `testing::run_plugin_sanity_check`, which follows the `testing::run_tests` conventions (`test_run_tracker`, `save_overall_result`, gzipped JUnit to `SHARED_DIR`, report artifacts) so Slack reporting includes it like any other run. It does **not** alter the existing cluster sanity-plugins deployment.

## Local runs

```bash
CATALOG_INDEX_IMAGE=quay.io/rhdh/plugin-catalog-index:next \
  ./e2e-tests/local-harness/populate-catalog-index.sh
yarn --cwd e2e-tests plugin-sanity
```

## Verification

- Against `:next` — 9 declared packages resolved, backend booted, all loaded.
- Unit tests cover the breadcrumb parser (`e2e-tests/unit/catalog-index-expectation.test.ts`).
- `oxlint`, `oxfmt`, `tsc`, `vitest`, `shellcheck` and `.ci` prettier clean.

## Scope note

An earlier revision of this PR also made the **cluster** sanity-plugins deployment derive its enabled plugin set from the same index, replacing the hand-maintained list. That change has a very different risk profile — it swaps a working curated set for a fully derived one, and a single bad plugin fails the whole `showcase-sanity-plugins` project — so it has been split into a follow-up PR and is **not** included here. Branch: `gustavolira:sanity-plugins-cluster-dynamic`.

## Jira

- Implements: [RHIDP-13508](https://redhat.atlassian.net/browse/RHIDP-13508)
- Part of epic: [RHIDP-13496](https://redhat.atlassian.net/browse/RHIDP-13496) (Test Layer Infrastructure)


gustavolira:
/review
--pr_reviewer.inline_code_comments=true
-i

rostalan:
Hey there, the first thing I notice is the imo excessive amount of comments (cursor counted a third of the additions are pure comments). Reading through some, they sometimes describe very obvious code structures and add very little value. 
Comments should (imho) be reserved for situations where an explanation is explicitly needed, for example if some non-obvious intent or potentially confusing syntax, not to restate what the code already says. (not to mention that reading very thoroughly commented files consumes more tokens and they are harder to read through when reviewing...). 

Recommendations from cursor: 

Assessment
Agree there is excess, but not all of it is pointless.

Worth keeping (non-obvious traps / failure modes):

- Top-down OCI layer walk (stale overlay copy)
- Green-but-empty / breadcrumb .catalog-index-refs rationale
- Missing excludes file must be fatal
- Number("") === 0 trap
- Throw vs soft-return would pass the test
- yarn --cwd cwd-leak note
- Exclusion rationales in plugin-sanity-excludes.txt (especially after the digest-claim retraction commit)

Excess / low value:

1. Same architecture story repeated 4–5 times — PR body, plugin-dynamic-loading.spec.ts (22-line JSDoc), playwright.plugin-sanity.config.ts (22-line JSDoc), testing::run_plugin_sanity_check header, ocp-nightly.sh call-site. One canonical place (README or config) is enough.
2. // Step 1 … // Step 6 in the spec — narrates what the next few lines already say. The “partial install would pass” and “throw not return” notes are the only bits worth keeping.
3. Restating the next if: // Must have at least one bundle format, // Modern dist-scalprum format requires…
4. Trivial JSDoc: e.g. validateFrontendBundles (“Validate bundle artifacts for a list…”), file-title “Plugin Loader Utilities”.
5. Usage blocks duplicated across script headers + README + PR description.

gustavolira:
> Hey there, the first thing I notice is the imo excessive amount of comments (cursor counted a third of the additions are pure comments). Reading through some, they sometimes describe very obvious code structures and add very little value. Comments should (imho) be reserved for situations where an explanation is explicitly needed, for example if some non-obvious intent or potentially confusing syntax, not to restate what the code already says. (not to mention that reading very thoroughly commented files consumes more tokens and they are harder to read through when reviewing...).
> 
> Recommendations from cursor:
> 
> Assessment Agree there is excess, but not all of it is pointless.
> 
> Worth keeping (non-obvious traps / failure modes):
> 
> * Top-down OCI layer walk (stale overlay copy)
> * Green-but-empty / breadcrumb .catalog-index-refs rationale
> * Missing excludes file must be fatal
> * Number("") === 0 trap
> * Throw vs soft-return would pass the test
> * yarn --cwd cwd-leak note
> * Exclusion rationales in plugin-sanity-excludes.txt (especially after the digest-claim retraction commit)
> 
> Excess / low value:
> 
> 1. Same architecture story repeated 4–5 times — PR body, plugin-dynamic-loading.spec.ts (22-line JSDoc), playwright.plugin-sanity.config.ts (22-line JSDoc), testing::run_plugin_sanity_check header, ocp-nightly.sh call-site. One canonical place (README or config) is enough.
> 2. // Step 1 … // Step 6 in the spec — narrates what the next few lines already say. The “partial install would pass” and “throw not return” notes are the only bits worth keeping.
> 3. Restating the next if: // Must have at least one bundle format, // Modern dist-scalprum format requires…
> 4. Trivial JSDoc: e.g. validateFrontendBundles (“Validate bundle artifacts for a list…”), file-title “Plugin Loader Utilities”.
> 5. Usage blocks duplicated across script headers + README + PR description.

Agreed 100%. Thanks for the review, I will remove some comments

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+test.describe("Plugin Dynamic Loading", () => {
+  test.beforeAll(() => {
+    test.info().annotations.push({
+      type: "component",
+      description: "plugins",
+    });
+  });
+
+  test("all catalog index plugins load in the RHDH backend", async ({ request }) => {
```
rostalan: from e2e-tests/CONTRIBUTING.MD - §6 General Rules II: "Each test has to test one and only one thing."
this should be split into multiple test cases to give a clear failure reason.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      ).toBe(process.env.CATALOG_INDEX_IMAGE);
+    }
+    expect(
+      expected.length,
+      `installed plugin count should match the ${indexExpectation.expectedOciPackages} ` +
+        `oci:// packages declared by ${indexExpectation.image}`,
+    ).toBe(indexExpectation.expectedOciPackages);
+
+    // loaded-plugins requires user credentials, so authenticate as guest.
+    const refresh = await request.get("/api/auth/guest/refresh", {
```
rostalan: The spec file makes direct request.get calls and consumes scanInstalledPlugins, readCatalogIndexExpectation, parseLoadedPluginNames, and parseRefreshToken inline. 
The e2e standard requires all such interactions to be encapsulated inside a POM class; the raw Playwright request fixture and utility helpers belong behind a named page/service object, not inside the spec body.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  test.beforeAll(() => {
+    test.info().annotations.push({
+      type: "component",
+      description: "plugins",
+    });
+  });
+
+  test("all catalog index plugins load in the RHDH backend", async ({ request }) => {
+    const manifest = scanInstalledPlugins(dynamicPluginsRoot);
+    console.log(
```
rostalan: The spec contains console.log blocks for test state updates. This could be done better via Playwright [test.step()](https://playwright.dev/docs/api/class-test#test-step) labels and structured annotations.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    // the test as PASSED without having validated anything.
+    const indexExpectation = readCatalogIndexExpectation(dynamicPluginsRoot);
+    if (indexExpectation === null) {
+      throw new Error(
+        "dynamic-plugins-root was not populated from the catalog index - run " +
+          "local-harness/populate-catalog-index.sh with CATALOG_INDEX_IMAGE set",
+      );
+    }
+
+    console.log(`[TEST] Catalog index: ${indexExpectation.image}`);
```
rostalan: as above

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+    expect(refresh.status(), "guest auth refresh should succeed").toBe(200);
+    const token = parseRefreshToken(await refresh.json());
+
+    const response = await request.get("/api/dynamic-plugins-info/loaded-plugins", {
+      headers: { Authorization: `Bearer ${token}` },
+    });
+    expect(response.status(), "loaded-plugins endpoint should respond").toBe(200);
+    const loadedNames = parseLoadedPluginNames(await response.json());
+    console.log(`[TEST] Backend reports ${loadedNames.size} loaded dynamic plugins`);
```
rostalan: s above

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      for (const { plugin, error } of frontendErrors) {
+        console.log(`   - ${plugin.name}: ${error}`);
+      }
+    }
+    expect(
+      frontendErrors.map(({ plugin, error }) => `${plugin.name}: ${error}`),
+      "every frontend plugin should ship valid bundle artifacts",
+    ).toEqual([]);
+
+    console.log("\n[TEST] Summary:");
```
rostalan: as above

e2e-tests/playwright/utils/installed-plugins.ts
```diff
@@ -0,0 +1,201 @@
...
+    throw new Error(`Malformed ${refsPath}: expected 'image=' and 'expected_oci_packages=' lines`);
+  }
+  return { image, expectedOciPackages };
+}
+
+/**
+ * Frontend plugins ship either the modern dist-scalprum/ bundle (which carries
+ * a plugin-manifest.json) or the legacy dist/remoteEntry.js one.
+ */
+export function validateFrontendBundle(plugin: PluginEntry): string | null {
```
rostalan, line 192: Story states: "Frontend plugins: navigate to the plugin route and assert no JavaScript errors"
This only checks the dirs exist on disk, no plugin route is inspected.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      console.log(`\n[TEST] Installed but not loaded (${notLoaded.length}):`);
+      for (const plugin of notLoaded) {
+        console.log(`   - ${plugin.name}@${plugin.version} (${plugin.dirName})`);
+      }
+      console.log("\nLoaded plugins reported by the backend:");
+      for (const name of [...loadedNames].toSorted()) {
+        console.log(`   - ${name}`);
+      }
+    }
+    expect(
```
rostalan: Story states that the check should be done via asserting each plugins api and expecting a healthy response. Here, we only check if the plugin is in a list of loaded plugins. Is there a reason for the difference?

e2e-tests/playwright/utils/installed-plugins.ts
```diff
@@ -0,0 +1,201 @@
...
+ * local-harness/populate-catalog-index.sh, which records the index image and
+ * how many oci:// packages that run resolved.
+ *
+ * Without it the sanity check only asserts "installed ⊆ loaded", which stays
+ * green when the install silently underran (registry hiccup, over-broad
+ * exclude pattern, or dynamic-plugins-root left over from the curated
+ * `populate.sh`). Returns null when absent so a local curated run fails with a
+ * clear message rather than a type error.
+ */
+export function readCatalogIndexExpectation(installDir: string): CatalogIndexExpectation | null {
```
rostalan, line 125: architectural review found this: 
"readCatalogIndexExpectation returns T | null. Global setup already throws if the .catalog-index-refs file is missing. The spec then calls the same function and has its own null-throw — dead code under this Playwright config, with a different error message.

Friction: two places “own” the same invariant; maintainers can’t tell which message is authoritative.

Deepen: a requireCatalogIndexExpectation() that always returns T (or throws). Setup and spec both use it; drop the null branch in the test."

.github/workflows/e2e-cluster-free.yaml
```diff
@@ -94,3 +94,87 @@ jobs:
...
+              exit 1
+            fi
+            image="${OVERRIDE}"
+          elif [[ "${BRANCH}" == release-* ]]; then
+            image="quay.io/rhdh/plugin-catalog-index:${BRANCH#release-}"
+          else
+            image="quay.io/rhdh/plugin-catalog-index:next"
+          fi
+          echo "image=${image}" >> "$GITHUB_OUTPUT"
+          echo "Catalog index: ${image}"
```
zdrapela: Not critical, but it would be nice to have the digest printed out alongside the tag. That would help with tracing where issues started happening etc.

.github/workflows/e2e-cluster-free.yaml
```diff
@@ -94,3 +94,87 @@ jobs:
           name: playwright-report-legacy-local
           path: e2e-tests/playwright-report-legacy-local
           retention-days: 7
+
+  plugin-sanity:
+    name: plugin-sanity
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
```
zdrapela: nit: 45 min timeout is quite high for a job that typically runs 5 minutes. Do you expect it to grow, or have edge cases where it would run for longer?

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
zdrapela, line 1: I'm not convinced we should mix the cluster-free test files with the regular E2E. Now we need to exclude it for every catch-all Playwright project in `e2e-tests/playwright.config.ts` and I imagine the number of tests like this will grow. What about finding a better home for it?

e2e-tests/local-harness/app-config.plugin-sanity.yaml
zdrapela, line 1: Should this dummy file live in the repo root if it's only for E2E? Same applies for `app-config.local-e2e.yaml` as both should live in the same home.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Done. Split into five tests, one per concern. Before, one test had six assertions across four concerns, so a failure never told you what actually broke.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Partly done. Auth and both endpoints now sit behind a DynamicPluginsApi service object, same shape as the existing RhdhRbacApi.

Left the plugin-loader helpers alone since they're filesystem/parsing, not HTTP, and they're unit-tested directly. Also worth noting the POM rule in CONTRIBUTING.MD is scoped to uiHelper, and the other request-only spec (instance-health-check) calls request.get inline too. Can go further if you want, but that's a wider change than this PR.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Partly done. Dropped the summary block and the progress noise, kept the failure dumps that list which plugins didn't load, since that's the bit you want in the log when it's red. Splitting the test removed most of the narration anyway.

Didn't switch to test.step: it has zero uses in the repo, and console.log is in 11 of 29 specs. Doing it in one spec just makes this file the odd one out. Happy to do it suite-wide as its own PR if we want that as the style.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Same as above.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Same as above.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Same as above.

e2e-tests/playwright/utils/installed-plugins.ts
gustavolira: Swapped it for a real per-plugin check that works without a browser. The spec now asserts every dist-scalprum plugin is actually served at /api/scalprum/plugins. That catches something the disk check couldn't, since scalprum-backend warns and skips a plugin with a broken bundle while it still shows up in loaded-plugins.

Didn't do route navigation: it needs a frontend + browser, and it's undefined for much of the index since a route only exists if the plugin declares dynamicRoutes. If you want it, showcase-sanity-plugins is the better home, it already has a deployment and a browser.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: I don't think this one works as written, regardless of the story's age:

- only 1 of 13 installed backend plugins has a /health route
- backend-plugin-module plugins mount no router at all, so there's no URL to hit
- /api/<pluginId>/ can't discriminate, 404 for both "not registered" and "no such sub-route"

Also loaded-plugins filters failures server-side, so absence from that list is the failure signal. Written up on RHIDP-13508 so the deviation lives with the story.

e2e-tests/playwright/utils/installed-plugins.ts
gustavolira: Good catch, done exactly as suggested. Added requireCatalogIndexExpectation() that throws instead of returning null. Global setup and the spec both use it now, so one owner and one message. Dead branch gone.

.github/workflows/e2e-cluster-free.yaml
gustavolira: Added.

.github/workflows/e2e-cluster-free.yaml
gustavolira: Fair, dropped to 20. Runs take 5-7 min; 45 was a guess from before there was any data.

e2e-tests/playwright/e2e/plugin-dynamic-loading.spec.ts
gustavolira: Agreed. Moved to playwright/cluster-free/ and dropped all four testIgnore entries, since anything outside testDir is invisible to the cluster projects for free.

e2e-tests/local-harness/app-config.plugin-sanity.yaml
gustavolira: It shouldn't, and it was actually leaking. Containerfile:206 copies app-config*.yaml and .dockerignore didn't cover this one, so a test-only config was shipping in the image. Moved to e2e-tests/local-harness/, which .dockerignore already excludes.

e2e-tests/local-harness/populate-catalog-index.sh
```diff
@@ -0,0 +1,68 @@
...
+  # the index's own list, so simply leaving one out installs it anyway.
+  if [[ -n "$excluded_refs" ]]; then
+    while read -r ref; do
+      echo "  - package: \"$ref\""
+      echo "    disabled: true"
+    done <<< "$excluded_refs"
+  fi
+} > "$workdir/dynamic-plugins.catalog-index.yaml"
+
+echo "Enabling $(echo "$refs" | grep -c '^oci://') OCI package(s) from ${CATALOG_INDEX_IMAGE}"
```
rostalan: Under set -e, grep -c '^oci://' exits 1 when refs exist but none are oci:// (all OCI excluded), so the script aborts before a clear “no OCI packages” failure / breadcrumb write

rostalan (approved):
looks good, just one more nit.

e2e-tests/local-harness/plugin-sanity-excludes.txt
```diff
@@ -36,3 +39,18 @@ red-hat-developer-hub-backstage-plugin-orchestrator-backend(@|:)
...
+# registry.access.redhat.com alike. Note these are tag-style refs
+# (:1.11.0--x.y.z); every digest-pinned ref the index declares does resolve
+# anonymously via the CLI's quay.io fallback, so this is about publication, not
+# about the digest-lookup trap described above.
+#
+# Excluding them costs real coverage - homepage in particular is a core plugin.
+# Re-check periodically with:
+#   skopeo inspect --no-creds docker://quay.io/rhdh/<name>:<tag>
+# and delete the entry the moment it resolves.
+red-hat-developer-hub-backstage-plugin-homepage(@|:)
```
zdrapela, line 56: Can we include a tracking `TODO` with a ticket number with reasoning of why they are disabled and to re-enable them? It would be a good rule to follow, so we don't forget to unskip them.
`homepage` is tracked here e.g.: https://redhat.atlassian.net/browse/RHDHBUGS-3515

e2e-tests/playwright.plugin-sanity.config.ts
```diff
@@ -0,0 +1,42 @@
...
+  workers: 1,
+  reporter: harnessReporters("plugin-sanity"),
+  use: {
+    baseURL: backendUrl,
+  },
+  webServer: [
+    {
+      // Dummy values for plugins that abort the backend when their config is
+      // missing; passed last so it wins over generated plugin defaults.
+      ...backendWebServer(["../../e2e-tests/local-harness/app-config.plugin-sanity.yaml"]),
```
zdrapela: What is the root? Isn't this equivalent?
```suggestion
      ...backendWebServer(["../local-harness/app-config.plugin-sanity.yaml"]),
```

e2e-tests/local-harness/populate-catalog-index.sh
gustavolira: It doesn't abort there, only a bare assignment trips set -e. Worse actually: it
ran through and wrote expected_oci_packages=0. Added an explicit guard anyway.

e2e-tests/local-harness/plugin-sanity-excludes.txt
gustavolira: Done. TODO(RHDHBUGS-3515) for homepage + intelligent-assistant, TODO(RHIDP-15961)
for the orchestrator pair

e2e-tests/playwright.plugin-sanity.config.ts
gustavolira: Not equivalent, paths resolve from the backend's cwd. Fair confusion though, so I
moved the ../../ into harnessConfigArgs and the call site now reads as you'd expect.

rostalan (approved):

PR #4979, opened 2026-06-22
[release-1.9] fix(e2e): fix orchestrator workflow deployment

rostalan:
## Description

On 1.9, orchestrator tests fail with timeout error, because workflow images do not exist.
This PR changes OSL channel from logic-operator-rhel8:alpha to logic-operator:stable and pins it to 1.37.2.
It also pins workflow images to 1.37.
And also unskips some skipped orchestrator jobs for 1.9.
And it also removes check for non-existing UI element that failed the tests ([RHDHBUGS-3363](https://redhat.atlassian.net/browse/RHDHBUGS-3363))

Fixes [RHDHBUGS-3383](https://redhat.atlassian.net/browse/RHDHBUGS-3383)
Also fixes [RHDHBUGS-3363](https://redhat.atlassian.net/browse/RHDHBUGS-3363)

## PR acceptance criteria

### DIFF
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -500,11 +500,19 @@ waitfor_crunchy_postgres_k8s_operator() {
 # Installs the OpenShift Serverless Logic Operator (SonataFlow) from OpenShift Marketplace
 # Use waitfor_serverless_logic_ocp_operator to wait for the operator to be ready
 install_serverless_logic_ocp_operator() {
+  log::info "Installing Serverless Logic Operator from channel stable with startingCSV logic-operator.v1.37.2"
   install_subscription logic-operator openshift-operators stable logic-operator redhat-operators openshift-marketplace logic-operator.v1.37.2
 }
 
 waitfor_serverless_logic_ocp_operator() {
   check_operator_status 300 "openshift-operators" "OpenShift Serverless Logic Operator" "Succeeded"
+  local osl_csv_info
+  osl_csv_info=$(oc get csv -n openshift-operators -l operators.coreos.com/logic-operator.openshift-operators -o jsonpath='{.items[0].metadata.name} {.items[0].spec.version}' 2> /dev/null || true)
+  if [[ -n "${osl_csv_info}" ]]; then
+    log::info "Installed OSL CSV/version: ${osl_csv_info}"
+  else
+    log::warn "Unable to determine installed OSL CSV/version from openshift-operators namespace"
+  fi
 }
 
 # Installs the OpenShift Serverless Operator (Knative) from OpenShift Marketplace
@@ -1557,8 +1565,10 @@ deploy_orchestrator_workflows() {
 
   rm -rf "${workflow_dir}"
   git clone "${workflow_repo}" "${workflow_dir}"
+  log::info "Using serverless-workflows ref: ${workflow_repo_ref}"
   git -C "${workflow_dir}" fetch --depth=1 origin "${workflow_repo_ref}"
   git -C "${workflow_dir}" checkout --detach "${workflow_repo_ref}"
+  log::info "Checked out serverless-workflows commit: $(git -C "${workflow_dir}" rev-parse --short HEAD)"
 
   if [[ "$namespace" == "${NAME_SPACE_RBAC}" ]]; then
     local pqsl_secret_name="postgres-cred"
@@ -1643,8 +1653,10 @@ deploy_orchestrator_workflows_operator() {
 
   rm -rf "${workflow_dir}"
   git clone --depth=1 "${workflow_repo}" "${workflow_dir}"
+  log::info "Using serverless-workflows ref: ${workflow_repo_ref}"
   git -C "${workflow_dir}" fetch --depth=1 origin "${workflow_repo_ref}"
   git -C "${workflow_dir}" checkout --detach "${workflow_repo_ref}"
+  log::info "Checked out serverless-workflows commit: $(git -C "${workflow_dir}" rev-parse --short HEAD)"
 
   # Wait for backstage and sonata flow pods to be ready before continuing
   wait_for_deployment $namespace backstage-psql 15
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -1553,7 +1553,8 @@ get_previous_release_value_file() {
   fi
 }
 
-readonly SERVERLESS_WORKFLOWS_DEFAULT_REF="daeeee8dec16beab6d96a81774ef500081a2c2b0"
+: "${SERVERLESS_WORKFLOWS_DEFAULT_REF:=daeeee8dec16beab6d96a81774ef500081a2c2b0}"
+readonly SERVERLESS_WORKFLOWS_DEFAULT_REF
 
 checkout_serverless_workflows_ref() {
   local workflow_dir=$1
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -378,7 +378,7 @@ install_subscription() {
   package=$4          # Package name of the operator
   source_name=$5      # Name of the source catalog
   source_namespace=$6 # Source namespace (typically openshift-marketplace or olm)
-  starting_csv=$7     # Optional starting CSV to pin subscription install
+  starting_csv=${7:-} # Optional starting CSV to pin subscription install
   starting_csv_line=""
   if [[ -n "$starting_csv" ]]; then
     starting_csv_line="  startingCSV: $starting_csv"
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -500,8 +500,9 @@ waitfor_crunchy_postgres_k8s_operator() {
 # Installs the OpenShift Serverless Logic Operator (SonataFlow) from OpenShift Marketplace
 # Use waitfor_serverless_logic_ocp_operator to wait for the operator to be ready
 install_serverless_logic_ocp_operator() {
-  log::info "Installing Serverless Logic Operator from channel stable with startingCSV logic-operator.v1.37.2"
-  install_subscription logic-operator openshift-operators stable logic-operator redhat-operators openshift-marketplace logic-operator.v1.37.2
+  local osl_starting_csv="${OSL_STARTING_CSV:-logic-operator.v1.37.2}"
+  log::info "Installing Serverless Logic Operator from channel stable with startingCSV ${osl_starting_csv}"
+  install_subscription logic-operator openshift-operators stable logic-operator redhat-operators openshift-marketplace "${osl_starting_csv}"
 }
 
 waitfor_serverless_logic_ocp_operator() {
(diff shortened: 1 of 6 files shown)


lholmquist:
@rostalan i recently push updated images for these four workflows and tagged them as `osl_1_38`

* [serverless-workflow-greeting](https://quay.io/repository/orchestrator/serverless-workflow-greeting)
* [serverless-workflow-sample-retry-test](https://quay.io/repository/orchestrator/serverless-workflow-sample-retry-test)
* [serverless-workflow-test-object-type-uiprops](https://quay.io/repository/orchestrator/serverless-workflow-test-object-type-uiprops)
* [fail-switch](https://quay.io/repository/orchestrator/fail-switch)


i'm not sure if those are the workflows that were timing out or not. i know there was another slack thread that mentioned those

rostalan:
> @rostalan i recently push updated images for these four workflows and tagged them as `osl_1_38`
> 
>     * [serverless-workflow-greeting](https://quay.io/repository/orchestrator/serverless-workflow-greeting)
> 
>     * [serverless-workflow-sample-retry-test](https://quay.io/repository/orchestrator/serverless-workflow-sample-retry-test)
> 
>     * [serverless-workflow-test-object-type-uiprops](https://quay.io/repository/orchestrator/serverless-workflow-test-object-type-uiprops)
> 
>     * [fail-switch](https://quay.io/repository/orchestrator/fail-switch)
> 
> 
> i'm not sure if those are the workflows that were timing out or not. i know there was another slack thread that mentioned those

Hey @lholmquist, thanks, 
I am not sure this helps here, since the release-1.9 branch is still using osl 1.36.1 ([via alpha channel of logic-operator-rhel8](https://github.com/redhat-developer/rhdh/blob/8f5b348c88f47bcf6c8d8e8afb738f09b31f2617/.ci/pipelines/utils.sh#L497)) and the decision was made yesterday to pin it to 1.37. 

It might however have an impact for main and 1.10, where we are currently resolving to osl 1.38 afaik. 
Not sure if there are any guards/warnings in place against using older workflow images (probably no?) and not sure how compatible the older images are with newer osl, since we keep seeing a lot of orchestrator related failures recently.

Will update this PR to pin the osl version to 1.37.2.

polasudo:
## CI Medic: Root Cause Analysis

The `pull-ci-redhat-developer-rhdh-release-1.9-e2e-ocp-helm` job failed before any tests ran.

**Root cause:** `.ci/pipelines/utils.sh` line 381 references bare `$7` (`starting_csv=$7`) but 4 of 5 callers of `install_subscription()` only pass 6 arguments. With `set -o nounset` (set in `openshift-ci-tests.sh` line 5), this is a fatal "unbound variable" error.

The crash happened during OpenShift Pipelines operator installation (`install_pipelines_operator` at line 968), which calls `install_subscription` with only 6 args.

**Callers affected:**
| Line | Caller | Args |
|------|--------|------|
| 457 | `install_crunchy_postgres_ocp_operator` | 6 (crashes) |
| 480 | `install_crunchy_postgres_k8s_operator` | 6 (crashes) |
| 503 | `install_serverless_logic_ocp_operator` | **7** (OK) |
| 513 | `install_serverless_ocp_operator` | 6 (crashes) |
| 968 | `install_pipelines_operator` | 6 (crashes) |

**Fix:** Change line 381 from:
```bash
starting_csv=$7
```
to:
```bash
starting_csv=${7:-}
```

This makes the 7th parameter optional with an empty default, which is the correct behavior since only the serverless logic operator needs CSV pinning.

e2e-tests/playwright/support/pages/orchestrator.ts
```diff
@@ -327,18 +330,31 @@ export class Orchestrator {
...
+      await expect
+        .poll(
+          async () => {
+            const completedVisible = await completedStatusIcon
+              .isVisible()
+              .catch(() => false);
+            const timestampVisible = await completionTimestamp
+              .isVisible()
+              .catch(() => false);
+            return completedVisible && timestampVisible;
```
gustavolira, line 350: This poll dropped the `SuccessOutlinedIcon` check the old code asserted (commit 6 removed it). With it gone, nothing here verifies the run actually succeeded - only that the "Completed" label and timestamp rendered. If the icon was just flaky, fair enough, but should we still assert success somewhere so a failed-but-labeled-Completed run can't slip through?

e2e-tests/playwright/e2e/plugins/orchestrator/failswitch-workflow.spec.ts
```diff
@@ -148,8 +150,37 @@ async function getHttpbinValue(ns: string): Promise<string | undefined> {
...
+        "sonataflow",
+        "failswitch",
+        "-o",
+        "jsonpath={.spec.podTemplate.container.env}",
+      ])
+    ).trim();
+    if (raw && raw !== "null") {
+      existing = JSON.parse(raw) as EnvEntry[];
+    }
+  } catch {
```
gustavolira: If this `oc get` throws, `existing` stays `[]` and the patch below replaces the container's entire `env` with just HTTPBIN, dropping everything else. The comment says best-effort, but a silent read failure here quietly wipes the env. Can we at least `console.warn` in the catch so it's visible, or bail out instead of patching with an empty base?

.ci/pipelines/utils.sh
```diff
@@ -1544,12 +1558,17 @@ deploy_orchestrator_workflows() {
   local release_name=$2
 
   local workflow_repo="https://github.com/rhdhorchestrator/serverless-workflows.git"
+  local workflow_repo_ref="${SERVERLESS_WORKFLOWS_REF:-daeeee8dec16beab6d96a81774ef500081a2c2b0}"
```
gustavolira: This ref-pinning block - including the default SHA `daeeee8...` - is duplicated verbatim in `deploy_orchestrator_workflows_operator` below. Worth pulling the default into a single `readonly SERVERLESS_WORKFLOWS_DEFAULT_REF` (and maybe a small `checkout_workflows_ref` helper) so the two can't drift. Side note: this function does a full `git clone` while the operator one uses `--depth=1` - since both immediately fetch the pinned SHA, the full clone just makes CI slower.

.ci/pipelines/utils.sh
```diff
@@ -1544,12 +1558,17 @@ deploy_orchestrator_workflows() {
...
   local workflow_dir="${DIR}/serverless-workflows"
   local failswitch_manifests="${workflow_dir}/workflows/fail-switch/src/main/resources/manifests/"
   local greeting_manifests="${workflow_dir}/workflows/greeting/manifests/"
 
   rm -rf "${workflow_dir}"
   git clone "${workflow_repo}" "${workflow_dir}"
+  log::info "Using serverless-workflows ref: ${workflow_repo_ref}"
+  git -C "${workflow_dir}" fetch --depth=1 origin "${workflow_repo_ref}"
```
gustavolira: These aren't guarded - if the fetch of the pinned SHA fails, the script keeps going and `checkout --detach` runs against whatever the clone left checked out, so tests silently run on the wrong workflow version. Given this PR is specifically about pinning that version, I'd fail fast here: `... || { log::error "failed to fetch workflows ref ${workflow_repo_ref}"; return 1; }`.

.ci/pipelines/utils.sh
```diff
@@ -494,11 +500,19 @@ waitfor_crunchy_postgres_k8s_operator() {
 # Installs the OpenShift Serverless Logic Operator (SonataFlow) from OpenShift Marketplace
 # Use waitfor_serverless_logic_ocp_operator to wait for the operator to be ready
 install_serverless_logic_ocp_operator() {
-  install_subscription logic-operator-rhel8 openshift-operators alpha logic-operator-rhel8 redhat-operators openshift-marketplace
+  log::info "Installing Serverless Logic Operator from channel stable with startingCSV logic-operator.v1.37.2"
+  install_subscription logic-operator openshift-operators stable logic-operator redhat-operators openshift-marketplace logic-operator.v1.37.2
```
gustavolira: The startingCSV `logic-operator.v1.37.2` is hard-coded, but the workflow ref further down honors `${SERVERLESS_WORKFLOWS_REF:-...}`. Same kind of version pin, two different override stories - mind making this `${OSL_STARTING_CSV:-logic-operator.v1.37.2}` too so a bump doesn't need a code change?

e2e-tests/playwright/e2e/plugins/orchestrator/greeting-workflow.spec.ts
```diff
@@ -6,7 +6,7 @@ import { skipIfJobName } from "../../../utils/helper";
 import { JOB_NAME_PATTERNS } from "../../../utils/constants";
 
 test.describe("Orchestrator greeting workflow tests", () => {
-  // TODO: https://issues.redhat.com/browse/RHDHBUGS-2184 fix orchestrator tests on Operator deployment
+  // skipping orchestrator tests on Operator deployment for 1.9
```
gustavolira: nit: this dropped the RHDHBUGS-2184 link the old comment had. Without it there's nothing tying the skip back to a tracking issue for when it can be re-enabled - can we keep the reference?

e2e-tests/playwright/e2e/plugins/orchestrator/workflow-all-runs-validations.spec.ts
```diff
@@ -6,7 +6,7 @@ import { skipIfJobName } from "../../../utils/helper";
 import { JOB_NAME_PATTERNS } from "../../../utils/constants";
 
 test.describe("Orchestrator Workflow Runs tests", () => {
-  // TODO: https://issues.redhat.com/browse/RHDHBUGS-2184 fix orchestrator tests on Operator deployment
+  // skipping orchestrator tests on Operator deployment for 1.9
```
gustavolira: nit: same as greeting-workflow - keeping the RHDHBUGS-2184 link here would make it clear why this stays skipped and when to revisit.

e2e-tests/playwright/e2e/plugins/orchestrator/failswitch-workflow.spec.ts
```diff
@@ -148,8 +150,37 @@ async function getHttpbinValue(ns: string): Promise<string | undefined> {
...
+        ns,
+        "get",
+        "sonataflow",
+        "failswitch",
+        "-o",
+        "jsonpath={.spec.podTemplate.container.env}",
+      ])
+    ).trim();
+    if (raw && raw !== "null") {
+      existing = JSON.parse(raw) as EnvEntry[];
```
gustavolira: nit: `as EnvEntry[]` trusts the jsonpath output blindly - if it's ever not an array, the `findIndex` below throws. An `Array.isArray(parsed)` guard before assigning would keep the best-effort intent.

e2e-tests/playwright/e2e/plugins/orchestrator/failswitch-workflow.spec.ts
```diff
@@ -166,28 +197,23 @@ async function patchHttpbin(ns: string, value: string): Promise<void> {
...
-    "-l",
-    "app.kubernetes.io/name=failswitch",
-    "--timeout=5s",
-  ];
-  await LogUtils.executeCommandWithRetries("oc", waitArgs, 5);
+    "rollout",
+    "status",
+    "deployment",
+    "failswitch",
+    "--timeout=60s",
```
gustavolira, line 232: nit: `--timeout=60s` here (vs the old 5s x 5 retries) could use a one-line note on why 60s - the surrounding orchestrator specs tend to justify their timeouts.

gustavolira (approved):

PR #5002, opened 2026-06-23
feat(e2e): strict oxlint, oxfmt, and type-aware linting

schultzp2020:
## Summary

PR **2** in the E2E modernization stack (after merged #5000).

- Migrate `e2e-tests/` from ESLint + Prettier + `tsc:check` to **oxlint** (type-aware + type-check) and **oxfmt**
- Enable **strict TypeScript** in `e2e-tests/tsconfig.json` and fix ~235 type errors across specs and utilities
- Add strict oxlint plugins (`import`, `node`, `promise`) with suspicious-as-error categories; promote Playwright and unsafe-* rules to error
- Wire Playwright lint rules via `eslint-plugin-playwright` and file naming via `eslint-plugin-check-file`
- Add `waitForConfigReconciled()` in `RHDHDeployment` and remove hard `waitForTimeout` calls from auth-provider specs
- Derive locale and audit-log types from `as const` tuples (single source of truth)
- Regression fixes from the lint migration: Playwright `testInfo` hook destructuring, audit-log parsing, card scoping, helper-based assertions
- Update `e2e-tests-lint.yaml` and lint-staged; add `yarn test:list` to CI so Playwright collection regressions fail in GHA

## Merge order

1. #5000 — CI hygiene *(merged)*
2. **This PR** — OXC toolchain (`feat/e2e-oxc-toolchain`)
3. #5003 — oxlint pedantic
4. #5004 — e2e hardening

## Test plan

- [x] `cd e2e-tests && yarn lint` — 0 errors
- [x] `cd e2e-tests && yarn fmt:check` — clean
- [x] `cd e2e-tests && yarn test:list` — Playwright collects all specs
- [x] CI `e2e-tests-lint` and Prow e2e jobs green

## Notes

- TypeScript 6.0.3 (TS 7 RC blocked by Yarn compat patch on Windows)
- `strict: true` in e2e-tests tsconfig; type-aware oxlint rules enforce safety without a separate `tsc:check` step

PR #5000, opened 2026-06-23
fix(ci): hygiene — working-directory, lint gaps, and cleanup

schultzp2020:
## Summary
- Fix `pr.yaml` dynamic-plugins steps to use `working-directory` instead of a no-op `cd`
- Add `lint:check` / `lint:fix` to `app-next` and `dynamic-plugins/_utils` so Turbo CI covers them
- Align GHA `setup-node` to v6.4.0 across e2e/bash/rulesync workflows
- Use `yarn lint-staged` in husky pre-commit
- Remove dead `jest` config and unused `eslint-plugin-filenames` from e2e-tests

## Test plan
- [ ] Verify `pr.yaml` dynamic-plugins job runs install/test/export from `dynamic-plugins/`
- [ ] `yarn lint:check --continue --affected` includes `app-next`
- [ ] `yarn lint:check` in `dynamic-plugins/_utils` succeeds

Made with [Cursor](https://cursor.com)

PR #5004, opened 2026-06-23
feat(e2e): harden browser lifecycle, smoke gate, and RHDH naming

schultzp2020:
## Summary
- Harden browser lifecycle with worker-scoped `rhdhPage` / `rhdhContext` fixtures, explicit `teardownBrowser()` in auth-provider suites, and `AuthProviderHarness` for shared deployment setup.
- Gate Playwright runs on RHDH readiness via `globalSetup` + `waitForRhdhReady()`, replace fixed sleeps with `pollUntil`, and enforce critical-only axe violations.
- Collapse shallow helper layers into focused page objects and semantic `ui-helper/*` modules; rename `BackstageShowcase` → `RhdhInstance`; separate E2E (`*.spec.ts`) from unit (`*.test.ts`) tests with scoped Oxlint plugins.
- Restore cluster-free compatibility: skip JSON `/healthcheck` polling on the legacy SPA base URL and fall back to legacy sidebar locators for `packages/app`.

PR **4** in the E2E modernization stack (after #5002 OXC toolchain and #5003 pedantic).

### Browser lifecycle and fixtures
```mermaid
graph TD
  Specs["E2E specs"] --> Fixtures["@support/coverage/test fixtures"]
  Fixtures --> Worker["rhdhPage / rhdhContext"]
  Worker --> Setup["setupBrowser()"]
  Worker --> Teardown["teardownBrowser()"]
  AuthSpecs["auth-provider specs"] --> Harness["AuthProviderHarness"]
  AuthSpecs --> Teardown
  Harness --> Deploy["RHDH deployment helpers"]
```

### Readiness gate and polling
```mermaid
graph TD
  GlobalSetup["global-setup.ts"] --> Health["waitForRhdhReady()"]
  GuestLogin["Common.loginAsGuest()"] --> AppReady["waitForAppReady()"]
  AppReady --> Health
  AppReady --> Load["waitForLoad()"]
  LegacyHarness["cluster-free legacy harness"] --> SkipHealth["skip JSON /healthcheck on SPA baseURL"]
  KubeWaits["kube / deployment waits"] --> PollUntil["pollUntil()"]
  LogUtils["audit-log retries"] --> PollUntil
```

### Page objects and naming
```mermaid
graph TD
  Specs --> POMs["support/pages/*"]
  POMs --> UIHelper["ui-helper/* modules"]
  OldPO["page-objects/"] -.removed.-> POMs
  BackstageShowcase["BackstageShowcase"] -.renamed.-> RhdhInstance["RhdhInstance"]
  SettingsSidebar["Settings / Sidebar / Home POMs"] --> ResidualLocators["residual spec locators absorbed"]
```

### Test separation and lint scope
```mermaid
graph TD
  E2E["*.spec.ts under playwright/e2e/"] --> Playwright["Playwright runner"]
  E2E --> PWPlugin["eslint-plugin-playwright via Oxlint"]
  Unit["*.test.ts under e2e-tests/unit/"] --> Vitest["Vitest runner"]
  Unit --> VitestPlugin["vitest plugin via Oxlint"]
  CI["e2e-tests-lint workflow"] --> Unit
```

## Merge order
1. #5000 — CI hygiene *(merged)*
2. #5002 — OXC toolchain
3. #5003 — oxlint pedantic
4. **This PR** — e2e hardening (`feat/e2e-hardening`)
5. #5022 — post-hardening architecture (`feat/e2e-post-hardening-architecture`)
6. PR 6+ — monorepo oxfmt and remaining modernization

## Test plan
- [x] `cd e2e-tests && yarn lint`
- [x] `cd e2e-tests && yarn fmt:check`
- [x] `cd e2e-tests && yarn test:list`
- [ ] GitHub `e2e` cluster-free job green (`@cluster-free` legacy harness)
- [ ] Prow `ci/prow/e2e-ocp-helm` green after Build Image publishes PR image


jonkoops (approved):

e2e-tests/local-run.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-RUNNER_IMAGE="${RUNNER_IMAGE:-quay.io/rhdh-community/rhdh-e2e-runner:main}"
+RUNNER_IMAGE="quay.io/rhdh-community/rhdh-e2e-runner:main"
```
gustavolira, line 5: This hard-codes the runner image and, together with dropping `-i/--runner-image` and the pull-failure fallback below, reverts what #5036 added for the multi-arch runner. Was that intentional or a rebase artifact? Without the `${RUNNER_IMAGE:-...}` default there's no way to test a locally built runner image anymore.

e2e-tests/local-run.sh
```diff
@@ -450,10 +434,9 @@ log::info "Container log: $CONTAINER_LOG"
 echo ""
 
 CONTAINER_EXIT_CODE=0
-# no -t: stdout is piped to tee and CI has no TTY
 podman run -v "$WORK_DIR":/tmp/rhdh \
   -v "$SCRIPT_DIR/container-init.sh":/tmp/container-init.sh:ro \
-  -i -u root --privileged \
+  -it -u root --privileged \
```
gustavolira, line 439: The comment that was deleted here said exactly why `-t` was removed: stdout is piped to tee and CI has no TTY. Re-adding `-it` will make podman garble or fail the non-TTY runs again.

e2e-tests/container-init.sh
```diff
@@ -23,8 +23,7 @@ set -e
 if ! command -v vault &> /dev/null; then
   VAULT_VERSION="${VAULT_VERSION:-1.15.4}"
   log::info "Installing vault ${VAULT_VERSION}..."
-  VAULT_ARCH=$(dpkg --print-architecture)
-  curl -fsSL "https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${VAULT_ARCH}.zip" -o /tmp/vault.zip
+  curl -fsSL "https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip" -o /tmp/vault.zip
```
gustavolira, line 26: `dpkg --print-architecture` was added in #5036 so the runner works on arm64 — hard-coding `linux_amd64` breaks the vault install there. Can we keep the arch detection?

e2e-tests/playwright/e2e/auth-providers/github.spec.ts
```diff
@@ -311,15 +227,14 @@ test.describe("Configure Github Provider", async () => {
...
     await context.clearCookies();
   });
 
-  test.afterAll(async () => {
-    console.log("[TEST] Starting cleanup...");
-    await deployment.killRunningProcess();
-    console.log("[TEST] Cleanup completed");
+  test.afterAll(async ({ rhdhPage }, testInfo) => {
+    await harness.cleanup();
+    await teardownBrowser(rhdhPage, testInfo);
```
gustavolira, line 238: `rhdhPage`/`rhdhContext` are worker-scoped, so they survive across spec files in the same worker — and the fixture in `support/coverage/test.ts` already calls `teardownBrowser` in its own teardown. Closing the page here means the next auth spec that lands on this worker (5 files, 3 workers) starts with an already-closed page, and the worker fixture won't re-run its setup. Same pattern in the gitlab/ldap/microsoft/oidc specs. Should we drop the afterAll teardown and leave it to the fixture?

e2e-tests/package.json
```diff
@@ -43,12 +45,11 @@
     "winston": "3.14.2"
   },
   "devDependencies": {
-    "@axe-core/playwright": "4.12.1",
+    "@axe-core/playwright": "4.11.3",
     "@microsoft/microsoft-graph-types": "2.43.1",
-    "@playwright/test": "1.61.1",
+    "@playwright/test": "1.61.0",
```
gustavolira, line 50: This downgrades `@playwright/test` 1.61.1 -> 1.61.0 (and `@axe-core/playwright` 4.12.1 -> 4.11.3 above) — intentional? Playwright 1.62 fixes the SSE teardown hang (microsoft/playwright#41513) that this PR still works around with the `about:blank` navigations, so moving forward instead would let us delete those workarounds. Also `vitest: ^4.1.9` is the only caret range in an otherwise fully pinned block.

e2e-tests/playwright/utils/authentication-providers/rhdh-deployment/wait.ts
```diff
@@ -109,129 +109,98 @@ function isDeploymentReady(deployment: k8s.V1Deployment, cr: BackstageCr): boole
...
-      return { rolloutStarted, initialGeneration };
-    }
-
-    const elapsedSinceStart = Date.now() - startTime;
+    console.log("[INFO] Rollout detected");
+    return { rolloutStarted: true, initialGeneration };
+  } catch {
     console.log(
-      `[INFO] Waiting for rollout to start... (${Math.round(elapsedSinceStart / 1000)}s elapsed)`,
+      `[INFO] No rollout detected after ${rolloutStartTimeout}ms, checking if deployment is already ready`,
```
gustavolira, line 173: This catch treats any error as "no rollout detected" — including `No deployment found with labels`, which used to fail fast out of the old loop. Now a missing deployment burns the full readiness timeout before failing with a generic label. Can we rethrow anything that isn't the pollUntil timeout?

e2e-tests/playwright/utils/authentication-providers/rhdh-deployment/logs.ts
```diff
@@ -93,17 +93,15 @@ async function streamPodLogsUntilMatch(
...
-    logStream.removeAllListeners();
-  }
-  return found;
+  await pollUntil(() => Promise.resolve(found), {
+    timeoutMs,
+    intervalMs: 500,
+    label: `Log pattern ${searchString} in pod ${podName}`,
+  });
+
+  logStream.end();
```
gustavolira, line 102: If `pollUntil` times out it throws, so these cleanup lines never run and the k8s log stream leaks — worth a try/finally. And with the unconditional `return true`, the `Promise<boolean>` contract (and `expect(synced).toBe(true)` in `waitForSynced`) is dead weight now.

e2e-tests/playwright/support/pages/settings-page.ts
```diff
@@ -0,0 +1,215 @@
...
+    const labels = t["user-settings"][locale];
+    await verification.verifyText(this.page, labels["profileCard.title"]);
+    await verification.verifyText(this.page, labels["appearanceCard.title"]);
+    await verification.verifyText(this.page, labels["themeToggle.title"]);
+    await verification.verifyText(this.page, labels["signOutMenu.title"]);
+    await verification.verifyText(this.page, labels["identityCard.title"]);
+    await verification.verifyText(this.page, `${labels["identityCard.userEntity"]}: Guest User`);
+    await verification.verifyText(
+      this.page,
+      `${labels["identityCard.ownershipEntities"]}: ownershipEntities`,
```
gustavolira, line 96: This asserts the literal string `: ownershipEntities` — the placeholder, not a value. Nothing calls `verifyLocalizedUserSettingsLabels` (the spec uses the `WithOwnership` variant), so maybe just delete it, or fold the two variants together with an optional param.

e2e-tests/playwright/support/selectors/rhdh-instance-table.ts
```diff
@@ -0,0 +1,18 @@
...
+export const RHDH_INSTANCE_TABLE = {
+  getNextPageButton: (page: Page) => page.getByRole("button", { name: "Next Page" }),
+
+  getPreviousPageButton: (page: Page) => page.getByRole("button", { name: "Previous Page" }),
+
+  getLastPageButton: (page: Page) => page.getByRole("button", { name: "Last Page" }),
+
+  getFirstPageButton: (page: Page) => page.getByRole("button", { name: "First Page" }),
+
+  getTableRows: (page: Page) => SemanticSelectors.table(page).getByRole("row"),
```
gustavolira, line 15: The helper this replaces used `tbody tr`; `getByRole("row")` also matches header rows, so `verifyPRRowsPerPage`'s `toHaveCount(rows)` is now off by the header count. The consuming spec is @blocked today, which is exactly why this will be confusing when RHDHBUGS-2099 unblocks it. `getByRole("rowgroup").last().getByRole("row")` keeps it body-only.

e2e-tests/playwright/utils/ui-helper/verification.ts
```diff
@@ -69,7 +69,7 @@ export async function verifyRowsInTable(
 }
 
 export async function waitForTextDisappear(page: Page, text: string) {
-  await page.waitForSelector(`text=${text}`, { state: "detached" });
+  await expect(page.getByText(text)).toHaveCount(0);
```
gustavolira, line 72: `waitForSelector` waited up to 30s by default; `expect()` retries for only 5s (we don't override `expect.timeout` in the config). For the scaffolder template list that's a 6x cut, and the specs relying on this also lost their `test.setTimeout(130000)` in this PR. Worth an explicit `{ timeout }` here and in `waitForTitle` below?

e2e-tests/playwright/support/pages/scaffolder-flow-page.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  async verifySelfServiceHeading(): Promise<void> {
+    await verification.verifyHeading(this.page, "Self-service");
+  }
+
+  async clickImportGitRepository(): Promise<void> {
+    await interaction.clickButton(this.page, "Import an existing Git repository");
+  }
+
+  async runCreateReactAppTemplate(details: ReactAppTemplateDetails): Promise<void> {
```
gustavolira, line 50: `runCreateReactAppTemplate` repeats `fillCreateReactAppTemplateForm` verbatim plus two nav calls, and has no callers — same for `openImportGitRepository`, `verifyCreateReactAppReviewTable` and `createAndOpenInCatalog`. `CatalogBrowsePage.openTemplateFromCatalog` is also unused and subtly diverges from the version in this class. Can we drop the dead ones before someone has to guess which one is canonical?

e2e-tests/oxlint.config.ts
```diff
@@ -1,5 +1,87 @@
 import { defineConfig } from "oxlint";
 
+/** POM and helper methods that perform assertions on behalf of E2E specs. */
+const playwrightAssertFunctions = [
```
gustavolira, line 4: About half of these names are pure actions (`clickPullRequestFilter`, `incrementFirstCardCounter`, `clearSearchIfVisible`, `sortCreatedAtDescending`, `loginAsGuest`...) — registering them as assert functions means a spec that only clicks still passes `expect-expect`. Could we trim this to the `verify*`/`validate*`/`waitFor*` helpers? Related: the rationale comments explaining why each override disables its rules were dropped in the restructure — they'd be worth keeping.

e2e-tests/playwright/utils/common/index.ts
```diff
@@ -36,9 +37,27 @@ export class Common {
     this.uiHelper = new UIhelper(page);
   }
 
+  private async hasJsonHealthcheck(): Promise<boolean> {
```
gustavolira, line 40: nit: this re-implements the status-200 + JSON content-type probe that `waitForRhdhReady` already does inside its poll. A shared `isJsonHealthcheck(response)` would keep the cluster-free detection and the readiness gate from drifting.

e2e-tests/playwright/support/pages/catalog-browse-page.ts
```diff
@@ -0,0 +1,168 @@
...
+    await expect(link).toBeVisible();
+    await link.click();
+  }
+
+  async verifyTableCell(text: string): Promise<void> {
+    await expect(this.page.getByRole("cell", { name: text })).toBeVisible();
+  }
+
+  async openLicensedUsersCatalog(): Promise<void> {
+    await this.page.goto("/catalog?filters%5Bkind%5D=user&filters%5Buser");
```
gustavolira, line 160: nit: this URL is already exported as `CATALOG_USERS_BASE_URL` in `catalog-users-page.ts` — import it instead of duplicating the string.

e2e-tests/playwright/utils/common/browser.ts
```diff
@@ -30,16 +30,23 @@ export function parseAuthStateCookies(content: string): Cookie[] {
...
   const context = await browser.newContext({
-    ...(testInfo.retry > 0 && {
-      recordVideo: {
-        dir: `test-results/${path
-          .parse(testInfo.file)
-          .name.replace(".spec", "")}/${testInfo.titlePath[1]}`,
-        size: { width: 1280, height: 720 },
-      },
-    }),
+    recordVideo: {
```
gustavolira, line 46: This used to record video only on retry; now every worker-scoped context records on every green run, and unlike the built-in `page` fixture there's no retain-on-failure cleanup. On nightlies that's real artifact volume — intentional? If always-on is wanted, deleting the file on pass in `teardownBrowser` would match the config's behavior.

PR #5005, opened 2026-06-23
feat(e2e): cluster-free local E2E harness

gustavolira:
## Summary

Adds a **cluster-free E2E harness** for RHDH: run the existing Playwright specs against a real RHDH instance **without an OpenShift/Kubernetes cluster or container image build**. Playwright boots the backend and the legacy app (`packages/app`) dev servers in-process, installs dynamic plugins from the **public OCI registry** (ghcr, via skopeo — no source build), and drives the browser against them.

Part of **RHIDP-13501** (E2E Test Optimization), Layer 4a spike **RHIDP-15075**.

## Status: 4 tests green in CI ✅

The GitHub Actions job (`.github/workflows/e2e-cluster-free.yaml`) runs cluster-free and passes in **~3.5 min**:

```
✓ guest-signin-happy-path › Homepage renders with Search Bar, Quick Access and Starred Entities
✓ guest-signin-happy-path › Verify Profile is Guest in the Settings page
✓ guest-signin-happy-path › Sign Out and Verify that you return to the Sign-in page
✓ learning-path-page      › Links in Learning Paths open in a new tab
  4 passed (1.5m)
```

Install (cached): ~90s · Populate plugins from OCI: ~11s · Boot + run: ~2m

The full `guest-signin-happy-path` spec passes **unmodified** — including Settings/Sign-out, which navigate through the **global-header** dynamic plugin's profile dropdown.

## How it works

1. Install deps + skopeo (Playwright container `v1.61.1-noble` — browsers preinstalled).
2. Populate `dynamic-plugins-root` via `./e2e-tests/local-harness/populate.sh` — the single source of truth for the populate step (CI, docs, and the fail-fast globalSetup message all point to it). It installs the plugin set from `e2e-tests/local-harness/dynamic-plugins.yaml` (dynamic-home-page + global-header, from ghcr) with the CLI pinned to the version CI uses.
3. `yarn e2e:legacy-local` — boots backend + legacy app with the config overlay, guest auth, and runs the specs.

Test scoping uses a `@cluster-free` tag + a `testMatch` file allowlist: enabling a validated spec is tagging the test and adding its file to `testMatch`.

### The global-header fix (why Settings/Sign-out work now)

The repo's static `app-config.dynamic-plugins.yaml` only mounts the bare `GlobalHeader` container with no children, so off-cluster the header rendered empty. In-cluster, the full mount points (ProfileDropdown, Settings/Sign-out items, search) come from the plugin's `pluginConfig` in the catalog index. The harness now carries that canonical `pluginConfig` (copied from rhdh-plugins `workspaces/global-header/.../app-config.dynamic.yaml`) on the OCI entry, and the webServer loads the **generated** `dynamic-plugins-root/app-config.dynamic-plugins.yaml` last — the same file and merge order the production container uses. This pattern generalizes to any plugin whose config is not in the repo's static file.

## Why OCI instead of the catalog index

The catalog index's `dynamic-plugins.default.yaml` references core plugins via local `./dynamic-plugins/dist/…` paths that only exist after a source build (which CI doesn't do). The overlays repo publishes those plugins as **OCI artifacts**, which `install-dynamic-plugins` can pull with skopeo directly — no build.

## Scope / follow-ups

- **Runs 4 tests today** (full guest-signin + learning-paths). Next candidates mapped in `docs/e2e-tests/local-e2e-harness.md` (Known issues): `settings.spec.ts` needs the CI catalog test entities, `home-page-customization.spec.ts` needs the CI home-card customization — follow-up PRs (spec inventory tracked as [RHIDP-15076](https://redhat.atlassian.net/browse/RHIDP-15076)).
- **OCI tags** are backstage `1.49.4` (RHDH is on `1.52.0`); they load fine, but bump when the overlays repo publishes matching tags (noted in the config file).
- **Overlay plugin tests** are out of scope here — they get their own PR in the overlay repo.
- Not wired into any required check yet; it's an additive cluster-free job.


gashcrumb (approved):
looks great!  Is it worth mentioning/linking to this in the e2e tests [README](https://github.com/redhat-developer/rhdh/blob/main/e2e-tests/README.md) at some point?

gashcrumb (approved):

PR #5022, opened 2026-06-25
refactor(e2e): consolidate auth and runtime test seams

schultzp2020:
## Summary

- Consolidate guest and provider login behind shared `@support/coverage/test` fixtures (`guestPage`, `rhdhPage`, `rhdhContext`, `authSession`) and dedicated `support/auth/*` modules so specs stop re-implementing browser/session setup.
- Deepen post-hardening architecture seams: `createBrowserSession()` factory, `PortForwardHarness`, `RuntimeHarness`, sidebar adapter functions (no intermediate `SidebarNav` class), worker-scoped cleanup, and an `entry-graph.ts` side-effect graph for static analysis (fallow: 0 dead-code issues).
- Restore intentionally blocked/future paths without polluting default discovery: GitHub happy-path ([RHDHBUGS-2099](https://redhat.atlassian.net/browse/RHDHBUGS-2099)), runtime auto-deploy (`RUNTIME_AUTO_DEPLOY=true`), and `monocart-coverage-reports`; consolidate YAML on `yaml@2.9.0` and drop redundant `e2e-tests` lint-staged config.
- Harden Playwright patterns that caused OCP helm CI failures: auto-retrying catalog-import assertions, MUI-based loading waits before a11y scans, and accumulated port-forward readiness detection.

### Auth and session seams

```mermaid
graph TD
  Specs["E2E specs"] --> Fixtures["@support/coverage/test fixtures"]
  Fixtures --> Guest["guest-auth.ts"]
  Fixtures --> Provider["provider-auth.ts"]
  Fixtures --> BrowserSession["createBrowserSession()"]
  Guest --> AppShell["app-shell.ts"]
  Provider --> AuthHarness["AuthProviderHarness"]
  BrowserSession --> WorkerCleanups["runWorkerCleanups()"]
  AuthHarness --> Deployment["RHDH deployment helpers"]
```

### Runtime and cluster mutation seams

```mermaid
graph TD
  GlobalSetup["global-setup.ts"] -->|"RUNTIME_AUTO_DEPLOY=true"| RuntimeDeploy["ensureRuntimeDeployed()"]
  RuntimeDeploy --> RuntimeConfig["runtime-config.ts"]
  RuntimeDeploy --> Kube["KubeClient"]
  RuntimeSpecs["runtime, schema-mode, redis, external-db specs"] --> RuntimeHarness["RuntimeHarness"]
  RuntimeSpecs --> PortForwardHarness["PortForwardHarness"]
  PortForwardHarness --> PortForwardSession["PortForwardSession"]
  RuntimeHarness --> Kube
  RuntimeHarness --> Postgres["postgres-config helpers"]
  PortForwardSession --> Cluster["oc/kubectl port-forward"]
```

### Deploy modes

| Mode | Trigger |
|---|---|
| CI / pre-deployed cluster | `BASE_URL` set → readiness wait only |
| Own cluster | `BASE_URL` + creds → `yarn showcase*` |
| Legacy local (no cluster) | `yarn e2e:legacy-local` |
| Auto-deploy runtime | `RUNTIME_AUTO_DEPLOY=true` + no `BASE_URL` → deploy then wait |

## Other changes

- Bump `e2e-tests` to **ES2023**; adopt `toSorted` / `toReversed` where applicable.
- Remove dead helpers, unused exports, and duplicate POM layers identified by fallow; wire restored blocked/runtime symbols through `playwright/blocked/entry-graph.ts` and `playwright/tool-deps.ts`.
- Document when `test.describe.serial` is actually required (ordered side-effect chains, not worker-scoped fixtures by default).

## Test plan

- [x] `cd e2e-tests && yarn lint`
- [x] `cd e2e-tests && yarn fmt:check`
- [x] `cd e2e-tests && npx fallow` (0 dead-code issues)
- [x] `cd e2e-tests && yarn test:list`
- [ ] Cluster-backed Playwright projects in Prow (`ci/prow/e2e-ocp-helm`) — re-triggered after CI wait hardening


e2e-tests/playwright/entry-graph.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+/* oxlint-disable import/no-unassigned-import -- intentional side-effect graph wiring */
+
+import "./support/coverage/test";
+import { createBrowserSession } from "./support/browser-session";
+import requireDynamicPluginsPopulated from "./support/local-harness-global-setup";
+import { runWorkerCleanups } from "./support/worker-session";
+import "./utils/common/browser";
+import "./utils/ui-helper/navigation";
+import "./blocked/entry-graph";
```
gustavolira, line 13: This import chain is what's breaking the "Verify Playwright test collection" step in CI: `playwright.config.ts` loads this file, which loads `blocked/entry-graph.ts`, which loads `github-happy-path.blocked.ts` — and that file calls `test.describe()` at module scope, so every Playwright invocation dies with `Playwright Test did not expect test.describe() to be called here`. The blocked spec can't be part of the config's import graph; if the goal is just keeping it compiling, tsc/oxlint already cover it via the tsconfig include.

e2e-tests/playwright/e2e/guest-signin-happy-path.spec.ts
```diff
@@ -13,42 +11,28 @@ test.describe("Guest Signing Happy path", () => {
...
-  test(
-    "Verify the Homepage renders with Search Bar, Quick Access and Starred Entities",
-    { tag: "@cluster-free" },
-    async () => {
-      await rhdhHomePage.verifyWelcomeHeading();
-      await rhdhHomePage.openHomeSidebar();
-      await homePage.verifyQuickAccess("Developer Tools", "Podman Desktop");
-    },
-  );
+  test("Verify the Homepage renders with Search Bar, Quick Access and Starred Entities", async () => {
```
gustavolira: These three tests lost their `@cluster-free` tags (and the "verified green on the cluster-free harness" comments), while learning-path kept its tag. The GitHub `e2e` job still passes, but it now runs essentially one spec instead of four — a green check hiding removed coverage. Was this intentional (e.g., the new `guestPage` fixture not working on the legacy harness)? If so, it's worth a callout in the PR description.

e2e-tests/playwright/e2e/plugin-division-mode-schema/verify-schema-mode.spec.ts
```diff
@@ -1,167 +1,166 @@
...
+  }
+
+  return { portForwardHarness, testSetup };
 }
 
 test.describe("Verify pluginDivisionMode: schema", () => {
   const namespace = process.env.NAME_SPACE_RUNTIME ?? "showcase-runtime";
-  const releaseName = process.env.RELEASE_NAME ?? "rhdh";
-  const installMethod = resolveInstallMethod();
+  const releaseName = process.env.RELEASE_NAME ?? "developer-hub";
```
gustavolira: Two defaults changed here that look off: the `RELEASE_NAME` fallback went from `"rhdh"` to `"developer-hub"`, which with `getDeploymentName()`'s helm naming yields `developer-hub-developer-hub` (while `getRhdhDeploymentName()` still defaults to `rhdh-developer-hub`). And the line below reads only `INSTALL_METHOD`, dropping the JOB_NAME fallback that `resolveInstallMethod()` in helper.ts still provides — operator nightlies that don't set `INSTALL_METHOD` will silently take the helm path.

e2e-tests/playwright/e2e/plugin-division-mode-schema/schema-mode-setup.ts
```diff
@@ -3,43 +3,64 @@
...
+    if (this.installMethod === "operator") {
+      return `backstage-${this.releaseName}`;
+    }
+    return `${this.releaseName}-developer-hub`;
   }
 
   private getSecretName(): string {
     if (this.installMethod === "operator") {
-      return `backstage-psql-secret-${this.releaseName}`;
+      return "postgres-cred";
```
gustavolira: This walks back the operator fixes from #5027: the operator secret was `backstage-psql-secret-${releaseName}`, the block preserving `POSTGRESQL_ADMIN_PASSWORD` is gone (its own comment said the operator's PostgreSQL pod fails to start without it), the operator branch of `resolveRhdhPostgresHost` (`backstage-psql-${releaseName}`) is gone, and the connection is now unconditionally `ssl: { rejectUnauthorized: false }` where internal DBs used non-SSL. If postgres-cred + CR extraEnvs is the new deliberate direction that's fine — but as a diff it reads like a rebase over #5027 gone wrong. Can you confirm which one it is?

e2e-tests/playwright/e2e/plugin-division-mode-schema/schema-mode-setup.ts
```diff
@@ -62,94 +83,61 @@ export class SchemaModeTestSetup {
...
-      }
-    }
-
+    // 1. Update secret with schema-mode credentials
     await this.kubeClient.createOrUpdateSecret(
       {
         metadata: { name: secretName },
-        data: secretData,
+        data: {
+          password: Buffer.from(this.env.dbPassword).toString("base64"),
```
gustavolira: nit: this inlines `Buffer.from(...).toString("base64")` seven times where `base64Encode` from utils/helper exists, hard-codes `"backstage-backend"` twice where `BACKSTAGE_BACKEND_CONTAINER` exists, and further down replaces `kubeClient.jsonPatchDeployment` with a raw `patchNamespacedDeployment` carrying five positional `undefined`s. The helpers were there for exactly this.

e2e-tests/playwright/support/harnesses/runtime-harness.ts
```diff
@@ -0,0 +1,90 @@
...
+          label: "Failed to restart deployment",
+        },
+      );
+    } catch {
+      const message = lastError instanceof Error ? lastError.message : "unknown error";
+      throw new Error(`Failed to restart deployment: ${message}`);
+    }
+  }
+
+  async configureExternalPostgres(options: ExternalPostgresOptions): Promise<void> {
```
gustavolira, line 76: The old flow called `prepareForExternalDatabase()` to patch app-config with `${POSTGRES_*}` placeholders and inject the env vars from the postgres-cred secret; that helper was deleted and this method only writes the cert/cred secrets and restarts. Unless the runtime deployment now ships that wiring by default, RHDH keeps using its internal DB and "Verify successful DB connection" passes without ever touching the external database. Where does the connection config come from now?

e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-azure-db.spec.ts
```diff
@@ -45,74 +37,54 @@ test.describe("Verify TLS configuration with Azure Database for PostgreSQL healt
...
-    // Validate certificates are available — skip gracefully if not set
+    // Validate certificates are available
     const azureCerts = readCertificateFile(process.env.AZURE_DB_CERTIFICATES_PATH);
-    if (azureCerts === null || azureCerts === undefined || !azureUser || !azurePassword) {
-      testInfo.skip(
-        true,
-        "Azure DB environment variables not configured (AZURE_DB_CERTIFICATES_PATH, AZURE_DB_USER, AZURE_DB_PASSWORD) — Azure DB tests are opt-in",
+    if (azureCerts === undefined || azureCerts === null || azureCerts === "") {
+      throw new Error(
+        "AZURE_DB_CERTIFICATES_PATH environment variable must be set and point to a valid certificate file",
```
gustavolira: This used to be `testInfo.skip(...)` with an explicit "these tests are opt-in" message — turning it into a throw means any environment without the Azure/RDS credentials now fails the suite instead of skipping it. Same change in the RDS spec. Was dropping the opt-in semantics intended?

e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-azure-db.spec.ts
```diff
@@ -45,74 +37,54 @@ test.describe("Verify TLS configuration with Azure Database for PostgreSQL healt
...
+      test("Configure and restart deployment", async () => {
+        await runtimeHarness.configureExternalPostgres({
+          credentials: {
+            host: config.host,
+            user: azureUser,
+            password: azurePassword,
+          },
         });
-        await kubeClient.restartDeployment(deploymentName, namespace);
+        expect(config.host).toBeTruthy();
```
gustavolira, line 94: This assert runs after `configureExternalPostgres` already consumed `config.host` — if the host is unset, the failure happens inside the harness with a worse message. The per-version `if (!config.host) skip` that used to guard this was removed; either assert first or restore the skip. Same in the RDS spec.

e2e-tests/playwright/utils/ui-helper/navigation.ts
```diff
@@ -1,159 +1,54 @@
...
-    name: sectionLabel,
-    exact: true,
-  });
-  await expect(sectionButton).toBeVisible();
-
-  const sectionGroup = sectionButton.locator("xpath=..");
-  const scopedLink = sectionGroup.getByRole("link", { name: linkName, exact: true });
-  if ((await scopedLink.count()) > 0 && (await scopedLink.first().isVisible())) {
-    return scopedLink.first();
+  if (await hasJsonHealthcheck(page)) {
```
gustavolira: `hasJsonHealthcheck` is a GET to `/healthcheck`, and `runSidebarAction` wraps `openSidebar`, `openSidebarButton`, and `waitForSideBarVisible` — so every sidebar interaction now costs an HTTP round-trip. The legacy-vs-RHDH answer can't change mid-run; detect once and cache it (module-level per baseURL, or in the fixture).

e2e-tests/playwright/support/fixtures/auth-provider-harness.ts
```diff
@@ -89,6 +130,16 @@ export class AuthProviderHarness {
     await this.deployment.waitForSynced();
   }
 
+  async runLoginCase(options: AuthLoginCase): Promise<void> {
```
gustavolira, line 133: `cleanup` only runs when `assert` succeeds — a failed assertion leaves the provider session cookies on the shared worker page and contaminates the next test (and the retry). Should login/assert be wrapped in try/finally with cleanup in the finally?

e2e-tests/playwright/tool-deps.ts
```diff
@@ -0,0 +1,8 @@
+/** Wires npm devDependencies used only via config or shell scripts into the entry graph. */
+import eslintPluginPlaywright from "eslint-plugin-playwright";
+import { CoverageReport } from "monocart-coverage-reports";
```
gustavolira: The `void X` entry-graph pattern fights the tool it's feeding: if an export's only reference is a `void` statement, it is unused — and this file pulls monocart-coverage-reports and shellcheck into the Playwright config's runtime import graph just to mark them "used". Dependencies consumed only by config/scripts are what the checker's ignore-list is for.

e2e-tests/playwright/support/harnesses/port-forward-harness.ts
```diff
@@ -0,0 +1,45 @@
+import { bindPortForwardRestarter } from "../../e2e/plugin-division-mode-schema/schema-mode-db";
```
gustavolira: A `support/` module importing from `e2e/plugin-division-mode-schema/` inverts the layering — support code now depends on spec-area code, so any spec reshuffle breaks the harness. `bindPortForwardRestarter` would sit better next to `PortForwardSession` in utils/.

e2e-tests/playwright/support/pages/techdocs-page.ts
```diff
@@ -1,24 +1,24 @@
...
-    this.ui = new UIhelper(page);
+  constructor(private readonly page: Page) {
     this.sidebar = new SidebarPage(page);
   }
 
   async openDocFromFavorites(docName: string): Promise<void> {
-    await this.sidebar.openFavoritesDocs();
-    await this.ui.clickLink(docName);
+    await this.sidebar.openSidebarButton("Favorites");
+    await this.sidebar.openSidebar("Docs");
```
gustavolira: nit: this hard-codes English "Docs" where the removed `openFavoritesDocs` used `t["rhdh"][lang]["menuItem.docs"]` — and sidebar.spec.ts now re-inlines that same localized pair itself. Worth keeping one localized helper.

e2e-tests/playwright/e2e/auth-providers/oidc.spec.ts
```diff
@@ -242,89 +277,94 @@ test.describe("Configure OIDC provider (using RHBK)", () => {
...
-    expect(process.env.AUTH_PROVIDERS_GH_ORG_CLIENT_ID!).toBeDefined();
-    // set up GitHub auth
-    harness.deployment.setAppConfigProperty("auth.providers.github", {
-      production: {
-        clientId: "${AUTH_PROVIDERS_GH_ORG_CLIENT_ID}",
-        clientSecret: "${AUTH_PROVIDERS_GH_ORG_CLIENT_SECRET}",
-        callbackUrl: "${BASE_URL:-http://localhost:7007}/api/auth/github/handler/frame",
+    await harness.runLoginCase({
+      configure: async () => {
+        await Promise.resolve();
```
gustavolira: nit: this test doesn't really fit the `runLoginCase` template — `configure` is just `await Promise.resolve()` plus env asserts, and the actual config change happens inside `assert` after the login. Leaving this one hand-rolled would read better than forcing the slots.

e2e-tests/playwright/e2e/configuration-test/config-map.spec.ts
```diff
@@ -1,14 +1,10 @@
...
-import { KubeClient, getRhdhDeploymentName, isRecord } from "../../utils/kube-client";
-import { ensureRuntimeDeployed } from "../../utils/runtime-deploy";
+import { RuntimeHarness } from "../../support/harnesses/runtime-harness";
+import { HomePage } from "../../support/pages/home-page";
 
 test.describe("Change app-config at e2e test runtime", () => {
-  test.beforeAll(async () => {
-    // 15 minutes — includes deployment if needed
-    test.setTimeout(900000);
+  test.beforeAll(() => {
```
gustavolira: Dropping `ensureRuntimeDeployed()` here breaks the contract documented in `.ci/pipelines/jobs/ocp-nightly.sh` ("The first test file (config-map.spec.ts) calls ensureRuntimeDeployed() which ... deploys RHDH with Helm"). CI always exports `BASE_URL` (`testing::run_tests`), so the new `RUNTIME_AUTO_DEPLOY` path in global-setup never fires — nothing deploys the runtime instance before this spec patches its ConfigMap. Is there a companion `.ci` change that sets `RUNTIME_AUTO_DEPLOY` instead of `BASE_URL`? Without one the nightly runtime job has no deploy step.

e2e-tests/playwright/e2e/plugin-division-mode-schema/verify-schema-mode.spec.ts
```diff
@@ -1,167 +1,167 @@
...
-      adminPassword === "" ||
-      dbPassword === undefined ||
-      dbPassword === "" ||
-      (!hasPortForwardMeta && !hasDirectHost)
-    ) {
-      testInfo.skip(
-        true,
-        "SCHEMA_MODE_* environment variables not set - schema mode tests are opt-in",
-      );
+    const setup = await setupSchemaModeTests(testInfo, namespace, releaseName, installMethod);
```
gustavolira, line 152: With `ensureRuntimeDeployed()` gone from this beforeAll, the `SCHEMA_MODE_*` vars are only set if an external-database spec happened to run earlier in the same worker process. Playwright starts a fresh worker after any failure and on every retry, and fresh workers inherit env from the main process — so a retried schema-mode test lands with the vars unset, `readSchemaModeEnv()` returns null, and the whole suite silently skips instead of failing. Calling `configureSchemaMode()` here (it's idempotent) would keep the spec self-sufficient.

e2e-tests/playwright/support/coverage/reporter.ts
```diff
@@ -15,41 +15,70 @@
...
       return;
     }
+    void this.prepareRawDir();
+  }
+
+  onEnd(): void {
+    if (!this.enabled) {
+      return;
+    }
+    void this.generateReport();
```
gustavolira: `onEnd` can return a Promise and Playwright awaits it — switching to `void this.generateReport()` means the runner starts shutting down while monocart is still merging raw files, so the CI report can come out truncated or missing. `onBegin` has the inverse race: the raw-dir cleanup can still be deleting files while the first tests write coverage JSON. `return this.generateReport()` / `return this.prepareRawDir()` keeps the old awaited behavior.

e2e-tests/playwright/e2e/localization/locale.ts
```diff
@@ -1,19 +1,36 @@
...
-import frRhdhPlugins from "../../../../translations/rhdh-plugins-fr.json" with { type: "json" };
-import itRhdhPlugins from "../../../../translations/rhdh-plugins-it.json" with { type: "json" };
-import jaRhdhPlugins from "../../../../translations/rhdh-plugins-ja.json" with { type: "json" };
-import en from "../../../../translations/test/all-en.json" with { type: "json" };
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+type TranslationFile = Record<string, Record<string, Record<string, string>>>;
+
+const TRANSLATIONS_DIR = join(process.cwd(), "..", "translations");
```
gustavolira: This assumes every invocation starts in `e2e-tests/` — the static JSON imports this replaces resolved relative to the module and worked from any cwd (IDE test runners, repo-root invocations). Resolving relative to the module (`import.meta.dirname` / `new URL(..., import.meta.url)`) keeps it location-independent.

e2e-tests/playwright.config.ts
```diff
@@ -59,7 +61,7 @@ export default defineConfig({
     locale: process.env.LOCALE ?? "en",
     baseURL: process.env.BASE_URL,
     ignoreHTTPSErrors: true,
-    trace: "on",
+    trace: "on-first-retry",
```
gustavolira: With `retries: 0` locally, `on-first-retry` means a local failure produces no trace at all — which is exactly the local-run.sh repro workflow where you want one. `retain-on-failure` gives failure traces both locally and on CI for roughly the same overhead (and matches the "retain on failure" behavior our docs describe).

e2e-tests/playwright/e2e/audit-log/auditor-rbac.spec.ts
```diff
@@ -17,23 +17,26 @@ import {
...
     test.info().annotations.push({
       type: "component",
       description: "audit-log",
     });
 
     await (await import("./log-utils")).LogUtils.loginToOpenShift();
-    common = new Common(rhdhPage);
-    await common.loginAsKeycloakUser();
+    authSession = rhdhAuthSession;
+    await authSession.loginWithKeycloak(
```
gustavolira: The old `loginAsKeycloakUser()` ended with `waitForSideBarVisible()`; here the popup result isn't asserted and there's no readiness wait before `buildRbacApi(rhdhPage)`, so building the API client races the post-popup redirect. Worth an `expect(result).toBe("Login successful")` plus a sidebar wait, same as the harness login path does.

e2e-tests/playwright/utils/ui-helper/navigation.ts
```diff
@@ -1,159 +1,98 @@
...
-    }
-    if (keepHandle !== null && (await button.evaluate((el, keep) => el === keep, keepHandle))) {
-      continue;
-    }
-    await button.click();
-    await expect(button).toHaveAttribute("aria-expanded", "false");
+async function usesRhdhSidebar(page: Page): Promise<boolean> {
+  const baseUrl = process.env.BASE_URL ?? page.url();
+  if (cachedSidebarBaseUrl !== baseUrl || cachedUsesRhdhSidebar === undefined) {
+    cachedUsesRhdhSidebar = await detectRhdhSidebar(page);
```
gustavolira: Caching fixes the healthcheck-per-interaction cost, but it also means whatever page state the first sidebar call sees decides the adapter for the entire worker. If that first call happens before the nav renders (or on a page where `login-button` is present), the worker gets pinned to the legacy adapter with no way to recover. Maybe cache only a positive RHDH detection and re-probe when the result is legacy?

e2e-tests/playwright/support/worker-session.ts
```diff
@@ -0,0 +1,16 @@
+import type { TestInfo } from "@playwright/test";
+
+/** Minimal scope info for worker-scoped cleanup registration. */
+export type WorkerCleanupScope = Pick<TestInfo, "workerIndex">;
+
+const workerCleanups = new Map<number, Array<() => Promise<void>>>();
```
gustavolira, line 6: Nothing can ever put an entry in this map — the file only exports `runWorkerCleanups`, so the call in `coverage/test.ts` is always a no-op. Either add the `registerWorkerCleanup()` this seam implies, or drop the file; as-is it's dead machinery that fallow can't flag because entry-graph voids it.

e2e-tests/playwright/support/harnesses/runtime-harness.ts
```diff
@@ -0,0 +1,96 @@
...
+    credentials: ExternalPostgresOptions["credentials"],
+  ): Promise<void> {
+    await configurePostgresCredentials(this.kubeClient, this.namespace, credentials);
+  }
+
+  async restartDeployment(): Promise<void> {
+    await this.kubeClient.restartDeployment(this.deploymentName, this.namespace);
+  }
+
+  async restartDeploymentWithRetry(timeoutMs = 90_000, intervalMs = 15_000): Promise<void> {
```
gustavolira, line 49: `KubeClient` already has a `restartDeploymentWithRetry` (schema-mode-setup uses it) — this one reimplements it with different semantics: the 90s pollUntil budget spans full scale-down/up cycles, so a slow-but-healthy first restart can eat the whole window before it ever retries. Can this delegate to `this.kubeClient.restartDeploymentWithRetry(...)` so there's one retry behavior?

e2e-tests/playwright/utils/ui-helper/verification.ts
```diff
@@ -38,22 +39,17 @@ export async function verifyTextVisible(
   page: Page,
   text: string,
   exact = false,
-  timeout = 10000,
+  timeout = 30_000,
```
gustavolira: Bumping the defaults from 5s/10s to 30s means every failing `verifyText`/`verifyTextVisible` call now takes 30s to report — double the global expect timeout of 15s the rest of the suite uses. If specific flaky call sites need 30s, passing it there keeps red runs fast everywhere else.

e2e-tests/playwright/support/pages/settings-page.ts
```diff
@@ -171,29 +161,37 @@ export class SettingsPage {
...
   async verifySidebarMenuItemHidden(text: string): Promise<void> {
-    await expect(this.page.getByText(text)).toBeHidden();
+    await expect(this.page.getByRole("link", { name: text })).toBeHidden();
   }
 
   async openFromProfile(userName: string): Promise<void> {
-    await this.page.getByText(userName).click();
+    const header = interaction.getGlobalHeader(this.page);
+    await expect(header).toBeVisible();
+    const profileName = userName.endsWith(" User") ? userName : `${userName} User`;
```
gustavolira, line 178: nit: this works for "Guest" but silently turns any real display name into e.g. "Zeus Giove User". Since the only caller passes "Guest", it seems simpler to take the exact profile-button name and pass "Guest User" at the call site.

e2e-tests/playwright/support/pages/settings-page.ts
```diff
@@ -12,6 +12,31 @@ const lang = getCurrentLanguage();
 
 const LANGUAGE_OPTIONS_PATTERN = /English|Deutsch|Español|Français|Italiano|日本語/u;
 
+type UserSettingsLocale = keyof (typeof t)["user-settings"];
+
+async function verifyLocalizedUserSettingsLabelsCore(
+  page: Page,
+  locale: UserSettingsLocale,
+  ownershipEntities: string,
+  includeSignOutMenu: boolean,
```
gustavolira: nit: `includeSignOutMenu` is always false now that `verifyLocalizedUserSettingsLabels` is gone — the param and its branch can go.

e2e-tests/playwright/e2e/verify-redis-cache.spec.ts
```diff
@@ -18,48 +14,40 @@ test.describe("Verify Redis Cache DB", () => {
...
+  let portForward: PortForwardHarness | null = null;
   let redis: Redis;
 
-  test.beforeEach(async ({ page }) => {
-    techDocsPage = new TechDocsPage(page);
-    common = new Common(page);
-    await common.loginAsGuest();
-
+  test.beforeAll(async () => {
+    await portForward?.stop();
```
gustavolira: nit: `portForward` is still null here, so this `stop()` is a no-op — leftover from an earlier iteration?

e2e-tests/local-run.sh
```diff
@@ -357,9 +366,16 @@ if [[ "$CLI_MODE" == "false" ]]; then
   echo ""
 fi
 
-# Pull runner image first (can take a while)
+# Pull runner image (always attempt; fall back to local copy if pull fails)
 log::section "Pulling runner container image"
-podman pull "$RUNNER_IMAGE" --platform=linux/amd64
+if ! podman pull "$RUNNER_IMAGE"; then
```
gustavolira, line 371: Dropping `--platform=linux/amd64` changes what Apple-silicon users pull (native arch instead of forced amd64). Looks intentional given the dpkg arch change in container-init.sh, but is the runner image actually published multi-arch? If it's amd64-only this is fine; if a stale arm64 tag exists somewhere, folks will get a different runtime than CI.

nickboldt: Paul says this is addressed.

e2e-tests/playwright/support/pages/settings-page.ts
nickboldt: Paul says this is addressed.

e2e-tests/playwright/e2e/plugin-division-mode-schema/verify-schema-mode.spec.ts
nickboldt: Paul says this is addressed.

nickboldt (approved):
If Gustavo & Paul are happy, so be it.

PR #5026, opened 2026-06-29
docs: fix lock file deadlock documentation for dynamic plugins cache

Fortune-Ndlovu:
## Summary

- Fix incorrect lock file name in manual cleanup command (`dynamic-plugins.lock` → `install-dynamic-plugins.lock`, matching [types.ts#L78](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts#L78))
- Document the 10-minute lock timeout behavior introduced in the TypeScript rewrite ([lock-file.ts#L22](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/lock-file.ts#L22))
- Document CrashLoopBackOff cycle that occurs after timeout with a stale lock
- Add `DYNAMIC_PLUGINS_LOCK_TIMEOUT_MS` environment variable documentation
- Update log message to match current implementation
- Clarify that lock contention only applies when using a persistent volume

Relates-to: [RHDHBUGS-450](https://redhat.atlassian.net/browse/RHDHBUGS-450)

## Test plan

- [x] Verified lock file name matches source code constant (`install-dynamic-plugins.lock`)
- [x] Verified log message matches [lock-file.ts#L50](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/lock-file.ts#L50)
- [x] Verified timeout error message matches [lock-file.ts#L46-L48](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/lock-file.ts#L46-L48)
- [x] Verified default timeout value matches [README.md#L107](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/README.md#L107)

### DIFF
--- a/docs/dynamic-plugins/installing-plugins.md
+++ b/docs/dynamic-plugins/installing-plugins.md
@@ -287,7 +287,7 @@ When using the Operator ....
 
 The directory where dynamic plugins are located is mounted as a volume to the `install-dynamic-plugins` init container and the `backstage-backend` container. The `install-dynamic-plugins` init container is responsible for downloading and extracting the plugins into this directory. Depending on the deployment method, the directory is mounted as an ephemeral or persistent volume. In the latter case, the volume can be shared between several Pods, and the plugins installation script is also responsible for downloading and extracting the plugins only once, avoiding conflicts.
 
-**Important Note:** When the `dynamic-plugins-root` directory is backed by a persistent volume, the `install-dynamic-plugins` init container uses a lock file (`/dynamic-plugins-root/install-dynamic-plugins.lock`) to prevent concurrent plugin installations across Pods that share the same volume. The lock is acquired before installation begins and released when it completes (or fails).
+**Important Note:** The `install-dynamic-plugins` init container always acquires a lock file (`/dynamic-plugins-root/install-dynamic-plugins.lock`) before installing plugins. The lock prevents concurrent installations and is released when the process completes (or fails). Lock contention — where one pod waits for another's lock — only occurs when the `dynamic-plugins-root` directory is backed by a persistent volume shared between pods.
 
 If the `install-dynamic-plugins` init container is killed with a SIGKILL signal, the lock file cannot be cleaned up. This may happen due to the following reasons:
 
@@ -323,4 +323,4 @@ The lock timeout can be configured via the `DYNAMIC_PLUGINS_LOCK_TIMEOUT_MS` env
 
 > **Note:** In RHDH 1.10.x and earlier, the install script used a Python implementation with no lock timeout. A stale lock file would cause the init container to wait indefinitely, and the only way to recover was to manually delete the lock file. The timeout, configurable environment variable, and automatic lock cleanup on exit were introduced with the TypeScript rewrite of the install script.
 
-Note: This lock file behavior only applies when using a persistent volume for the `dynamic-plugins-root` directory. With the default ephemeral volume, each pod gets its own volume, so no lock contention can occur.
+> **Note:** Lock contention only applies when using a persistent volume for the `dynamic-plugins-root` directory. With the default ephemeral volume, each pod gets its own volume, so no lock contention can occur.
--- a/docs/dynamic-plugins/installing-plugins.md
+++ b/docs/dynamic-plugins/installing-plugins.md
@@ -320,7 +320,3 @@ oc exec -n <namespace-name> deploy/backstage-<backstage-name> -c install-dynamic
 ```
 
 The lock timeout can be configured via the `DYNAMIC_PLUGINS_LOCK_TIMEOUT_MS` environment variable on the `install-dynamic-plugins` init container (value in milliseconds, default: `600000` which is 10 minutes).
-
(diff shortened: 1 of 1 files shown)


docs/dynamic-plugins/installing-plugins.md
```diff
@@ -287,23 +287,38 @@ When using the Operator ....
...
+```
+
+After the timeout expires, the init container exits with an error:
+
 ```
+Timed out after 600000ms waiting for lock file /dynamic-plugins-root/install-dynamic-plugins.lock.
+Another install may be stuck — remove the file manually to proceed.
+```
+
+The pod then enters a CrashLoopBackOff cycle, restarting and waiting again every 10 minutes, until the stale lock file is manually removed.
```
rm3l: After a quick test, it doesn't seem to lead to a CLBO. An exception is thrown, but the lock file is actually unconditionally removed even if the process exits after an exception. See https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/installer.ts#L102 and the exit handler in https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/lock-file.ts#L79
So the pod is restarted, but the next run will start with no lock file present, so it won't be stuck in CLBO.
Same behavior even if the pod is abruptly killed. The next pod will try to delete the lock file after this timeout expires. Manual intervention would indeed be needed if there is an issue with deletion or if I don't want to wait for the timeout to expire.

Fortune-Ndlovu: Yeah done testing also it seems like:
- RHDH 1.10.x uses Python,  infinite deadlock, no timeout, DYNAMIC_PLUGINS_LOCK_TIMEOUT_MS is ignored
 - RHDH 1.11.0 uses TypeScript, 10-minute default timeout, configurable via env var, exit handler auto-cleans the stale lock

your correct no CrashLoopBackOff. The pod recovers after one timeout cycle (restart + clean start) manual intervention only needed if you don't want to wait for the timeout to expire

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -287,23 +287,40 @@ When using the Operator ....
 
 The directory where dynamic plugins are located is mounted as a volume to the `install-dynamic-plugins` init container and the `backstage-backend` container. The `install-dynamic-plugins` init container is responsible for downloading and extracting the plugins into this directory. Depending on the deployment method, the directory is mounted as an ephemeral or persistent volume. In the latter case, the volume can be shared between several Pods, and the plugins installation script is also responsible for downloading and extracting the plugins only once, avoiding conflicts.
 
-**Important Note:** If `install-dynamic-plugins` init container was killed with SIGKILL signal, which may happen due to the following reasons:
+**Important Note:** When the `dynamic-plugins-root` directory is backed by a persistent volume, the `install-dynamic-plugins` init container uses a lock file (`/dynamic-plugins-root/install-dynamic-plugins.lock`) to prevent concurrent plugin installations across Pods that share the same volume. The lock is acquired before installation begins and released when it completes (or fails).
```
rm3l: > When the `dynamic-plugins-root` directory is backed by a persistent volume, the `install-dynamic-plugins` init container uses a lock file

Might be a bit misleading, as the lock file is always checked/created when the install-dynamic-plugins init container starts, regardless of the volume type. Only the contention scenario (another pod holding the lock or a pod that was SIGKILL'ed and reusing the volume) requires a persistent volume.

Fortune-Ndlovu: I have clarified the wording, thanks!

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -287,23 +287,36 @@ When using the Operator ....
...
+After the timeout expires, the init container exits with an error:
+
+```
+Timed out after 600000ms waiting for lock file /dynamic-plugins-root/install-dynamic-plugins.lock.
+Another install may be stuck — remove the file manually to proceed.
+```
+
+The exit handler automatically removes the stale lock file during shutdown. The pod restarts, and the next init container run starts with no lock file present, so it proceeds normally. The total recovery time equals the configured lock timeout (10 minutes by default). No manual intervention is required.
+
+To skip the timeout wait and recover immediately, delete the lock file manually:
```
rm3l: I'd also suggest adding that manual deletion would be needed if the exit handler could not delete the lock file.

Fortune-Ndlovu: sounds good ill add a note that manual deletion would still be needed if the exit handler itself fails to delete the lock file

rm3l (approved):

PR #4909, opened 2026-06-01
chore: breaking change - remove wrappers...

nickboldt:
### What does this PR do?

chore: breaking change - remove wrappers from RHDH entirely (RHDHPLAN-256, [RHDHPLAN-934](https://redhat.atlassian.net/browse/RHDHPLAN-934))

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### Screenshot/screencast of this PR
N/A

### What issues does this PR fix or reference?
N/A (or see commit message above for issue number)

### How to test this PR?
N/A

### PR Checklist

As the author of this Pull Request I made sure that:

- [x] Code produced is complete
- [ ] Code builds without errors
- [ ] Tests are covering the bugfix
- [ ] Relevant user documentation updated
- [ ] Relevant contributing documentation updated

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

PR #5027, opened 2026-06-29
fix(e2e): fix runtime tests for operator deployments and CI reporting

zdrapela:
Follow-up to #4809. Fixes issues found during CI verification of the runtime tests refactor.

### 1. WebSocket teardown hang (`about:blank` in `afterEach`)
Move `page.goto('about:blank')` from per-test `try/finally` blocks to `test.afterEach` hooks. RHDH UI establishes WebSocket connections that prevent Playwright from closing the page cleanly — navigating to `about:blank` drops them before teardown.

### 2. External DB tests for operator deployments
- Add `postgres-cred` to the Backstage CR's `extraEnvs.secrets` so the operator injects `POSTGRES_*` env vars automatically
- Skip direct deployment env var patching in `prepareForExternalDatabase` for operator installs (operator reconciliation reverts direct patches)

### 3. CI reporting: pessimistic test status
Write a pessimistic default (`failed=true`, `count=N/A`) immediately after registering a test run. If Prow kills the job mid-test, `STATUS_TEST_FAILED.txt` and `STATUS_NUMBER_OF_TEST_FAILED.txt` still have entries for all registered test runs — preventing misaligned arrays that break Slack notifications.

`save_status_test_failed` and `save_status_number_of_test_failed` now regenerate the file from the in-memory array instead of appending, making them safe to call multiple times for the same deployment ID.

Jira: [redhat.atlassian.net/browse/RHIDP-9140](https://redhat.atlassian.net/browse/RHIDP-9140)
Jira: [redhat.atlassian.net/browse/RHIDP-9141](https://redhat.atlassian.net/browse/RHIDP-9141)

### DIFF
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -56,7 +56,7 @@ testing::run_tests() {
   # If the job is killed (Prow timeout) or Playwright hangs, the STATUS files
   # still have entries for all registered test runs — preventing misaligned
   # arrays that break downstream reporting (Slack notifications).
-  test_run_tracker::mark_test_result "false" "N/A"
+  test_run_tracker::mark_test_result "false" "${UNKNOWN_FAILURE_COUNT}"
 
   BASE_URL="${url}"
   export BASE_URL
@@ -189,15 +189,15 @@ testing::run_tests() {
     failed_tests=$((_junit_failures + _junit_errors))
     if [[ "${failed_tests}" -eq 0 ]]; then
       # Playwright exited non-zero but JUnit reports 0 failures and 0 errors —
-      # the process likely crashed or timed out globally. Report "some" so the
-      # Slack alert doesn't misleadingly say "0 tests failed".
-      failed_tests="some"
+      # the process likely crashed or timed out globally. Use the sentinel so
+      # the Slack alert doesn't misleadingly say "0 tests failed".
+      failed_tests="${UNKNOWN_FAILURE_COUNT}"
     fi
     echo "Number of failed tests: ${failed_tests}"
   else
     echo "JUnit results file not found: ${e2e_tests_dir}/${JUNIT_RESULTS}"
-    failed_tests="some"
-    echo "Number of failed tests unknown, saving as $failed_tests."
+    failed_tests="${UNKNOWN_FAILURE_COUNT}"
+    echo "Number of failed tests unknown, saving as ${failed_tests}."
   fi
   test_run_tracker::mark_test_result "$test_passed" "${failed_tests}"
   return "$test_result"
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -52,6 +52,12 @@ testing::run_tests() {
   test_run_tracker::register "$artifacts_subdir"
   test_run_tracker::mark_deploy_success
 
+  # Pessimistic default: assume tests failed until Playwright proves otherwise.
+  # If the job is killed (Prow timeout) or Playwright hangs, the STATUS files
+  # still have entries for all registered test runs — preventing misaligned
+  # arrays that break downstream reporting (Slack notifications).
+  test_run_tracker::mark_test_result "false" "N/A"
+
   BASE_URL="${url}"
   export BASE_URL
   log::info "BASE_URL: ${BASE_URL}"
--- a/.ci/pipelines/reporting.sh
+++ b/.ci/pipelines/reporting.sh
@@ -54,6 +54,13 @@ save_status_number_of_test_failed() {
 # Writes the file from scratch each time so that the same deployment ID
 # can be safely updated multiple times (e.g. pessimistic default written
 # before Playwright runs, then overwritten with the real result).
+#
+# IMPORTANT: All callers must run in the same shell process.
+# Associative arrays are not inherited by child processes (export -f
+# only exports function definitions, not array contents). If this
+# function runs in a subshell, it will see an empty array and truncate
+# the file. This is fine today — all test_run_tracker calls are
+# sequential in the main shell.
 _regenerate_status_file() {
   local var_name=$1
   local -n _arr="${var_name}"
--- a/.ci/pipelines/reporting.sh
+++ b/.ci/pipelines/reporting.sh
@@ -10,10 +10,11 @@ readonly REPORTING_LIB_SOURCED=1
 source "$(dirname "${BASH_SOURCE[0]}")"/lib/log.sh
 
 # Variables for reporting
-export STATUS_DEPLOYMENT_NAMESPACE # Array that holds the namespaces of deployments.
-export STATUS_FAILED_TO_DEPLOY     # Array that indicates if deployment failed. false = success, true = failure
-export STATUS_TEST_FAILED          # Array that indicates if test run failed. false = success, true = failure
-export OVERALL_RESULT              # Overall result of the test run. 0 = success, 1 = failure
+export STATUS_DEPLOYMENT_NAMESPACE  # Array that holds the namespaces of deployments.
+export STATUS_FAILED_TO_DEPLOY      # Array that indicates if deployment failed. false = success, true = failure
+export STATUS_TEST_FAILED           # Array that indicates if test run failed. false = success, true = failure
+export STATUS_NUMBER_OF_TEST_FAILED # Array that holds the number of test failures per deployment.
+export OVERALL_RESULT               # Overall result of the test run. 0 = success, 1 = failure
 
 mkdir -p "$ARTIFACT_DIR/reporting"
 
@@ -22,8 +23,7 @@ save_status_deployment_namespace() {
   local current_namespace=$2
   log::debug "Saving STATUS_DEPLOYMENT_NAMESPACE[\"${current_deployment}\"]=${current_namespace}"
   STATUS_DEPLOYMENT_NAMESPACE["${current_deployment}"]="${current_namespace}"
-  printf "%s\n" "${STATUS_DEPLOYMENT_NAMESPACE["${current_deployment}"]}" >> "$SHARED_DIR/STATUS_DEPLOYMENT_NAMESPACE.txt"
-  cp "$SHARED_DIR/STATUS_DEPLOYMENT_NAMESPACE.txt" "$ARTIFACT_DIR/reporting/STATUS_DEPLOYMENT_NAMESPACE.txt"
+  _regenerate_status_file "STATUS_DEPLOYMENT_NAMESPACE"
 }
 
 save_status_failed_to_deploy() {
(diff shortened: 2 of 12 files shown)


.ci/pipelines/reporting.sh
```diff
@@ -22,35 +23,47 @@ save_status_deployment_namespace() {
...
-  printf "%s\n" "${STATUS_NUMBER_OF_TEST_FAILED["${current_deployment}"]}" >> "$SHARED_DIR/STATUS_NUMBER_OF_TEST_FAILED.txt"
-  cp "$SHARED_DIR/STATUS_NUMBER_OF_TEST_FAILED.txt" "$ARTIFACT_DIR/reporting/STATUS_NUMBER_OF_TEST_FAILED.txt"
+  _regenerate_status_file "STATUS_NUMBER_OF_TEST_FAILED"
+}
+
+# Regenerate a STATUS file from its in-memory associative array.
+# Writes the file from scratch each time so that the same deployment ID
+# can be safely updated multiple times (e.g. pessimistic default written
+# before Playwright runs, then overwritten with the real result).
+_regenerate_status_file() {
```
gustavolira, line 64: The old append approach tolerated writers in different processes because the file itself was the accumulator; now the in-memory array is the source of truth and every save truncates the file. That's fine today — all call sites run sequentially in the main shell — but test-run-tracker.sh does `export -f` on these functions "for subshell compatibility", and if a tracker function ever runs in a child bash process (arrays don't survive exec), the regenerate wipes the file down to that process's empty view. Worth a sentence here stating the same-process requirement so the constraint travels with the code.

Side note: `local -n` needs bash >= 4.3 — fine in the CI containers, but it hard-fails under macOS system bash 3.2 if anyone sources this locally outside the container.

.ci/pipelines/lib/testing.sh
```diff
@@ -52,6 +52,12 @@ testing::run_tests() {
   test_run_tracker::register "$artifacts_subdir"
   test_run_tracker::mark_deploy_success
 
+  # Pessimistic default: assume tests failed until Playwright proves otherwise.
+  # If the job is killed (Prow timeout) or Playwright hangs, the STATUS files
+  # still have entries for all registered test runs — preventing misaligned
+  # arrays that break downstream reporting (Slack notifications).
+  test_run_tracker::mark_test_result "false" "N/A"
```
gustavolira: This writes `N/A` as the unknown count, but the fallback paths further down write `some` (missing JUnit file, or non-zero exit with 0 reported failures), and `mark_deploy_failed` uses `N/A` again. Whatever formats the Slack message now has to handle both sentinels plus real numbers. Should we unify on one string and pull it into a named constant? If the downstream consumer renders them differently on purpose, a short comment saying so would stop someone from "fixing" it later.

e2e-tests/playwright/utils/postgres-config.ts
```diff
@@ -371,3 +394,49 @@ async function ensurePostgresCredEnvVars(
...
+  namespace: string,
+): Promise<void> {
+  const releaseName = getReleaseName();
+  const [group, version] = BACKSTAGE_CR_API_VERSION.split("/");
+
+  const patch = {
+    spec: {
+      application: {
+        extraEnvs: {
+          secrets: [{ name: "rhdh-runtime-config" }, { name: "postgres-cred" }],
```
gustavolira: This literal duplicates the base secrets list from `generateBackstageCR` (runtime-config.ts:425). If someone later adds a third secret to the generated CR, this patch silently strips it from the running deployment mid-test — painful to debug. Could we export the base list from runtime-config and build the patch as `[...BACKSTAGE_CR_EXTRA_SECRETS, { name: "postgres-cred" }]` so the two can't drift?

e2e-tests/playwright/utils/postgres-config.ts
```diff
@@ -371,3 +394,49 @@ async function ensurePostgresCredEnvVars(
...
+        },
+      },
+    },
+  };
+
+  await kubeClient.customObjectsApi.patchNamespacedCustomObject(
```
gustavolira: nit: KubeClient already wraps this awkward positional signature for deployments (`jsonPatchDeployment` in kube-client/index.ts), and auth-providers does another raw custom-object patch in rhdh-deployment/k8s.ts. A `mergePatchNamespacedCustomObject(...)` helper on KubeClient would give the content-type header one home and this call site a readable shape.

e2e-tests/playwright/e2e/plugin-division-mode-schema/verify-schema-mode.spec.ts
```diff
@@ -156,6 +156,10 @@ test.describe("Verify pluginDivisionMode: schema", () => {
     await killPortForward(portForwardProcess);
   });
 
+  test.afterEach(async ({ page }) => {
+    await page.goto("about:blank").catch(() => {});
```
gustavolira, line 167: Since this hook requests the `page` fixture, Playwright creates a browser context and page even for tests that never touch the UI — "Verify database user has restricted permissions" here, and "Configure and restart deployment" in both external-db specs. With video enabled for all tests, each of those runs now produces a video artifact for a page that only ever visits about:blank. Not wrong, but it adds per-test overhead exactly where this PR is trying to speed up teardown. Is the broad hook intentional for simplicity? If so, fine — just worth a note.

e2e-tests/playwright/utils/runtime-config.ts
```diff
@@ -301,16 +307,44 @@ export function generateAppConfigYaml(runtimeUrl: string): string {
...
  */
 export function generateDynamicPluginsYaml(): string {
-  return yaml.dump({ includes: [] as string[], plugins: [] as unknown[] }, { lineWidth: -1 });
+  return yaml.dump(
+    {
+      includes: [] as string[],
+      plugins: [
+        {
+          package:
+            "./dynamic-plugins/dist/red-hat-developer-hub-backstage-plugin-dynamic-home-page",
```
gustavolira, line 343: This pins a new `./dynamic-plugins/dist/...` path just as e2e/CI is moving off wrapper dist paths onto OCI refs (the #4909 direction). Intentional? If OCI isn't viable for the runtime deploy path yet, a brief note here would save the migration sweep from missing this spot later.

e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-azure-db.spec.ts
```diff
@@ -89,6 +89,10 @@ test.describe("Verify TLS configuration with Azure Database for PostgreSQL healt
         });
       });
 
+      test.afterEach(async ({ page }) => {
+        await page.goto("about:blank").catch(() => {});
```
gustavolira, line 95: nit: this exact afterEach is now copy-pasted across four specs, and only config-map.spec.ts kept the comment explaining why it exists. A small helper in utils (e.g. `closeWebSocketsAfterEach(test)`) would keep the rationale in one place — and longer-term, an automatic fixture in @support/coverage/test would give every spec the teardown for free so new specs can't forget it.

e2e-tests/playwright/e2e/configuration-test/config-map.spec.ts
```diff
@@ -60,11 +67,6 @@ test.describe("Change app-config at e2e test runtime", () => {
     } catch (error) {
```
gustavolira: This catch only logs and rethrows — it existed to pair with the finally that's gone now. Playwright already reports the error with its stack, so the whole try/catch can go and the test body loses an indent level.

e2e-tests/playwright/utils/helper.ts
```diff
@@ -131,6 +131,16 @@ export function resolveInstallMethod(): "helm" | "operator" {
   return job.includes("operator") ? "operator" : "helm";
 }
 
+/**
+ * Canonical release name resolution. Returns the RELEASE_NAME env var if set
+ * and non-empty, otherwise defaults to "rhdh".
+ */
+export function getReleaseName(): string {
+  return process.env.RELEASE_NAME !== undefined && process.env.RELEASE_NAME !== ""
```
gustavolira, line 143: nit: `return process.env.RELEASE_NAME || "rhdh";` is exactly the documented semantic (unset and empty both fall back) — one of the rare spots where `||` on a string is the right tool rather than a bug. Take it or leave it if you prefer the explicit form.

.ci/pipelines/reporting.sh
zdrapela: Okay, I'll add that. 

I wouldn't care about native macOS bash - this file is not meant to be executed directly. Both CI and local runs are in a container. And you can always install GNU Bash and get rid of all problems on Mac.

e2e-tests/playwright/utils/postgres-config.ts
zdrapela: Okay, why not. Added

e2e-tests/playwright/e2e/plugin-division-mode-schema/verify-schema-mode.spec.ts
zdrapela: Good catch. Dug into this and it's a workaround for a known Playwright issue: microsoft/playwright#41513 - fixed in v1.62 (we're on 1.61). Once we upgrade, the about:blank workaround can be removed.

Updated the comments across all 4 spec files to reference the upstream issue.

e2e-tests/playwright/utils/runtime-config.ts
zdrapela: We need to stay on wrappers until the OCIs are available. Let's keep it for the full migration. Added a note.

e2e-tests/playwright/e2e/external-database/verify-tls-config-with-external-azure-db.spec.ts
zdrapela: As mentioned earlier, it's temporary workaround, until we update to 1.62 (fixed microsoft/playwright#41513). I would say it's fine as it's now. I'll keep in mind removing it.

e2e-tests/playwright/e2e/configuration-test/config-map.spec.ts
zdrapela: You're right, I missed this before. Done

e2e-tests/playwright/utils/helper.ts
zdrapela: I had this previously, but the recent update in #5003 made this error out.

The exact rules are:
- typescript/strict-boolean-expressions — rejects process.env.RELEASE_NAME in boolean context (nullable string)
- typescript/prefer-nullish-coalescing — rejects || on string operands, wants ?? instead (but ?? doesn't catch empty string "")

I'm not convinced those rules are doing us a good service, but I aligned this PR with them for now.

PR #5036, opened 2026-06-30
chore(ci): multi-arch e2e-runner image & add `oc-mirror` and `podman`

zdrapela:
Make the e2e-runner container image buildable for both `linux/amd64` and `linux/arm64`, and add tools required for disconnected environment smoke tests.

### Dockerfile

- Add **podman** and **oc-mirror** to the image (required for disconnected env. smoke test)
- Update **oc** to `4.22`
- Pin base image to **manifest list digest** (resolves per-platform, Renovate-compatible)
- Add `ARG TARGETARCH` — all tool downloads use it instead of hardcoded `amd64`/`x86_64`
- Upgrade umoci v0.4.7 → v0.6.0 (first version with arm64 binaries)
- Remove unused Go env vars (`GO_VERSION`, `GO_SHA256`, `GOPATH`)
- Clean up AWS CLI install artifacts (`rm -rf awscliv2.zip aws/`)
- Add `apt-get clean` to skopeo/podman install step

### CI Workflow

- Add matrix strategy: `ubuntu-24.04` (amd64) + `ubuntu-24.04-arm` (arm64)
- New `merge` job assembles per-arch digests into a multi-arch manifest list
- Add `pull_request` trigger — builds both platforms (without pushing) on Dockerfile/yarnrc changes to catch build failures before merge
- Push-only steps (login, digest export, merge) guarded with `if: github.event_name != 'pull_request'`
- Detect platform via `dpkg --print-architecture` instead of mapping runner label names
- Add `HAS_QUAY_AUTH` guard, `timeout-minutes: 120`, pin merge runner to `ubuntu-24.04`
- Widen branch glob from `release-1.*` to `release-*` for future major versions

### local-run.sh

- Remove `--platform=linux/amd64` from `podman pull` (auto-detects host arch)

### Verified locally

Both images built and all tools (helm, oc, oc-mirror, ocm, yq, aws, az, gcloud, umoci, opm, operator-sdk, node, yarn, psql, skopeo, podman) confirmed working on both architectures.

https://redhat.atlassian.net/browse/RHIDP-13974

### DIFF
--- a/.github/workflows/push-e2e-runner.yaml
+++ b/.github/workflows/push-e2e-runner.yaml
@@ -25,12 +25,17 @@ env:
 
 jobs:
   build-image:
-    name: Build & Push e2e-runner Image
+    name: Build e2e-runner (${{ matrix.os }})
+    env:
+      HAS_QUAY_AUTH: ${{ secrets.QUAY_USERNAME != '' && secrets.QUAY_TOKEN != '' }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04]
+        os:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 120
     permissions:
       contents: read
       packages: write
@@ -44,10 +49,22 @@ jobs:
       - name: Prepare Environment Variables
         env:
           INPUT_BRANCH: ${{ inputs.branch }}
+          MATRIX_OS: ${{ matrix.os }}
         run: |
-          echo "PLATFORM=linux/amd64" >> $GITHUB_ENV
+          # Derive platform from runner OS
+          if [ "$MATRIX_OS" == "ubuntu-24.04" ]; then
+            platform="linux/amd64"
+          elif [ "$MATRIX_OS" == "ubuntu-24.04-arm" ]; then
+            platform="linux/arm64"
+          else
+            echo "::error::Unknown runner OS: $MATRIX_OS"
+            exit 1
+          fi
+          echo "PLATFORM=$platform" >> $GITHUB_ENV
+          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
+          echo "PLATFORM_ARCH=${platform#*/}" >> $GITHUB_ENV
 
-          # create image tag from the correct branch (either from a push or a workflow_dispatch trigger)
+          # Create image tag from the correct branch (either from a push or a workflow_dispatch trigger)
           if [[ "$INPUT_BRANCH" ]] && [[ "$INPUT_BRANCH" != "NONE" ]]; then
             echo "Switch to $INPUT_BRANCH"
             git checkout "$INPUT_BRANCH"
@@ -58,6 +75,13 @@ jobs:
           fi
           echo "Use IMAGE_TAG = $IMAGE_TAG"
           echo "IMAGE_TAG=$IMAGE_TAG" >> $GITHUB_ENV
+
+      - name: Check Quay credentials
+        run: |
+          if [ "$HAS_QUAY_AUTH" != "true" ]; then
+            echo "::error::Missing QUAY_USERNAME or QUAY_TOKEN secrets"
+            exit 1
+          fi
 
       - name: Get the last commit short SHA
         uses: ./.github/actions/get-sha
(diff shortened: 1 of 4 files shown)


zdrapela:
I believe the issues flagged by Qodo and SonarCloud aren't relevant.
- Qodo - works fine in used environments, or already fixed.
- SonarCloud - we need it for executing the E2E tests.

e2e-tests/local-run.sh
```diff
@@ -357,9 +366,16 @@ if [[ "$CLI_MODE" == "false" ]]; then
   echo ""
 fi
 
-# Pull runner image first (can take a while)
+# Pull runner image (always attempt; fall back to local copy if pull fails)
 log::section "Pulling runner container image"
-podman pull "$RUNNER_IMAGE" --platform=linux/amd64
+if ! podman pull "$RUNNER_IMAGE" 2>/dev/null; then
```
gustavolira: Redirecting the pull's stderr to /dev/null throws away the actual reason it failed (auth, network, no such tag, arch mismatch). When someone hits the `Failed to pull image and no local copy` branch they get a generic message with none of podman's output, which makes a local-run failure hard to debug. Could we let podman's error through here (drop the `2>/dev/null`) and keep the silencing only on the `podman image exists` check below, where you're using it purely as a boolean?

.github/workflows/push-e2e-runner.yaml
```diff
@@ -44,41 +53,138 @@ jobs:
...
+        with:
+          name: digests-${{ env.PLATFORM_PAIR }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    name: Create multi-arch manifest
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-24.04
```
gustavolira, line 135: The `build-image` job got `timeout-minutes: 120`, but this `merge` job has none, so it falls back to the 360-minute default. `docker buildx imagetools create`/`inspect` against Quay can hang on a registry hiccup. Worth adding a small explicit timeout here for consistency with the build job.

e2e-tests/local-run.sh
```diff
@@ -73,6 +78,10 @@ while [[ $# -gt 0 ]]; do
       CLI_TAG_NAME="pr-$2"
       shift 2
       ;;
+    -i | --runner-image)
+      RUNNER_IMAGE="$2"
```
gustavolira, line 82: nit: this flag works because parsing runs after line 5, but a user exporting `RUNNER_IMAGE=...` would get it silently clobbered by the hard assignment at the top. Making line 5 `RUNNER_IMAGE="${RUNNER_IMAGE:-quay.io/rhdh-community/rhdh-e2e-runner:main}"` keeps this flag working and also respects an env override.

.github/workflows/push-e2e-runner.yaml
```diff
@@ -44,41 +53,138 @@ jobs:
...
+
+      - name: Create manifest list and push
+        working-directory: /tmp/digests
+        run: |
+          for tag in \
+            "${REGISTRY}/${REGISTRY_IMAGE}:${IMAGE_TAG}" \
+            "${REGISTRY}/${REGISTRY_IMAGE}:${IMAGE_TAG}-${SHORT_SHA}"; do
+            echo "Creating multi-arch manifest for: $tag"
+            docker buildx imagetools create -t "$tag" \
+              $(printf "${REGISTRY}/${REGISTRY_IMAGE}@sha256:%s " *)
```
gustavolira, line 188: nit: `REGISTRY`/`REGISTRY_IMAGE` are trusted constants today so this is safe, but putting expanded variables in the format-string slot is a footgun if either ever gains a `%`. Cleaner: `$(printf '%s@sha256:%s ' "$REGISTRY/$REGISTRY_IMAGE" *)`.

e2e-tests/local-run.sh
```diff
@@ -436,7 +452,7 @@ echo ""
 CONTAINER_EXIT_CODE=0
 podman run -v "$WORK_DIR":/tmp/rhdh \
   -v "$SCRIPT_DIR/container-init.sh":/tmp/container-init.sh:ro \
-  -it -u root --privileged \
+  -i -u root --privileged \
```
gustavolira, line 456: nit: dropping `-t` is correct here (stdout is piped to `tee`, so there was never a real TTY and `-t` would warn/fail in CI), but a future reader may helpfully add it back. A short comment like `# no -t: stdout is piped to tee and CI has no TTY` would lock in the intent.

.github/workflows/push-e2e-runner.yaml
```diff
@@ -44,41 +53,138 @@ jobs:
...
         uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5.4.0
         with:
           context: .
           file: .ci/images/Dockerfile
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
           tags: |
-            ${{ env.REGISTRY }}/${{ env.REGISTRY_IMAGE }}:${{ env.IMAGE_TAG }}
-            ${{ env.REGISTRY }}/${{ env.REGISTRY_IMAGE }}:${{ env.IMAGE_TAG }}-${{ env.SHORT_SHA }}
+            ${{ env.REGISTRY }}/${{ env.REGISTRY_IMAGE }}:${{ env.IMAGE_TAG }}-${{ env.PLATFORM_ARCH }}
```
gustavolira, line 108: nit: these per-arch tags (`-amd64`, `-arm64`, plus the `-${SHORT_SHA}-<arch>` variants) get pushed permanently, and the `merge` job only layers the clean multi-arch tags on top. Over time Quay will accumulate them. If that's intentional for debugging individual arches, ignore this; otherwise the merge can rely purely on the exported digests without also pushing named per-arch tags.

zdrapela: Yes, this is intentional. What about setting the per-arch tags to expire after a week? That would make it accessible for debugging, but it won't accumulate in Quay.

.github/workflows/push-e2e-runner.yaml
zdrapela: Okay, added timeout of 20 mins.

.github/workflows/push-e2e-runner.yaml
zdrapela: True, but that would still be a rare case. The current implementation is aligned with how `next-build-image.yaml` does it. It'd keep it as it is.

gustavolira (approved):

rm3l:
/override "SonarCloud Code Analysis"

Changes scoped to the E2E runner. Might make sense to explicitly exclude those E2E-related folders from SonarCloud analysis: https://github.com/redhat-developer/rhdh/blob/main/.sonarcloud.properties

PR #5079, opened 2026-07-10
fix(i18n): correct German navigation labels

lokanandaprabhu:
## Summary
- correct German Home and Docs labels in RHDH translations
- align colocated and raw JSON/YAML tab labels with preferred wording

## Test plan
- not run (translation updates only)


christoph-jerolimov (commented):
Thanks @lokanandaprabhu 

As a native speaker and reporter of this isuse I can confirm this is much better. 😏 👍

christoph-jerolimov (approved):

PR #5083, opened 2026-07-10
fix(e2e): restore readiness gates for runtime, auth, and localization

schultzp2020:
## Summary

Restores and hardens RHDH E2E readiness after #5004 / #5022 broke nightlies, then deepens auth/runtime/operator install so defaults (especially lightspeed) cannot recreate the same failure class.

- **Instance readiness** — `globalSetup` classifies `BASE_URL` as `unset` | `router-stub` | `instance-url`; runtime jobs keep a predicted URL + `RUNTIME_AUTO_DEPLOY=true`; auth leaves `BASE_URL` empty so the harness self-deploys.
- **Localization** — drop smoke-test deps that asserted English `"Welcome back!"`.
- **Shared operator install profiles** — `OperatorInstallProfile` + `DynamicPluginsProfile` own `apiVersion`, `flavours: []`, `includes: []`, and `BACKEND_SECRET` for runtime and auth adapters.
- **Auth deploy** — `AuthInstanceDeployer` stages **created → HTTP `/healthcheck` → catalog synced**; config liveness, IdP waits, cookie/session isolation, and OIDC autologout hardenings for reconcile races.
- **CI operator CRs** — pin `flavours: []` on `rhdh-start*.yaml` (same lightspeed invariant as Playwright); unit test locks the seam; fail-closed wait for customization-provider S2I so Deployment image is never `" "`.
- Restore `e2e-tests/.lintstagedrc.js` and document `/test ?` / `/test <job>` for agents/contributors.

```mermaid
flowchart TB
  subgraph globalSetup["Playwright globalSetup"]
    A[classify BASE_URL] --> B{mode?}
    B -->|router-stub, no auto-deploy| C[no-op]
    B -->|RUNTIME_AUTO_DEPLOY| D[ensureRuntimeDeployed]
    D --> E[set instance BASE_URL]
    E --> F["GET /healthcheck"]
    B -->|instance-url| F
    B -->|unset, no auto-deploy| C
  end

  subgraph authDeploy["AuthInstanceDeployer"]
    P[OperatorInstallProfile + DynamicPluginsProfile] --> Q[apply CR / CMs]
    Q --> R[wait Deployment created]
    R --> S["HTTP /healthcheck ≤ 600s"]
    S --> T[wait catalog synced]
  end

  subgraph ciOperator["CI operator nightlies"]
    CR["rhdh-start*.yaml flavours: []"] --> DP[install-dynamic-plugins]
    DP --> Ready[showcase / showcase-rbac]
  end

  globalSetup -.->|auth BASE_URL empty| authDeploy
```

## Why auth / operator were still failing after the gate fix

1. Auth YAML omitted `flavours` / `includes`, so the operator reintroduced defaults (e.g. lightspeed) and readiness stayed on HTTP 503. Runtime already encoded those invariants; profiles close the drift.
2. Auth tests then hit reconcile races (sessionDuration leakage, progressbar/`main` gates, aborted navigations, OIDC refresh-cookie timing) — hardened in follow-up commits.
3. Operator nightly CrashLooped `install-dynamic-plugins` with a `lightspeed-core` sidecar because CI start CRs lacked `flavours: []`. Pinned to match `OperatorInstallProfile`.

## Follow-up (only if operator nightly still red)

`flavours: []` should remove the `lightspeed-core` sidecar. Operator CI still uses helm value-file `includes: [dynamic-plugins.default.yaml]`. Helm is already green with that list — leave it alone unless a retest CrashLoops **without** `lightspeed-core`, then pin `includes: []` (or disable the bad OCI) on the operator path.

## Test plan

Rebased onto `main`; tip is `823bbd91`. Re-trigger nightlies after the force-push:

- [x] `cd e2e-tests && yarn lint` / `yarn fmt:check`
- [ ] `/test e2e-ocp-helm-nightly`
- [ ] `/test e2e-ocp-helm-localization-nightly`
- [ ] `/test e2e-ocp-operator-nightly` — expect no `lightspeed-core`; no CrashLoop on showcase / showcase-rbac
- [ ] `/test e2e-ocp-operator-auth-providers-nightly` — OIDC autologout + provider flows green
- [ ] Localization: no smoke `"Welcome back!"` failure
- [ ] showcase-runtime progresses past globalSetup
- [ ] auth-providers: deploy reaches HTTP ready / synced


.ci/pipelines/jobs/ocp-operator.sh
```diff
@@ -68,14 +68,14 @@ initiate_operator_deployments_osd_gcp() {
...
-  #   - Deploys RHDH via the operator with internal PostgreSQL
-  #   - Configures schema-mode env vars for port-forwarding
+  # Runtime tests self-deploy from Playwright global setup when
+  # RUNTIME_AUTO_DEPLOY=true. Pass the predicted route as BASE_URL so
+  # playwright.config.ts freezes a usable use.baseURL before globalSetup.
+  # globalSetup then deploys via ensureRuntimeDeployed() and healthchecks.
   # Subsequent test files reuse the existing deployment (workers: 1).
   #
   # INSTALL_METHOD=operator is already exported in handle_ocp_operator().
+  export RUNTIME_AUTO_DEPLOY="true"
```
zdrapela: The exported `RUNTIME_AUTO_DEPLOY` should be unset after use, as this should be set only for the `showcase-runtime`. 
Or maybe it can be scoped only to the `testing::run_tests` here.

.ci/pipelines/jobs/ocp-nightly.sh
```diff
@@ -47,16 +47,15 @@ run_standard_deployment_tests() {
...
+  # RUNTIME_AUTO_DEPLOY=true. Pass the predicted route as BASE_URL so
+  # playwright.config.ts freezes a usable use.baseURL before globalSetup
+  # (Playwright resolves config before globalSetup runs). globalSetup then
+  # deploys via ensureRuntimeDeployed() and healthchecks that URL.
   # Subsequent test files reuse the existing deployment (workers: 1).
-  #
-  # The CI wrapper only needs to set environment variables and invoke Playwright.
 
   export INSTALL_METHOD="helm"
+  export RUNTIME_AUTO_DEPLOY="true"
```
zdrapela: The exported `RUNTIME_AUTO_DEPLOY` should be unset after use, as this should be set only for the `showcase-runtime`, and the export will be carried over to the following parts of the Bash scripts.
Or maybe it can be scoped only to the `testing::run_tests` here.

e2e-tests/playwright/utils/runtime-deploy.ts
```diff
@@ -315,13 +317,17 @@ export async function ensureRuntimeDeployed(): Promise<void> {
 
   // Check if deployment already exists and is ready
   const deploymentName = getRhdhDeploymentName();
+  const runtimeUrl = `https://backstage-${releaseName}-${namespace}.${routerBase}`;
```
zdrapela: nit: We already compute this URL in `deployWithOperator`. Would be nice to extract it

e2e-tests/.lintstagedrc.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @type {import('lint-staged').Configuration}
+ */
+export default {
+  "*.sh": "shellcheck --severity=warning --color=always",
```
zdrapela: During the refactor, we lost the bash formatting with `prettier-plugin-sh` in `e2e-tests`. Would be nice to have it back or replace it with `shfmt` for example. Alongside `shellcheck`.

.ci/pipelines/jobs/ocp-operator.sh
schultzp2020: Addressed in ef6615d2: `RUNTIME_AUTO_DEPLOY=true` is now scoped to the `testing::run_tests` invocation only (no lasting export). Readiness also ignores a leaked flag unless `BASE_URL` matches the predicted runtime URL.

.ci/pipelines/jobs/ocp-nightly.sh
schultzp2020: Addressed in ef6615d2: scoped as `RUNTIME_AUTO_DEPLOY=true testing::run_tests …` so it cannot leak into sanity-plugins. Defense-in-depth in `shouldAutoDeployRuntime()` refuses to redeploy when `BASE_URL` is a non-runtime instance.

e2e-tests/playwright/utils/runtime-deploy.ts
schultzp2020: Addressed in ef6615d2: extracted `instance-route-identity.ts` as the single source for helm/operator route + deployment naming. Early-exit, `deployWithHelm` fallback, `deployWithOperator`, and auth `getBackstageUrl` all use it.

e2e-tests/.lintstagedrc.js
schultzp2020: Addressed in ef6615d2: restored `prettier-plugin-sh` beside shellcheck in `e2e-tests/.lintstagedrc.js` (with `.prettierrc.cjs`), while oxfmt continues to format non-shell files.

PR #5100, opened 2026-07-14
chore: clean up EOL release references

jonkoops:
Refs: [RHIDP-13595](https://redhat.atlassian.net/browse/RHIDP-13595)

This PR cleans up references to EOL RHDH releases from the codebase.

The dedicated `pr-1.8.yaml` workflow targeted branches that have been deleted and contained a comment stating it should be removed once 1.10 is live; it is deleted entirely.

The `scripts/update-versions-doc.mjs` script accumulated dead code after #4470 changed it from multi-branch discovery to single-branch invocation. The `skipBranches` array, `listBranchNames()` function, and the associated guard in `main()` are all unreachable now. The `skipBranches` list was never maintained; only 3 entries were added at initial creation in #2808, with no additions since. Rather than extending it, we remove it entirely. The root cause — EOL branches lingering on the remote — will be addressed by deleting them, as @nickboldt originally suggested in https://github.com/redhat-developer/rhdh/pull/2808#issuecomment-2886987162.

The dynamic plugin examples file (`docs/dynamic-plugins/examples.md`) from @gashcrumb contained only examples targeting RHDH 1.2–1.5; none were updated for supported releases. The file and its references from `docs/dynamic-plugins/index.md` and `docs/auth.md` are removed. If updated examples for current RHDH versions are created in the future, a new file can be added.

Broken links to the deleted `release-1.5` branch in `docs/e2e-tests/examples.md` are fixed, and stale Red Hat docs URLs pointing to `/1.7/` and `/1.8/` are updated to `/1.10/`.

**This PR cannot be merged before the EOL branches (`release-1.4` through `release-1.8`) are deleted from the remote.**


scripts/update-versions-doc.mjs
```diff
@@ -1,5 +1,3 @@
-// Assisted by watsonx Code Assistant
```
polasudo, line 1: not exactly sure why we are deleting so much from this script ? is it in need to delete from here ? can we review it if it is working as expected after such delete process

jonkoops: Yes, all of this code is dead code (see PR description). #4470 introduced conditions that make all of this code unreachable.

scripts/update-versions-doc.mjs
```diff
@@ -155,61 +99,47 @@
...
-    const backstageVersion = JSON.parse(backstageJson).version
+  const backstageJson = await getFileFromGithub(repository, targetBranch, backstageJsonPath);
+  const backstageVersion = JSON.parse(backstageJson).version
 
-    if (!backstageVersion) {
-      console.error(`Unable to find backstage version in ${backstageJson}`);
-      process.exit(1);
-    }
+  if (!backstageVersion) {
+    console.error(`Unable to find backstage version in ${backstageJson}`);
```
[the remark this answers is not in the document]
jonkoops: This is existing code; I have better things to do.

scripts/update-versions-doc.mjs
polasudo: superb, thank you

polasudo (approved):

PR #5115, opened 2026-07-15
chore(app-next): add runnable RHDH NFS shell for local development

ciiay:
## Description

Make `app-next` a runnable RHDH **NFS** shell for local development and Module Federation plugin testing.

This PR makes NFS testing easier for [RHIDP-14018](https://redhat.atlassian.net/browse/RHIDP-14018): stock app-next does not yet have an NFS-ready homepage plugin path, so teams cannot verify homepage-dependent features (e.g. unread notifications cards) without ad-hoc setup.

Optional plugins (notifications, signals, etc.) are **not** bundled in the shell — export them into `dynamic-plugins-root` (see README).

### App shell (`packages/app-next/`)

- **`src/App.tsx`** — slim NFS shell: nav, visualizer, catalog/home/scaffolder/search/user-settings, `appDrawerModule`, temporary static `homePageModule`, and `rhdhDynamicFrontendFeaturesLoader()`
- **Removed** static custom sign-in and quickstart-help modules (use **app-auth** + DP quickstart instead)
- **`src/modules/dynamicFeatures/`** — temporary MF loader that also registers **named** `FrontendFeature` exports (needed for global-header’s `globalHeaderModule` and notifications’ `notifications-home-module`)
- **`dynamic-plugins.example.yaml`** — example OCI set (app-auth, app-integrations, global-header, quickstart; homepage/theme left commented — OFS-shaped)
- **`README.md`** — install/run, local config overlay, homepage-card testing (e.g. unread notifications), optional local NFS export of homepage/theme from `rhdh-plugins`
- **`package.json` / `App.test.tsx`** — deps and test mocks aligned with the slim shell

### Repo config

- **`app-config.yaml`** — **reverts** NFS `app.extensions` and local auth that were added earlier; NFS/auth live in `app-config.local.yaml` only (shared config stays safe for legacy `packages/app` / CI)

## Review follow-ups / temporary exceptions

| Item | Status |
|------|--------|
| Don’t bake custom sign-in | Done — use app-auth (+ app-integrations) via DP |
| Don’t statically bake RHDH UX; use example `dynamic-plugins.yaml` | Done for auth/header/quickstart |
| Homepage via DP | Blocked by OFS OCI; static `homePageModule` until NFS overlay or local export |
| Named MF exports | Temporary `rhdhDynamicFrontendFeaturesLoader` |

- Fixes #?

[RHIDP-14018](https://redhat.atlassian.net/browse/RHIDP-14018)

## PR acceptance criteria

### Build and run

- [ ] `yarn install` (from repo root)
- [ ] `cp packages/app-next/dynamic-plugins.example.yaml dynamic-plugins.yaml`
- [ ] `npx -y @red-hat-developer-hub/cli-module-install-dynamic-plugins@0.2.0 install dynamic-plugins-root` (needs `skopeo`)
- [ ] Put NFS `app.extensions` + guest auth in `app-config.local.yaml` (example in `packages/app-next/README.md`)
- [ ] `EXPERIMENTAL_MODULE_FEDERATION=true yarn workspace app-next build`
- [ ] `yarn workspace backend start:next`

### Shell verification

- [ ] Sign in as **guest** (via app-auth DP)
- [ ] `/` renders RHDH homepage (static `homePageModule` + `homePlugin`)
- [ ] Sidebar: Search, Home, Catalog, Scaffolder, Settings
- [ ] Global header + quickstart drawer via DP remotes
- [ ] `/visualizer/tree` shows core + DP extensions

### Homepage cards (optional — e.g. unread notifications)

- [ ] Export notifications (with `notifications-home-module` MF expose) + signals into `dynamic-plugins-root`
- [ ] Enable `home-page-widget:home/unread-notifications` in `app-config.local.yaml`
- [ ] Confirm widget on `/` and in `/visualizer/tree`  
  Details: `packages/app-next/README.md` → **Testing homepage cards**

### Unit tests

- [ ] `yarn workspace app-next test`

## Screenshot

image


packages/app-next/src/modules/signIn/index.ts
```diff
@@ -0,0 +1,14 @@
+import { createFrontendModule } from '@backstage/frontend-plugin-api';
+import { SignInPageBlueprint } from '@backstage/plugin-app-react';
+
+export const signInModule = createFrontendModule({
```
gashcrumb: It would be better to pull in `app-auth` (link to artifact [here](https://github.com/orgs/redhat-developer/packages/container/package/rhdh-plugin-export-overlays%2Fred-hat-developer-hub-backstage-plugin-app-auth)) for this, I believe going with this approach will prevent being able to potentially swap out the default sign-in.  And then for most cases the corresponding `app-integration` plugin would be used to configure source control authentication (link [here](https://github.com/orgs/redhat-developer/packages/container/package/rhdh-plugin-export-overlays%2Fred-hat-developer-hub-backstage-plugin-app-integrations)).

packages/app-next/src/App.tsx
```diff
@@ -5,15 +5,48 @@ import scaffolderPlugin from '@backstage/plugin-scaffolder/alpha';
...
+import { quickstartHelpModule } from './modules/quickstartHelp';
+import { signInModule } from './modules/signIn';
 
 const app = createApp({
   features: [
-    appVisualizerPlugin, 
-    catalogPlugin, 
-    scaffolderPlugin, 
-    searchPlugin, 
+    rhdhThemeModule,
```
gashcrumb: Shouldn't these all be provided by dynamic plugins instead?

gashcrumb: With the exception of the catalogPlugin, scaffolderPlugin, searchPlugin and userSettingsPlugin, only because scaffolder-backend can't be provided via a dynamic plugin at the moment, and so the related plugins may as well stay in place.  But I thought generally we wanted to avoid baking too much into the RHDH frontend/backend and rely on providing functionality via dynamic plugins.

app-config.yaml
```diff
@@ -3,6 +3,52 @@ developerHub:
...
+    - api:home/visits: true
+    - app-root-element:home/visit-listener: true
+    - home-page-layout:home/dynamic-homepage-layout:
+        config:
+          customizable: true
+          widgetLayout:
+            RhdhTemplateSection:
+              priority: 300
+              breakpoints:
+                xl: { w: 12, h: 5 }
```
gashcrumb: Probably not exactly related to this particular PR but it seems odd that we're still exposing this level of layout customization via yaml, how often is this actually customized by users?  All of this breakpoint config seems like it'd be better left in the plugin code and customized via css instead.

gashcrumb (changes requested):
I would avoid adding these plugins to the app initialization and instead rely on the frontend dynamic feature loader discovery and dynamic plugins.  We could instead include an example "dynamic-plugns.yaml" file that could be used to populate the default set of dynamic plugins that make up the typical RHDH user experience.  Otherwise we'll wind up having to removing these later on when users find they are unable to swap out plugins.

packages/app-next/package.json
```diff
@@ -72,7 +76,11 @@
     "@material-ui/core": "4.12.4",
     "@material-ui/icons": "4.11.3",
     "@material-ui/lab": "4.0.0-alpha.61",
+    "@module-federation/enhanced": "2.5.1",
+    "@mui/icons-material": "5.18.0",
     "@octokit/rest": "19.0.13",
+    "@red-hat-developer-hub/backstage-plugin-app-react": "0.0.5",
+    "@red-hat-developer-hub/backstage-plugin-homepage": "1.15.0",
```
christoph-jerolimov: We don't want that this are direct dependencies, we want load them as "dynamic plugins" from container images.

christoph-jerolimov (changes requested):
We need a way to make testing easier, but without adding new dependencies from rhdh-plugins here.

packages/app-next/package.json
```diff
@@ -51,15 +52,18 @@
...
     "@backstage/plugin-auth-react": "0.1.28",
-    "@backstage/plugin-catalog": "2.0.6",
+    "@backstage/plugin-catalog": "2.0.7",
     "@backstage/plugin-catalog-common": "1.1.10",
     "@backstage/plugin-catalog-graph": "0.6.5",
     "@backstage/plugin-catalog-import": "0.13.14",
-    "@backstage/plugin-catalog-react": "3.1.0",
+    "@backstage/plugin-catalog-react": "3.2.0",
     "@backstage/plugin-catalog-unprocessed-entities": "0.2.32",
+    "@backstage/plugin-home": "0.9.4",
```
debsmita1, line 66: @backstage/plugin-home was intentionally removed in https://github.com/redhat-developer/rhdh/pull/4921. https://github.com/redhat-developer/rhdh-plugins/pull/4032 updates the dynamic homepage so it includes the home plugin itself, so you no longer need to install @backstage/plugin-home separately.

cc @gashcrumb @rohitkrai03 @christoph-jerolimov

debsmita1 (changes requested):

PR #4860, opened 2026-05-20
fix(ci): dynamically resolve catalog index image from branch + remove trigger-nightly-job command

zdrapela:
Jira: https://redhat.atlassian.net/browse/RHIDP-14348

## Summary

Dynamically resolve the catalog index image from `RELEASE_BRANCH_NAME` instead of hardcoding it, and add Gangway API override support for catalog index and chart version.

Also replaces the `trigger-nightly-job` agent command with an install guide pointing to the external skill in rhdh-skills.

### Dynamic catalog index resolution
- **`env_variables.sh`**: Add `RELEASE_VERSION` canonical variable (`main`→`next`, `release-X.Y`→`X.Y`) and single `CATALOG_INDEX_IMAGE` variable with fallback default (`quay.io/rhdh/plugin-catalog-index:$RELEASE_VERSION`). Parse into `CATALOG_INDEX_REGISTRY`/`CATALOG_INDEX_REPO`/`CATALOG_INDEX_TAG` for Helm chart consumption.
- **5 operator Backstage CRs**: Replace hardcoded `quay.io/rhdh/plugin-catalog-index:1.10` with `$CATALOG_INDEX_IMAGE`
- **`lib/helm.sh`**: Simplified — uses pre-parsed catalog index variables from `env_variables.sh`

### Gangway override support
- **`trigger-nightly-job.sh`**: Add `--catalog-index-image` and `--chart-version` flags with `MULTISTAGE_PARAM_OVERRIDE_CATALOG_INDEX_IMAGE` Gangway env var

### Agent command update
- Replace `trigger-nightly-job` command content with install guide pointing to [rhdh-skill/prow-trigger-nightly](https://github.com/redhat-developer/rhdh-skill/tree/main/skills/prow-trigger-nightly)

## Companion PR
- openshift/release: Gangway override declaration for `MULTISTAGE_PARAM_OVERRIDE_CATALOG_INDEX_IMAGE` (independent, merge order doesn&#39;t matter due to fallback defaults)

## Testing
- Trigger OCP Helm + Operator nightly jobs to verify

PR #5122, opened 2026-07-16
fix(ci): derive catalog index image dynamically from branch

jonkoops:
PR #4860 ([RHIDP-14348](https://redhat.atlassian.net/browse/RHIDP-14348)) added `RELEASE_VERSION` and `CATALOG_INDEX_IMAGE` infrastructure to `env_variables.sh` with the stated goal of replacing the hardcoded catalog index tag, but left the default empty. #5263 added a temporary `CATALOG_INDEX_IMAGE_OVERRIDE` pin to `:1.10` as an interim measure while the catalog index on `:next` was broken, and #5282 moved the pin to `:2.0` after migrating wrappers to `ref://` references.

This PR removes the override and populates the default from `RELEASE_VERSION`, giving us `main` → `:next`, `release-1.10` → `:1.10` automatically. Gangway overrides for RC/GA verification continue to work; an explicit `CATALOG_INDEX_IMAGE` provided externally takes precedence over the default.

[RHIDP-14348](https://redhat.atlassian.net/browse/RHIDP-14348)

### DIFF
--- a/.ci/pipelines/env_variables.sh
+++ b/.ci/pipelines/env_variables.sh
@@ -56,17 +56,17 @@ IMAGE_REPO="${IMAGE_REPO:-${QUAY_REPO:-rhdh-community/rhdh}}"
 QUAY_REPO="${IMAGE_REPO}" # Keep QUAY_REPO in sync for backward compatibility
 
 # Catalog index image reference.
+# Defaults to quay.io/rhdh/plugin-catalog-index tagged with RELEASE_VERSION so that
+# main uses :next and release branches (e.g., release-1.10) use :1.10 automatically.
 # Override via Gangway for RC (e.g., --catalog-index-image quay.io/rhdh/plugin-catalog-index:1.9-60) or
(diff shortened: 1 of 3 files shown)


nickboldt (approved):
works for me, as we'll always have a :x.y tag even if :next and :latest float around

polasudo (approved):

polasudo (approved):

jonkoops:
@nickboldt please don't sync the branch with `main`, especially not with merge commits. The problem is that the catalog for 1.11 is borked currently, which is causing the test failures.

jonkoops:
Moving to draft while the prerequisite wrapper removal PR is prepared. This PR will be rebased on top of it once the wrappers have been replaced with `ref://` references and the catalog index is switched to `:2.0`. The catalog index issues that originally blocked this ([RHIDP-15881](https://redhat.atlassian.net/browse/RHIDP-15881)) have been resolved.

jonkoops:
With the 2.0 catalog now functional, this should be ready to be reviewed.

PR #5135, opened 2026-07-21
refactor(e2e): rename @cluster-free tag to @cluster-free-capable

gustavolira:
## Summary

Renames the `@cluster-free` Playwright tag to `@cluster-free-capable`, and documents it.

The tag reads as a statement about **how a run executed**, but it is a property of the **test**: it marks specs validated to also work on the cluster-free harness, and `playwright.legacy-local.config.ts` selects on it (`grep: /@cluster-free-capable/u`).

Playwright shows a test's tags in *every* report, so a **cluster** run's report shows a `cluster-free` chip next to tests that ran on the cluster. That has now twice been misread as "the cluster-free harness ran these". What actually identifies the execution mode is the **project name** — e.g. `showcase-sanity-plugins`, `smoke-test`.

`@cluster-free-capable` is true in either report, and matches the house style of `@layer3-equivalent`.

## Why not just remove the tag

It selects per **test**, which the file-level `testMatch` allowlist cannot express. For example, three of the four tests in `home-page-customization.spec.ts` carry it and one (`Verify Customized Quick Access`) does not. Dropping the tag would silently widen the cluster-free run to tests that were never validated for it.

## Also

`e2e-tests/README.md` documented every other tag (`@smoke`, `@layer3-equivalent`, `@ga-plugin`, `@non-ga-plugin`, `@blocked`) but not this one — a gap that made the meaning easy to guess wrong. The new entry spells out that it describes a capability, not an execution mode.

## Verification

- `playwright.legacy-local.config.ts --list` selects the **same 14 tests in 10 files** before and after.
- `oxlint`, `oxfmt`, `tsc`, `vitest` clean.

## Note

Mechanical rename, no behaviour change. Touches `playwright.legacy-local.config.ts` and `docs/e2e-tests/local-e2e-harness.md`, which #4967 also touches — but in non-overlapping regions (this PR: config lines 53/72 and doc lines 74-75; #4967: config lines 1-7/22-48/99-125 and doc lines ~32-38), so the two merge cleanly in either order.

### DIFF
--- a/docs/e2e-tests/local-e2e-harness.md
+++ b/docs/e2e-tests/local-e2e-harness.md
@@ -64,8 +64,8 @@ dev server with `app-config.yaml` + `app-config.dynamic-plugins.yaml` +
 `app-config.local-e2e.yaml`. A `globalSetup` first fails fast with the populate command
 if `dynamic-plugins-root` has no plugins.
 
-The run is scoped to tests tagged `@cluster-free` within the spec files allowlisted in
-`testMatch`. To widen coverage, tag a validated test with `@cluster-free` and add its
+The run is scoped to tests tagged `@cluster-free-capable` within the spec files allowlisted in
+`testMatch`. To widen coverage, tag a validated test with `@cluster-free-capable` and add its
 spec file to `testMatch`; if the test needs extra plugins, add them (with their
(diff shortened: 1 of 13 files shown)


subhashkhileri (approved):
/lgtm

PR #5140, opened 2026-07-22
docs: add a guide for writing Layer 1-3 tests (RHIDP-13234)

gustavolira:
## Description

Implements **[RHIDP-13234](https://redhat.atlassian.net/browse/RHIDP-13234)**.

No document in this repository explained **how to write a test**. `mockServices`,
`renderInTestApp`, `TestApiProvider` and `@backstage/frontend-test-utils` appear in
zero docs today, so the only way to learn the patterns is to find a neighbouring test
file and infer them.

`docs/testing.md` fills exactly that gap and nothing else.

## What it covers

Per layer: which utilities to import, which file to copy, and what to assert. It links
to verified in-repo examples rather than pasting code that would drift from them.

| Layer | Worked example |
|---|---|
| L1 unit | `plugins/scalprum-backend/src/service/router.test.ts` |
| L2 backend integration | `plugins/scalprum-backend/src/service/router.integration.test.ts` |
| L3 component | `packages/app/src/components/learningPaths/LearningPathsPage.test.tsx` (+ two smaller components) |

For L2 it also states what *not* to assert: the HTTP contract is the subject; that the
plugin booted is already implied by the request succeeding.

## What it deliberately does not cover

It does **not** restate the layer ladder or the "which layer?" decision framework. Both
already exist — in [`docs/e2e-tests/layer-migration-matrix.md`](https://github.com/redhat-developer/rhdh/blob/main/docs/e2e-tests/layer-migration-matrix.md)
and in the `test-placement` skill — and a third copy would rot out of sync. The guide
links to them.

L4a and L4b are links only, to `docs/e2e-tests/local-e2e-harness.md` and
`e2e-tests/README.md`.

## Two additions worth flagging for review

**Terminology.** The guide disambiguates two names the team uses for tests at different
layers: "a frontend test" is usually **L3**, and only becomes **L4a** when a dynamic
plugin must load into a real app shell; "an integration test" may mean **L2** or
**L4a/L4b**, which differ by orders of magnitude in cost.

**Coverage.** It documents that Codecov is informational only — both the `project` and
`patch` statuses in `codecov.yml` set `informational: true`, so no coverage number can
block a PR and there is no threshold to meet. This is deliberate, and worth stating
explicitly so that a test is justified by the failure it catches rather than by the
coverage percentage it adds.

## Scope note

The ticket originally also asked for a plugin support-level matrix and a third-party
wrapper model. Both belong to **RHIDP-13497** ("Plugin Testing by Support Level") and
have been moved there — the draft matrix additionally specifies coverage thresholds
("Layer 1: 60% minimum, required for release") that do not exist in this repo, so
publishing it here would have documented a gate we deliberately do not have. The ticket
scope has been updated accordingly.

## Testing

- [x] All 10 relative links resolve
- [x] `prettier:check` and `lint:check` pass
- [x] `mkdocs.yaml` has no `nav:` block — auto-discovery, no nav edit needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ Information on running RHDH can be found in our [Getting Started](https://github
 
 We are excited to see people wanting to contribute to our project and welcome anyone who wishes to participate. You are more than welcome to browse through our open issues and tackle anything you feel confident in working on.
 
+When your change needs a test, [Writing tests in this repository](docs/testing.md) covers which utilities to use at each layer and which existing test to copy.
+
 We also welcome non code contributions in the form of bug reporting and documentation writing. If you run across any bugs, please raise an issue here in [JIRA](https://issues.redhat.com/browse/RHDHBUGS).
 
 ## Community, Discussion, and Support
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,0 +1,130 @@
+# Writing tests in this repository
+
+This guide covers **how to write** a test at Layers 1–3 in `rhdh`: which utilities to
+import, which file to copy, and what to assert.
+
+It deliberately does **not** re-explain _which_ layer to choose — that already exists:
+
+- [`docs/e2e-tests/layer-migration-matrix.md`](e2e-tests/layer-migration-matrix.md) —
+  layer definitions and the migration rule of thumb.
+- The `test-placement` skill in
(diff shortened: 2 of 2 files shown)


hopehadfield (approved):
/lgtm

PR #5204, opened 2026-08-03
docs: document ref:// plugin reference syntax

jonkoops:
Documents the `ref://` plugin reference syntax for looking up OCI plugins from an included configuration by image name. Updates the `package` field description and the default plugin activation example to prefer `ref://`, and replaces the "OCI Package Version Inheritance" section with "Plugin References": `ref://` as the recommended form, and `{{inherit}}` when an explicit `!plugin-path` is required (including path omission with auto-detection).

[RHIDP-15876](https://redhat.atlassian.net/browse/RHIDP-15876)

### DIFF
--- a/docs/dynamic-plugins/installing-plugins.md
+++ b/docs/dynamic-plugins/installing-plugins.md
@@ -7,7 +7,7 @@ For more information, see [Installing Dynamic Plugins with the Red Hat Developer
 
 Plugins are defined in the `plugins` array in the `dynamic-plugins.yaml` file. Each plugin is defined as an object with the following properties:
 
-- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path. For OCI packages ONLY, the tag or digest can be replaced by the `{{inherit}}` tag to inherit the version from an included configuration. Additionally, when using single-plugin OCI images, the plugin path can also be omitted.
+- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path. You can also use `ref://plugin-name` to reference a plugin by name from an included configuration (see [Plugin References](#plugin-references)). For OCI packages ONLY, the tag or digest can be replaced by the `{{inherit}}` tag to inherit the version from an included configuration. Additionally, when using single-plugin OCI images, the plugin path can also be omitted.
 - `enabled`: A boolean value that determines whether the plugin is enabled (`true`) or disabled (`false`). The legacy `disabled` field is still accepted for backward compatibility; when both are present, `enabled` takes precedence.
 - `integrity`: The integrity hash of the package. This is required for `tgz` archives and npm packages.
 - `pluginConfig`: The configuration for the plugin. For backend plugins this is optional and can be used to pass configuration to the plugin. For frontend plugins this is required, see [Frontend Plugin Wiring](frontend-plugin-wiring.md) for more information on how to configure bindings and routes. This is a fragment of the `app-config.yaml` file. Anything that is added to this object will be merged into a `app-config.dynamic-plugins.yaml` file whose config can be merged with the main `app-config.yaml` config when launching RHDH.
@@ -24,13 +24,23 @@ On application start, for each plugin that is not enabled, the `install-dynamic-
 ======= Skipping disabled dynamic plugin oci://registry.access.redhat.com/rhdh/backstage-community-plugin-analytics-provider-segment
 ```
 
-To activate this plugin, simply add a package with the same name and set `enabled: true`.
+To activate this plugin, reference it by name using `ref://` and set `enabled: true`.
+
+```yaml
+plugins:
+  - enabled: true
+    package: ref://backstage-community-plugin-analytics-provider-segment
+```
+
+This is equivalent to specifying the full OCI URL with `{{inherit}}`, but avoids repeating the registry path:
 
 ```yaml
 plugins:
   - enabled: true
     package: oci://registry.access.redhat.com/rhdh/backstage-community-plugin-analytics-provider-segment:{{inherit}}
 ```
+
+See [Plugin References](#plugin-references) for more details on `ref://` syntax.
 
 While the plugin's default configuration comes from the `dynamic-plugins.default.yaml` file (now delivered via the catalog index image; see [Using a Catalog Index Image](#using-a-catalog-index-image-for-default-plugin-configurations) below), you still have the option to override it by incorporating a `pluginConfig` entry into the plugin configuration.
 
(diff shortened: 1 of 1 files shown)


docs/dynamic-plugins/installing-plugins.md
```diff
@@ -171,6 +181,31 @@ When the path is omitted, the installer will inspect the OCI image manifest for
...
+  - enabled: true
+    package: ref://backstage-plugin-catalog
+    pluginConfig:
+      catalog:
+        # your config overrides
+```
+
+This resolves to the full package URL from the included configuration, so you do not need to know the registry, path, or version.
+
+`ref://` supports OCI, HTTP(S), and local path targets. NPM package references are not supported as `ref://` targets.
```
gazarenkov, line 203: Note, we support only oci: based catalog and it is not going to change AFAIK.
In general, I would remove this sentence.

jonkoops: Removed. Catalog entries are OCI-only, so calling out other package forms was noise.

polasudo (approved):

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -7,7 +7,7 @@ For more information, see [Installing Dynamic Plugins with the Red Hat Developer
 
 Plugins are defined in the `plugins` array in the `dynamic-plugins.yaml` file. Each plugin is defined as an object with the following properties:
 
-- `package`: The package definition of the plugin. This can be an OCI image, `tgz` archive, npm package, or a directory path. For OCI packages ONLY, the tag or digest can be replaced by the `{{inherit}}` tag to inherit the version from an included configuration. Additionally, when using single-plugin OCI images, the plugin path can also be omitted.
+- `package`: The plugin source: an OCI image, `tgz` archive, npm package, or directory path. For OCI plugins from an included configuration, use `ref://` or `{{inherit}}` (see [Plugin References](#plugin-references)).
```
nickboldt: > from an included configuration

this seems vague to me. Do you mean from a helm chart or a operator-bundle referencing the catalog index, or do you mean from a configmap / customresource yaml that defines install params?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -24,14 +24,16 @@ On application start, for each plugin that is not enabled, the `install-dynamic-
 ======= Skipping disabled dynamic plugin oci://registry.access.redhat.com/rhdh/backstage-community-plugin-analytics-provider-segment
 ```
 
-To activate this plugin, simply add a package with the same name and set `enabled: true`.
+To activate this plugin, reference it from the included configuration with `ref://` and set `enabled: true`:
```
nickboldt: again, "the included configuration" is vague to me. Details, examples, links?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -171,67 +173,64 @@ When the path is omitted, the installer will inspect the OCI image manifest for
 
 Images MUST be packaged with the `@red-hat-developer-hub/cli` to ensure the proper `io.backstage.dynamic-packages` annotation is applied.
 
-#### OCI Package Version Inheritance
+#### Plugin References
+
+When enabling or configuring plugins from an included configuration (such as the default catalog), you can reference those entries instead of hard-coding a full OCI URL.
```
nickboldt: at least here you clarify "included configuration (such as the default catalog)" but we could do better since there are OTHER ways to define these refs, right?

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -171,67 +173,64 @@ When the path is omitted, the installer will inspect the OCI image manifest for
...
+
+That keeps your overrides stable when the include changes. A newer catalog or default config may ship a different registry, digest, or tag for the same image name; your `ref://` (or `{{inherit}}`) entry still resolves to whatever the include currently provides, so you do not need to chase image references in every consumer config.
```
nickboldt:
```suggestion
This will ensure that your custom configuration's overrides refer to the correctly updated artifacts, if your installed version of RHDH, or the referenced catalog, is updated. A newer catalog or default config may ship a different registry, digest, or tag for the same image name; your `ref://` (or `{{inherit}}`) entry still resolves to whatever the include currently provides, so you do not need to chase image references in every consumer config.
```

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -171,67 +173,64 @@ When the path is omitted, the installer will inspect the OCI image manifest for
...
 
-When working with OCI-packaged dynamic plugins, you may want to avoid specifying the version (tag or digest) in multiple places, especially when including plugins from other configuration files such as `dynamic-plugins.default.yaml`. Setting the tag of the OCI package to `{{inherit}}` allows a plugin configuration override to inherit the plugin version from an included configuration.
+##### `ref://` (recommended)
```
nickboldt:
```suggestion
##### `ref://plugin-name` (recommended)
```

docs/dynamic-plugins/installing-plugins.md
```diff
@@ -171,67 +173,64 @@ When the path is omitted, the installer will inspect the OCI image manifest for
...
+# your override
+plugins:
+  - enabled: true
+    package: oci://quay.io/example/backstage-plugin-myplugin:{{inherit}}!backstage-plugin-myplugin-backend
+    pluginConfig:
+      exampleName: test
 ```
 
-Then we can just use `{{inherit}}` without a path, and we will inherit both the version `v0.0.2` and the plugin path `example-path`
+You can also omit `!plugin-path` on the `{{inherit}}` entry. The path then comes from the include, or from single-plugin image [auto-detection](#oci-package-plugin-path-auto-detection) when the include omitted it too. This only works when exactly one plugin from that image appears in the includes.
```
nickboldt:
```suggestion
You can also omit `!plugin-path` on the `{{inherit}}` entry. The path then comes from the include, or from single-plugin image [auto-detection](#oci-package-plugin-path-auto-detection) when the include omitted it too. This only works when the included artifact contains only a single plugin.
```

docs/dynamic-plugins/installing-plugins.md
jonkoops: Clarified: this means plugins already listed under `includes` (usually catalog defaults), not Helm/Operator packaging. Deployment placement is covered earlier on the page.

docs/dynamic-plugins/installing-plugins.md
jonkoops: Updated to say your `dynamic-plugins.yaml` matches an image name from the catalog defaults under `includes`.

docs/dynamic-plugins/installing-plugins.md
jonkoops: Rewrote to name `includes`, with `dynamic-plugins.default.yaml` from the catalog index as the typical case. Other include files work the same way; we did not expand beyond that.

docs/dynamic-plugins/installing-plugins.md
jonkoops: Took the intent, kept it shorter: overrides stay stable when an include file's registry/digest/tag changes. Avoided mixing in "installed RHDH version."

docs/dynamic-plugins/installing-plugins.md
jonkoops: Done.

docs/dynamic-plugins/installing-plugins.md
jonkoops: Kept the `includes` constraint (one matching entry). Added that pathless includes also need a single-plugin image for auto-detection; that matches the auto-detect section rather than replacing the merge rule.

PR #5211, opened 2026-08-04
fix(ci): harden GKE readiness via /healthcheck probes

zdrapela:
## Summary
- Probe `GET /healthcheck` (not `HEAD /`) in `testing::check_backstage_running` so CI readiness matches what Playwright uses and retries through transient TLS/connect failures.
- Re-probe `/healthcheck` for up to ~30s (warn-only) immediately before `yarn playwright test` after yarn install, via `common::retry`.
- Harden `testing::probe_rhdh_healthcheck` with curl connect/max timeouts, `jq` status validation, and `_TESTING_LAST_HEALTH_DETAIL` for triage logs.
- Harden `instance-health-check.spec.ts` with `expect.poll` (~30s), try/catch, JSON content-type check, and a clear `lastDetail` timeout error ([RHDHBUGS-3508](https://redhat.atlassian.net/browse/RHDHBUGS-3508)).

Follow-up: slim `testing.sh`-only PR targeting `main` (does not absorb #5083 Playwright work).

GKE Helm nightly stability on this branch: consecutive `/test e2e-gke-helm-nightly` passes before this review round.

## Test plan
- [x] `bash -n .ci/pipelines/lib/testing.sh`
- [x] pre-commit: shellcheck + e2e lint/tsc/prettier
- [x] Wait for GKE Helm nightly / rehearse on this PR if available
- [ ] Confirm nightlies no longer fail solely on early `/healthcheck` TLS disconnects

### DIFF
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -2,7 +2,7 @@
 
 # Testing utilities for CI pipelines
 # Handles Playwright test execution, Backstage health checks, and upgrade verification
-# Dependencies: oc, kubectl, yarn, playwright, lib/log.sh
+# Dependencies: oc, kubectl, yarn, playwright, curl, jq, lib/log.sh, lib/common.sh
 
 # Prevent re-sourcing
 if [[ -n "${TESTING_LIB_SOURCED:-}" ]]; then
@@ -12,6 +12,8 @@ readonly TESTING_LIB_SOURCED=1
 
 # shellcheck source=.ci/pipelines/lib/log.sh
 source "${DIR}/lib/log.sh"
+# shellcheck source=.ci/pipelines/lib/common.sh
+source "${DIR}/lib/common.sh"
 # shellcheck source=.ci/pipelines/lib/test-run-tracker.sh
 source "${DIR}/lib/test-run-tracker.sh"
 
@@ -75,20 +77,12 @@ testing::run_tests() {
 
   # Quick re-check after yarn install; do not fail the job — deploy readiness
   # already waited. Warn-only so a brief TLS blip does not abort the suite.
+  local health_retries=6 health_backoff_seconds=5
   if [[ -n "${url}" ]]; then
-    local ready=false
-    local i
-    for ((i = 1; i <= 6; i++)); do
-      if testing::probe_rhdh_healthcheck "${url}"; then
-        ready=true
-        break
-      fi
-      sleep 5
-    done
-    if [[ "${ready}" == "true" ]]; then
+    if common::retry "${health_retries}" "${health_backoff_seconds}" testing::probe_rhdh_healthcheck "${url}"; then
       log::success "Pre-Playwright /healthcheck OK"
     else
-      log::warn "Pre-Playwright /healthcheck still flaky after ~30s; continuing to tests"
+      log::warn "Pre-Playwright /healthcheck still flaky after ~$((health_retries * health_backoff_seconds))s; continuing to tests"
     fi
   fi
 
@@ -157,10 +151,11 @@ testing::run_tests() {
 
 # Probe GET ${url}/healthcheck for a JSON liveness response.
 # Connect/TLS failures and non-OK bodies are treated as not ready (return 1).
+# Sets _TESTING_LAST_HEALTH_DETAIL for caller logging (e.g. "HTTP 000", "jq status!=ok").
 # Args:
 #   $1 - url: Base URL of the RHDH instance (no trailing path required)
 # Returns:
-#   0 - HTTP 200 and body contains "status":"ok"
+#   0 - HTTP 200 and JSON body .status == "ok"
 #   1 - Not ready
 testing::probe_rhdh_healthcheck() {
   local url=$1
@@ -168,18 +163,24 @@ testing::probe_rhdh_healthcheck() {
   local response http_status body
 
   # Append http_code on its own line so connect/TLS failures can be retried.
-  response=$(curl --insecure -sS -w "\n%{http_code}" "${health_url}" 2> /dev/null || true)
-  if [[ -z "${response}" ]]; then
-    return 1
-  fi
+  # Connect failures typically yield "\n000"; bounds avoid hung probes.
+  response=$(curl --insecure -s --connect-timeout 5 --max-time 15 -w "\n%{http_code}" "${health_url}" 2> /dev/null || true)
 
   http_status=$(printf '%s' "${response}" | tail -n 1)
   body=$(printf '%s' "${response}" | sed '$d')
 
-  if [[ "${http_status}" == "200" ]] && [[ "${body}" =~ \"status\"[[:space:]]*:[[:space:]]*\"ok\" ]]; then
-    return 0
+  if [[ "${http_status}" != "200" ]]; then
+    _TESTING_LAST_HEALTH_DETAIL="HTTP ${http_status:-000}"
+    return 1
   fi
-  return 1
+
+  if ! printf '%s' "${body}" | jq -e '.status == "ok"' > /dev/null 2>&1; then
+    _TESTING_LAST_HEALTH_DETAIL="jq status!=ok"
+    return 1
+  fi
+
+  _TESTING_LAST_HEALTH_DETAIL="HTTP 200"
+  return 0
 }
 
 # Check if Backstage is up and running at the given URL
@@ -216,7 +217,7 @@ testing::check_backstage_running() {
       log::success "Backstage is up and running!"
       return 0
     else
-      log::warn "Attempt ${i} of ${max_attempts}: Backstage /healthcheck not yet available"
+      log::warn "Attempt ${i} of ${max_attempts}: Backstage /healthcheck not yet available (${_TESTING_LAST_HEALTH_DETAIL:-unknown})"
       oc get pods -n "${namespace}"
 
       # Early crash detection: fail fast if RHDH pods are in CrashLoopBackOff
--- a/.ci/pipelines/lib/testing.sh
+++ b/.ci/pipelines/lib/testing.sh
@@ -73,13 +73,22 @@ testing::run_tests() {
 
   yarn playwright install chromium
 
-  # Re-validate /healthcheck after yarn install so transient GKE TLS blips
-  # between deploy readiness and Playwright start do not fail the suite.
+  # Quick re-check after yarn install; do not fail the job — deploy readiness
+  # already waited. Warn-only so a brief TLS blip does not abort the suite.
   if [[ -n "${url}" ]]; then
-    if ! testing::wait_for_rhdh_healthcheck "${url}" 24 5; then
-      save_overall_result 1
-      test_run_tracker::mark_test_result "false" "healthcheck"
-      return 1
+    local ready=false
+    local i
+    for ((i = 1; i <= 6; i++)); do
+      if testing::probe_rhdh_healthcheck "${url}"; then
+        ready=true
+        break
+      fi
+      sleep 5
+    done
+    if [[ "${ready}" == "true" ]]; then
+      log::success "Pre-Playwright /healthcheck OK"
+    else
+      log::warn "Pre-Playwright /healthcheck still flaky after ~30s; continuing to tests"
     fi
   fi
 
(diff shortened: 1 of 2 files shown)


.ci/pipelines/lib/testing.sh
```diff
@@ -136,6 +155,33 @@ testing::run_tests() {
...
 # ==============================================================================
 
+# Probe GET ${url}/healthcheck for a JSON liveness response.
+# Connect/TLS failures and non-OK bodies are treated as not ready (return 1).
+# Args:
+#   $1 - url: Base URL of the RHDH instance (no trailing path required)
+# Returns:
+#   0 - HTTP 200 and body contains "status":"ok"
+#   1 - Not ready
+testing::probe_rhdh_healthcheck() {
```
gustavolira, line 160: This probe needs to land on `main` too. `origin/main:.ci/pipelines/lib/testing.sh` still does `curl -I` for readiness, and per RHDHBUGS-3508 the flake last hit main on Aug 1.

The ticket says main is waiting on #5083, but that PR has been open since Jul 10, is currently `CONFLICTING`, and touches 90 files — not a realistic vehicle for a CI flake fix. Could we land the slim extract on main directly, as the ticket comment suggests as the alternative? Otherwise 1.11 branches off main and this fix disappears with it.

.ci/pipelines/lib/testing.sh
```diff
@@ -136,6 +155,33 @@ testing::run_tests() {
...
+  local url=$1
+  local health_url="${url%/}/healthcheck"
+  local response http_status body
+
+  # Append http_code on its own line so connect/TLS failures can be retried.
+  response=$(curl --insecure -sS -w "\n%{http_code}" "${health_url}" 2> /dev/null || true)
```
gustavolira: No `--connect-timeout` / `--max-time` here. If the LB accepts the connection but the backend never answers, curl blocks with no ceiling — that turns the GKE budget of `50 x 30s` into unbounded wall clock inside a job with a hard Prow timeout, and makes the pre-Playwright loop open-ended too.

```bash
response=$(curl --insecure -sS --connect-timeout 5 --max-time 15 -w "\n%{http_code}" "${health_url}" 2> /dev/null || true)
```

.ci/pipelines/lib/testing.sh
```diff
@@ -136,6 +155,33 @@ testing::run_tests() {
...
+  if [[ -z "${response}" ]]; then
```
gustavolira: nit: this branch is effectively unreachable. With `-w` set, curl still writes the format string on a connect failure, so `response` comes back as `\n000` rather than empty (I checked against a dead port). The retry actually comes from `http_status != 200` below, not from here — worth either dropping the guard or rewording the comment above it so the next reader doesn't think it's load-bearing.

Same line above: `-S` doesn't buy anything with stderr going to `/dev/null`.

.ci/pipelines/lib/testing.sh
```diff
@@ -136,6 +155,33 @@ testing::run_tests() {
...
+    return 1
+  fi
+
+  http_status=$(printf '%s' "${response}" | tail -n 1)
+  body=$(printf '%s' "${response}" | sed '$d')
+
+  if [[ "${http_status}" == "200" ]] && [[ "${body}" =~ \"status\"[[:space:]]*:[[:space:]]*\"ok\" ]]; then
```
gustavolira: This substring match passes on a nested hit — `{"db":{"status":"ok"},"status":"error"}` reads as healthy. `jq` is already in the CI image (`.ci/images/Dockerfile:39`), so we can be exact and fail closed on malformed JSON at the same time:

```bash
if [[ "${http_status}" == "200" ]] && printf '%s' "${body}" | jq -e '.status == "ok"' > /dev/null 2>&1; then
```

.ci/pipelines/lib/testing.sh
```diff
@@ -73,6 +73,25 @@ testing::run_tests() {
 
   yarn playwright install chromium
 
+  # Quick re-check after yarn install; do not fail the job — deploy readiness
+  # already waited. Warn-only so a brief TLS blip does not abort the suite.
+  if [[ -n "${url}" ]]; then
```
gustavolira, line 81: Should this use `common::retry` (`.ci/pipelines/lib/common.sh:182`) rather than a hand-rolled loop?

```bash
if [[ -n "${url}" ]]; then
  if common::retry 6 5 testing::probe_rhdh_healthcheck "${url}"; then
    log::success "Pre-Playwright /healthcheck OK"
  else
    log::warn "Pre-Playwright /healthcheck still flaky; continuing to tests"
  fi
fi
```

Same behavior, ~15 fewer lines, and consistent logging with the rest of `.ci/`. It also drops the wasted final `sleep 5` — the loop as written sleeps after the 6th failed probe before falling through.

.ci/pipelines/lib/testing.sh
```diff
@@ -73,6 +73,25 @@ testing::run_tests() {
...
+    local ready=false
+    local i
+    for ((i = 1; i <= 6; i++)); do
```
gustavolira: nit: `6` and `5` are bare literals, and the resulting "~30s" lives only in the warn string below — they'll drift apart the first time someone tunes one. A named local (`local health_retries=6 health_backoff_seconds=5`) or a one-line comment on why 30s is the right window would keep them honest.

Related: the PR description still says "up to ~2 minutes" while this does ~30s. Looks stale after the `slim GKE healthcheck harden` commit.

.ci/pipelines/lib/testing.sh
```diff
@@ -161,18 +207,16 @@ testing::check_backstage_running() {
...
-
-    if [[ "${http_status}" -eq 200 ]]; then
+    # GET /healthcheck (not HEAD /) so TLS + JSON liveness match what Playwright uses.
+    # Connect/TLS failures are not-ready and retry within the existing attempt budget.
+    if testing::probe_rhdh_healthcheck "${url}"; then
       log::success "Backstage is up and running!"
       return 0
     else
-      log::warn "Attempt ${i} of ${max_attempts}: Backstage not yet available (HTTP Status: ${http_status})"
+      log::warn "Attempt ${i} of ${max_attempts}: Backstage /healthcheck not yet available"
```
gustavolira: We lost the HTTP status here — this used to print `(HTTP Status: 503)`. That's the one number that separates a TLS disconnect (`000`) from a real `503` or a `200` whose body isn't `ok` yet, which is exactly the distinction RHDHBUGS-3508 is about, so dropping it makes the next occurrence harder to triage than the code being replaced.

#5083 keeps that signal on the TS side via its `detail` string. Could the probe echo the status (or stash it in a namespaced global like `_TESTING_LAST_HEALTH_STATUS`) so this warn can include it?

e2e-tests/playwright/e2e/instance-health-check.spec.ts
```diff
@@ -9,14 +9,27 @@ test.describe("Application health check", () => {
...
-    const healthCheckEndpoint = "/healthcheck";
-
-    const response = await request.get(healthCheckEndpoint);
-
-    const responseBody = await response.json();
-
-    expect(response.status()).toBe(200);
-
-    expect(responseBody).toHaveProperty("status", "ok");
+    // Short poll: one TLS blip should not fail; deploy readiness already waited.
```
gustavolira, line 12: Heads up: this spec doesn't run in `showcase-rbac-k8s`. That project uses `testMatch: ["**/playwright/e2e/**/*-rbac.spec.ts"]` (`playwright.config.ts:151`), and `instance-health-check.spec.ts` doesn't match it — so for one of the two projects the ticket names as affected, this change is a no-op and only the shell probe covers it.

Was the failure there coming through the `smoke-test` dependency instead? If so, is `retries: 10` on smoke-test already absorbing it, or is something still left to fix on that path?

e2e-tests/playwright/e2e/instance-health-check.spec.ts
```diff
@@ -9,14 +9,27 @@ test.describe("Application health check", () => {
...
+    await expect
```
gustavolira: Two things about this poll.

`.toBe(true)` means a genuine regression reports `Expected: true / Received: false` after 30s — no status, no body. The ticket already notes there are no screenshots because nothing ran, so the failure string is the only evidence the next occurrence produces.

And #5083 already solves this on main with `probeHealthcheck` returning `{ok, detail}` and `waitForRhdhReady` rethrowing `RHDH not ready within ${timeoutMs}ms: ${lastDetail}` — so we'd land two different shapes of the same fix and have to reconcile them at cherry-pick time. Could we mirror it here?

```ts
let lastDetail = "no response yet";
await expect
  .poll(
    async () => {
      try {
        const response = await request.get("/healthcheck");
        if (response.status() !== 200) {
          lastDetail = `HTTP ${response.status()}`;
          return false;
        }
        const body: unknown = await response.json();
        const ok =
          typeof body === "object" &&
          body !== null &&
          Reflect.get(body, "status") === "ok";
        lastDetail = ok ? "status ok" : "HTTP 200 unexpected body";
        return ok;
      } catch (error) {
        lastDetail = `request failed: ${error instanceof Error ? error.message : String(error)}`;
        return false;
      }
    },
    { timeout: 30_000, intervals: [2_000] },
  )
  .toBe(true);
```

(The `try` wrapping the whole request is the right call — `expect.poll` runs `await actual()` outside its own try/catch, so a transport throw aborts the poll instead of retrying. That's the actual root cause in the ticket.)

e2e-tests/playwright/e2e/instance-health-check.spec.ts
```diff
@@ -9,14 +9,27 @@ test.describe("Application health check", () => {
...
+      .poll(
+        async () => {
+          try {
+            const response = await request.get("/healthcheck");
+            if (response.status() !== 200) {
```
gustavolira: nit: main's `wait-for-rhdh-ready.ts` — and #5083's rewrite of it — also checks that `content-type` contains `json` before parsing, which guards against a proxy returning an HTML error page with a 200. Worth matching so the 1.10 and main versions don't drift further apart.

.ci/pipelines/lib/testing.sh
zdrapela: Agreed — I’ll open a slim `testing.sh`-only PR targeting `main` with the hardened probe + `check_backstage_running` switch + pre-Playwright `common::retry` block. That path does **not** absorb or block #5083 (Playwright `wait-for-rhdh-ready` / globalSetup stays with Paul). Will link both PRs from the main extract.

.ci/pipelines/lib/testing.sh
zdrapela: Fixed — probe now uses `--connect-timeout 5 --max-time 15`.

.ci/pipelines/lib/testing.sh
zdrapela: Dropped the empty-`response` guard and `-S` (stderr discarded). Comment now notes connect failures typically yield `\n000`.

.ci/pipelines/lib/testing.sh
zdrapela: Done — pre-Playwright re-probe now uses `common::retry` with named `health_retries` / `health_backoff_seconds`.

.ci/pipelines/lib/testing.sh
zdrapela: Switched to `jq -e '.status == "ok"'` for exact top-level status and fail-closed malformed JSON.

.ci/pipelines/lib/testing.sh
zdrapela: Named locals added and PR description updated to ~30s (was stale after the slim commit).

.ci/pipelines/lib/testing.sh
zdrapela: Restored triage signal via `_TESTING_LAST_HEALTH_DETAIL` (e.g. `HTTP 000` / `jq status!=ok` / `HTTP 200`) included in the warn line.

e2e-tests/playwright/e2e/instance-health-check.spec.ts
zdrapela: Acknowledged — this spec is a no-op for `showcase-rbac-k8s` (`testMatch: *-rbac.spec.ts`). The shared gate for that project is the shell probe in `testing::check_backstage_running` / pre-Playwright re-probe. Smoke-test retries remain a separate path and aren’t changed here.

e2e-tests/playwright/e2e/instance-health-check.spec.ts
zdrapela: Added JSON `content-type` check before treating the response as ok.

e2e-tests/playwright/e2e/instance-health-check.spec.ts
zdrapela: Mirrored the detail idea inline: track `lastDetail` and rethrow a clear timeout error after the poll. Kept try/catch so transport throws don’t abort `expect.poll`. Intentionally did **not** add `wait-for-rhdh-ready.ts` on 1.10 (leave that shape to #5083 on main).

.ci/pipelines/lib/testing.sh
zdrapela: Opened: https://github.com/redhat-developer/rhdh/pull/5215 (`testing.sh` only; complementary to #5083).

zdrapela:
@gustavolira I intended this PR as a simple fix, and I didn't want to waste too much time and tokens on it. It fixed the issue perfectly, as you can see in the CI history. But I addressed your comments and opened a cherry-pick of the logic in `testing.sh` into `main`: https://github.com/redhat-developer/rhdh/pull/5215. The `e2e-tests` folder has already drifted significantly due to Paul's previous refactors, and it's not realistic to backport them.

I still expect @schultzp2020 to update his PR https://github.com/redhat-developer/rhdh/pull/5083 when the `plugin-catalog-index` image is finally fixed. The PR fixes too many issues that were caused by the previous refactor PRs. It would be ideal to have his PR split into smaller parts, as some of them can already be verified, reviewed, and merged.

gustavolira (approved):

PR #5242, opened 2026-08-11
fix(e2e): stub pingIdentity and email config for plugin-sanity

nickboldt:
## Summary
- Add dummy `catalog.providers.pingIdentityOrg` and `notifications.processors.email` stubs to `app-config.plugin-sanity.yaml` so cluster-free plugin-sanity can boot when the catalog index enables those modules.
- Matches existing harness pattern (`.invalid` hosts / dummy credentials); email uses debug-only `stream` transport.

## Test plan
- [ ] `plugin-sanity` job reaches readiness 200 (no timeout on `config.webServer`)
- [ ] Backend no longer logs missing `apiPath` / `email.sender` startup errors
- [ ] Scheduled E2E Cluster-free on main recovers after merge

Generated-by: cursor

Ref: https://redhat.atlassian.net/browse/RHIDP-13884

### DIFF
--- a/e2e-tests/local-harness/app-config.plugin-sanity.yaml
+++ b/e2e-tests/local-harness/app-config.plugin-sanity.yaml
@@ -119,3 +119,28 @@ catalog:
             seconds: 15
           timeout:
             minutes: 15
+    # Catalog index enables pingidentity with incomplete placeholder config;
+    # apiPath/authPath/envId are required at init and abort the backend if missing.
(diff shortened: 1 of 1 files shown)


nickboldt:
this PR seeks to fix a known test failure that prevented me from merging https://github.com/redhat-developer/rhdh/pull/5226#issuecomment-5256645127 until I was able to determine the test was NOT blocking my change.

PR #5268, opened 2026-08-18
feat(ci): add disconnected OCP smoke test for Helm and Operator

zdrapela:
Jira: https://redhat.atlassian.net/browse/RHIDP-13974

_Human written_

- Configures two paths - Helm and Operator.
- Each of them has the Local path that uses the internal OCP registry, which helps with development and verification, and the CI path, which uses the external registry prepared alongside the cluster.
- The Operator path uses `prepare-restricted-enviroment.sh` script with `oc-mirror` and avoids using Podman for rebuilding the index. To allow using nested Podman in CI, we would need to use a special image instead of the `e2e-test-runner` image, or adjust the `e2e-test-runner` to include all the specifics of the special image. This will also complicate running it locally, so I settled on the `oc-mirror` path.
- I tried to align on what's recommended in the docs and used workarounds documented in customer issues where needed.
- You can check the history of runs here: https://prow.ci.openshift.org/pr-history/?org=redhat-developer&repo=rhdh&pr=5268
  - the Helm path passed: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-developer_rhdh/5268/pull-ci-redhat-developer-rhdh-main-e2e-ocp-disconnected-helm-nightly/2090430694005149696
  - the Operator path passed: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-developer_rhdh/5268/pull-ci-redhat-developer-rhdh-main-e2e-ocp-disconnected-operator-nightly/2090713947052183552
---
_AI Generated_

Mirrors RHDH images, chart, operator, and dynamic plugins into an isolated mirror registry on a disconnected OCP cluster, deploys RHDH, and runs a Playwright smoke test (guest login → homepage). Covers both **Helm** and **Operator** installs, each runnable in **CI** (Prow bastion mirror) and **Local** (`LOCAL_DISCONNECTED=1`, cluster integrated registry).

## Path selection

```mermaid
flowchart TD
  CI["CI: openshift-ci-tests.sh<br/>*disconnected*{helm,operator}*nightly*"] --> F
  L["Local: local-run.sh<br/>detects *disconnected*, requires oc login<br/>sets DISCONNECTED=true, LOCAL_DISCONNECTED=1"] --> F
  F["lib/disconnected.sh (facade)<br/>sources env/mirror/plugins/namespace/operator/helm"] --> D{LOCAL_DISCONNECTED=1?}
  D -- no --> CIH["CI no-op hooks<br/>external MIRROR_REGISTRY_URL"]
  D -- yes --> LO["source lib/disconnected/local.sh<br/>overrides _hook_* + integrated registry"]
  CIH --> H["handle_ocp_disconnected_helm"]
  CIH --> O["handle_ocp_disconnected_operator"]
  LO --> H
  LO --> O
```

## Paths

**Helm.** Validate env + auth → pull chart (GA from charts.openshift.io, CI from OCI) → `oc-mirror` the hub, PostgreSQL, and catalog-index images → apply IDMS/ITMS and wait for MachineConfigPools → mirror plugins via `mirror-plugins.sh` and re-pin homepage/catalog-index digests → create mirror CA + registry-auth secret and plugin-mirror ConfigMap → `helm upgrade -i` with a post-renderer that appends disconnected volumes (registries.conf, mirror CA) to the rendered Deployment → EBS-safe hub recovery → Playwright smoke.

**Operator.** Validate env + auth → fetch and run `prepare-restricted-environment.sh` with `--use-oc-mirror` (mirrors operator/operand images, installs the OLM v1 CatalogSource) → wait for ClusterCatalog `Serving=True` and the Backstage CRD → mirror plugins via `mirror-plugins.sh` and re-pin digests → create mirror CA + auth + `policy.json` and homepage/plugin-mirror ConfigMaps → render the Backstage CR from an `envsubst` template (extraFiles/extraEnvs for disconnected mounts) → `deploy_rhdh_operator` → Playwright smoke.

**CI vs Local.** CI uses an external bastion mirror (`MIRROR_REGISTRY_URL`) and `prepare`'s default flags plus external CA/pull-secret helpers. Local uses the cluster's integrated registry via `lib/disconnected/local/`: `registry.sh` brings it up (MIRROR_* bootstrap, amd64 skopeo shim), `mirror-host.sh` rewrites the push-route IDMS to the in-cluster registry service, `access.sh` creates push-target projects and workload/OLM pull grants, `plugins.sh` fully overrides plugin mirroring (digest list + imagestream tags), and `hooks.sh` supplies the `_hook_*` / `cluster_mirror_host` overrides.

## Key decisions

- **Modular facade + hook/override** — `lib/disconnected.sh` sources concern-based modules; CI defines no-op `_hook_*`, `local.sh` overrides them for integrated-registry behaviour. Keeps CI functions clean.
- **Reuse rhdh-operator scripts** instead of reimplementing mirroring (see below).
- **oc-mirror over podman** — nested Podman in the CI pod cannot init storage (`newuidmap`/VFS chown fail under `hostUsers: false`), so the operator path forces `--use-oc-mirror`.
- **Helm post-renderer** — a static values file would clobber the chart's whole `extraVolumes[]`; the post-renderer appends disconnected volumes to the rendered Deployment instead.
- **Template-driven YAML** — inline heredocs/yq replaced with `envsubst` templates (Backstage CR, homepage dynamic-plugins, post-renderer patch, skopeo shim).
- **Shared `CATALOG_INDEX_IMAGE` contract** — CI and Local mirror and consume the same index; homepage/catalog-index re-pinned to the actually-mirrored digest.
- **EBS-safe hub recovery** — rolling `oc rollout restart` (not `oc delete pod`) so the PostgreSQL StatefulSet is never disturbed (avoids ~5 min EBS re-attach).
- **TLS/auth/signatures** — per-registry mirror CA mount, chart registry-auth Secret, and permissive `policy.json` for unsigned mirrored images.

## rhdh-operator scripts

Both are fetched at runtime by `disconnected::fetch_operator_repo_script` from `redhat-developer/rhdh-operator` `.rhdh/scripts/` on `refs/heads/${RELEASE_BRANCH_NAME}` (defaults to `main` locally):

| Script | Used by | Purpose |
|---|---|---|
| `mirror-plugins.sh` | Helm + Operator | Mirror dynamic plugins and the catalog index into the mirror registry |
| `prepare-restricted-environment.sh` | Operator only | Mirror operator/operand images and install the OLM v1 CatalogSource |

## Companion PR

- openshift/release#81128 (already merged)


.ci/pipelines/lib/disconnected/env.sh
```diff
@@ -0,0 +1,82 @@
...
+    ref_label="sha: ${ref}"
+  elif [[ "${ref}" =~ ^(refs/)?pull/([0-9]+)(/head)?$ ]]; then
+    # Always the current PR head (updates as the PR is pushed).
+    pull_number="${BASH_REMATCH[2]}"
+    url="https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/pull/${pull_number}/head/.rhdh/scripts/${script_name}"
+    ref_label="pull/${pull_number} head"
+  else
+    url="https://raw.githubusercontent.com/redhat-developer/rhdh-operator/refs/heads/${ref}/.rhdh/scripts/${script_name}"
+    ref_label="branch: ${ref}"
+  fi
```
zdrapela: Always fetch the script from `RELEASE_BRANCH` (same as current branch - main, release-1.10, etc.)

.ci/pipelines/lib/disconnected/env.sh
```diff
@@ -0,0 +1,82 @@
...
+  local rc=0
+  "$@" || rc=$?
+
+  if [[ -n "${saved_registry_auth_file}" ]]; then
+    export REGISTRY_AUTH_FILE="${saved_registry_auth_file}"
+  fi
+  return "${rc}"
+}
+
+disconnected::fetch_script() {
```
zdrapela: Ensure all scripts are located where they are supposed to be. For example this function is clearly related to Operator only.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,265 @@
...
+  # Resolve PostgreSQL image from chart values
+  local helm_values
+  helm_values=$(helm show values "${CHART_LOCAL_TGZ}" 2> /dev/null || true)
+
+  export PG_REGISTRY PG_REPO PG_TAG PG_SEPARATOR
+  PG_REGISTRY=$(echo "${helm_values}" | yq '.upstream.postgresql.image.registry' || true)
+  PG_REPO=$(echo "${helm_values}" | yq '.upstream.postgresql.image.repository' || true)
+  PG_TAG=$(echo "${helm_values}" | yq '.upstream.postgresql.image.tag' || true)
+  PG_REGISTRY="${PG_REGISTRY:-registry.redhat.io}"
+  PG_REPO="${PG_REPO:-rhel9/postgresql-15}"
```
zdrapela: Default to quay fedora. Share the default with all the other jobs that use the integrated db

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,265 @@
...
+  log::info "PostgreSQL image from chart: ${PG_REGISTRY}/${PG_REPO}${PG_SEPARATOR}${PG_TAG}"
+
+  # Resolve catalog index image (same digest encoding as PG).
+  # The init container pulls this by digest; it must be in the mirror.
+  export CI_REGISTRY CI_REPO CI_TAG CI_SEPARATOR
+  CI_REGISTRY=$(echo "${helm_values}" | yq '.global.catalogIndex.image.registry' || true)
+  CI_REPO=$(echo "${helm_values}" | yq '.global.catalogIndex.image.repository' || true)
+  CI_TAG=$(echo "${helm_values}" | yq '.global.catalogIndex.image.tag' || true)
+  CI_REGISTRY="${CI_REGISTRY:-quay.io}"
+  CI_REPO="${CI_REPO:-rhdh/plugin-catalog-index}"
```
zdrapela: Align this with https://github.com/redhat-developer/rhdh/pull/5263 usage of CATALOG_INDEX_IMAGE

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,265 @@
...
+  CI_TAG="${CI_TAG:-latest}"
+  CI_SEPARATOR=":"
+  if [[ "${CI_REPO}" == *"@"* ]]; then
```
zdrapela: Can we unify the separator across .ci? I mean the separator handling of sha vs tag.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,265 @@
...
+  log::info "Catalog index from chart: ${CI_REGISTRY}/${CI_REPO}${CI_SEPARATOR}${CI_TAG}"
+
+  # LOCAL_DISCONNECTED: prefer chart-pinned catalog digest for mirror-plugins
+  # (RELEASE_VERSION=next is not published on registry.access.redhat.com).
+  # CI keeps Gangway/env CATALOG_INDEX_IMAGE or mirror-plugins RELEASE_VERSION default.
+  if [[ "${LOCAL_DISCONNECTED:-}" == "1" && -z "${CATALOG_INDEX_IMAGE:-}" && -n "${CI_REGISTRY}" && -n "${CI_REPO}" && -n "${CI_TAG}" ]]; then
+    export CATALOG_INDEX_IMAGE="${CI_REGISTRY}/${CI_REPO}${CI_SEPARATOR}${CI_TAG}"
+    export CATALOG_INDEX_REGISTRY="${CI_REGISTRY}"
+    # Chart digest encoding: repository "…@sha256" + tag "<hash>".
+    if [[ "${CI_SEPARATOR}" == @sha256: ]]; then
```
zdrapela: Can we unify the separator across .ci? I mean the separator handling of sha vs tag.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,265 @@
...
+
+  local helm_set_flags=(
+    --set global.clusterRouterBase="${K8S_CLUSTER_ROUTER_BASE}"
+    --set upstream.postgresql.image.registry="${image_registry}"
+  )
+
+  if [[ "${LOCAL_DISCONNECTED:-}" == "1" ]]; then
+    log::info "LOCAL_DISCONNECTED: Helm hub image from chart + IDMS (not ${IMAGE_REPO}:${TAG_NAME})"
+  else
+    helm_set_flags+=(
```
zdrapela: Can we use helm::get_image_params or maybe rewrite that function to work universally?

.ci/pipelines/lib/disconnected/local.sh
```diff
@@ -0,0 +1,802 @@
...
+
+  jq -n \
+    --slurpfile base "${base_json}" \
+    --arg host "${MIRROR_REGISTRY_URL}" \
+    --arg cluster_host "${MIRROR_REGISTRY_CLUSTER_URL}" \
+    --arg auth "${auth_b64}" \
+    --arg user "${user}" \
+    --arg token "${token}" \
+    '
+    (($base[0] // {auths:{}}).auths // {}) as $auths |
```
zdrapela: Can this be a template file with envsubst?

.ci/pipelines/lib/disconnected/local.sh
zdrapela, line 1: Should this be split into a separate local lib under disconnected?

.ci/pipelines/lib/disconnected/local.sh
```diff
@@ -0,0 +1,802 @@
...
+  }
+  # Avoid wrapping our own shim if setup_local_ocp_mirror is re-entered.
+  if [[ "${real_skopeo}" == */disconnected-skopeo-shim/* ]]; then
+    return 0
+  fi
+
+  local shim_dir="${DISCONNECTED_TMPDIR}/disconnected-skopeo-shim"
+  mkdir -p "${shim_dir}"
+  cat > "${shim_dir}/skopeo" << EOF
+#!/usr/bin/env bash
```
zdrapela: Should this be a separate template file with envsubst

.ci/pipelines/lib/disconnected/plugins.sh
```diff
@@ -0,0 +1,323 @@
...
+# :1.10) or tracks :next, and survives any future rename.
+#
+# Always emits registry.access.redhat.com/rhdh/... (or the local ImageStream
+# tag form — see _hook_homepage_package_ref) so registries.conf/the
+# integrated registry serves the mirrored copy.
+# Must run before disconnected::resolve_catalog_index_image, which
+# overwrites CATALOG_INDEX_IMAGE with a canonical (mirror-consumption) form
+# that is not necessarily pullable directly from its origin registry.
+# Exports HOMEPAGE_PLUGIN_PACKAGE and HOMEPAGE_PLUGIN_FRONTEND_ID.
+disconnected::resolve_homepage_plugin_package() {
```
zdrapela: this shouldn't be required. It can be hardcoded to the name that's currently used. This function is redundant.

.ci/pipelines/lib/disconnected.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
```
zdrapela, line 1: Is this lib aligned with how sourcing and exporting is used across this repository? Should we adjust?

.ci/pipelines/resources/disconnected/app-config-rhdh-disconnected-smoke.yaml
```diff
@@ -0,0 +1,12 @@
+# Minimal app-config for disconnected operator smoke (guest login only).
```
zdrapela, line 1: move to resources/disconnected.

zdrapela:
Just rebased to fix Cluster-free check failures.

.ci/pipelines/lib/disconnected.sh
```diff
@@ -0,0 +1,95 @@
...
+readonly DISCONNECTED_LIB_SOURCED=1
+
+# shellcheck source=.ci/pipelines/lib/log.sh
+source "${DIR}/lib/log.sh"
+# shellcheck source=.ci/pipelines/lib/common.sh
+source "${DIR}/lib/common.sh"
+
+# Shared temp directory for disconnected CI artifacts.
+DISCONNECTED_TMPDIR=$(mktemp -d)
+export DISCONNECTED_TMPDIR
```
Fortune-Ndlovu, line 22: **[HIGH] No cleanup of `DISCONNECTED_TMPDIR`  credentials left on disk**

This `mktemp -d` is never cleaned up. The main CI `cleanup.sh` (sourced by `openshift-ci-tests.sh` line 39) only removes `~/tmpbin`, not this temp dir. It contains pull secrets (`auth.json`) and CA files.

In CI pods this is ephemeral, but for `LOCAL_DISCONNECTED` developer runs, sensitive credentials persist in `/tmp/tmp.XXXXXX` until reboot.

**Suggestion:** Add cleanup to `cleanup.sh`:
```bash
[[ -n "${DISCONNECTED_TMPDIR:-}" ]] && rm -rf "${DISCONNECTED_TMPDIR}"
```
Or register a trap here:
```bash
trap 'rm -rf "${DISCONNECTED_TMPDIR}"' EXIT
```

.ci/pipelines/jobs/ocp-disconnected-operator.sh
```diff
@@ -0,0 +1,252 @@
...
+  disconnected::create_mirror_registry_ca_configmap "${NAME_SPACE}" || return 1
+  disconnected::create_plugin_registry_auth_secret "${NAME_SPACE}" || return 1
+  disconnected::create_homepage_plugins_configmap "${NAME_SPACE}" || return 1
+
+  # Operator mounts one volume per extraFiles ConfigMap name. policy.json cannot
+  # share rhdh-plugin-mirror-conf or reconcile fails with duplicate volume keys.
+  oc create configmap rhdh-plugin-mirror-policy \
+    --from-literal='policy.json={"default":[{"type":"insecureAcceptAnything"}]}' \
+    -n "${NAME_SPACE}" \
+    --dry-run=client -o yaml | oc apply -f - || {
```
Fortune-Ndlovu, line 218: **[HIGH] Duplicate `policy.json` definition, two sources of truth**

This inline `insecureAcceptAnything` policy is identical to the one in `resources/disconnected/plugin-mirror-configmap.yaml:18-19`. Two separate ConfigMaps now contain the same policy:
- `rhdh-plugin-mirror-conf` (from the shared template, applied by `apply_plugin_mirror_configmap`)
- `rhdh-plugin-mirror-policy` (this inline creation)

If the policy ever needs tightening, both must be updated in sync. Consider extracting to a shared template file, or at minimum add a cross-reference comment here and in the template:
```bash
# Keep in sync with resources/disconnected/plugin-mirror-configmap.yaml policy.json
```

.ci/pipelines/lib/disconnected/mirror.sh
```diff
@@ -0,0 +1,268 @@
...
+#   - registry.redhat.io (GA): uses helm.local with chart pulled from charts.openshift.io
+#   - anything else (CI/upstream): uses helm.local with chart pulled from OCI
+# Args:
+#   $1 - output_path: Path to write the ImageSetConfiguration YAML
+disconnected::build_imageset_config() {
+  local output_path=$1
+
+  cat > "${output_path}" << EOF
+kind: ImageSetConfiguration
+apiVersion: mirror.openshift.io/v2alpha1
```
Fortune-Ndlovu: **[MEDIUM] `v2alpha1` ImageSetConfiguration API version is hardcoded**

As oc-mirror v2 exits alpha, this API version will change. The `--v2` flag on line 162 is also hardcoded. Not urgent (v2alpha1 is current for OCP 4.18+), but worth adding a comment noting the coupling so it's caught during OCP version upgrades:
```bash
# NOTE: apiVersion tied to oc-mirror v2. Update both
# when oc-mirror graduates from v2alpha1.
```

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,242 @@
...
+    log::info "LOCAL_DISCONNECTED: Helm hub image from chart + IDMS (not ${IMAGE_REPO}:${TAG_NAME})"
+    image_param_opts+=(--omit-backstage-image)
+  fi
+
+  local image_params
+  image_params=$(helm::get_image_params "${image_param_opts[@]}") || return 1
+  # get_image_params returns a space-separated --set string; split into the array
+  # (values never contain spaces).
+  local -a image_param_flags
+  read -r -a image_param_flags <<< "${image_params}"
```
Fortune-Ndlovu: **[MEDIUM] Word-splitting for `image_params` is fragile**

`read -r -a image_param_flags <<< "${image_params}"` relies on values never containing spaces. The comment on line 199 acknowledges this ("values never contain spaces"), which is true today for registry URLs and image refs.

Consider having `helm::get_image_params` return an array via nameref instead of a space-separated string, which would eliminate this class of bug entirely. Not a blocker — just a hardening suggestion for a future pass.

.ci/pipelines/lib/disconnected/local/registry.sh
```diff
@@ -0,0 +1,227 @@
...
+  local pull_secret_path="${mirror_dir}/pull-secret.json"
+  local base_json="${mirror_dir}/base-pull-secret.json"
+  if ! oc get secret pull-secret -n openshift-config \
+    -o jsonpath='{.data.\.dockerconfigjson}' | base64 -d > "${base_json}" 2> /dev/null \
+    || [[ ! -s "${base_json}" ]]; then
+    echo '{"auths":{}}' > "${base_json}"
+  fi
+
+  local token user auth_b64 whoami_user
+  token=$(oc whoami -t) || {
```
Fortune-Ndlovu: **[MEDIUM] Token-based pull secret has no expiration handling**

`oc whoami -t` returns a short-lived token (typically 24h by default on OCP). The pull secret built from it is used for the entire LOCAL_DISCONNECTED run. For long-running sessions or if someone re-uses the generated credentials, registry auth will silently expire.

Not a blocker for CI, but worth adding a comment documenting this limitation for local developers:
```bash
# NOTE: Token from 'oc whoami -t' is typically valid for 24h. For long-running
# LOCAL_DISCONNECTED sessions, re-run setup_local_ocp_mirror to refresh.
```

.ci/pipelines/resources/disconnected/helm-post-renderer.sh
```diff
@@ -0,0 +1,53 @@
...
+#   - the install-dynamic-plugins initContainer's volumeMounts[] += patch .volumeMounts
+#
+# Note: mikefarah/yq's `==` treats "*" as a glob wildcard for strings (unlike
+# jq's strict equality), so "*-developer-hub" intentionally matches the
+# rhdh-chart's computed fullname (<release>-developer-hub).
+# The initContainer is selected by name rather than index [0] so a future
+# chart reorder (e.g. an additional initContainer inserted ahead of it)
+# cannot silently misdirect the mount injection.
+PATCH_YAML="${PATCH_YAML}" yq eval '
+  (select(.kind == "Deployment" and .metadata.name == "*-developer-hub") |
```
Fortune-Ndlovu: **[LOW] yq glob matching for Deployment name**

The `"*-developer-hub"` pattern relies on mikefarah/yq's non-standard glob matching for `==` (as noted in the comment on lines 41-43). Good that this is documented. Just noting: if yq ever changes this behavior, or if the chart adds a second Deployment matching this glob, this would silently misbehave. Low risk given the chart's stable naming.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
```diff
@@ -0,0 +1,242 @@
...
+
+  if [[ -n "${HTTPS_PROXY:-}" ]]; then
+    log::info "HTTPS_PROXY is set (Playwright will use it): ${HTTPS_PROXY%%@*}@***"
+  fi
+
+  local url="https://${RELEASE_NAME}-developer-hub-${NAME_SPACE}.${K8S_CLUSTER_ROUTER_BASE}"
+  # 40 x 30s = 20 min: covers the worst-case AWS EBS CSI volume re-attach
+  # delay (~5 min, see disconnected::ensure_helm_hub_after_postgres) plus
+  # normal disconnected plugin-init time (default window is 15 min).
+  testing::check_and_test "${RELEASE_NAME}" "${NAME_SPACE}" "${PW_PROJECT_SMOKE_TEST}" "${url}" 40 30
```
Fortune-Ndlovu, line 237: **[Discussion] Smoke test only follow-up coverage worth planning**

The disconnected CI runs a smoke test only (guest login → homepage). This is the right starting point for validating the infrastructure, is there a plan for expanding disconnected test coverage as a follow-up:

- **Dynamic plugin functionality** beyond the homepage plugin
- **RBAC** in disconnected mode (does policy evaluation work without outbound access?)
- **Catalog entity resolution** from mirrored registries (are catalog entities properly served?)
- **Upgrade scenarios** in disconnected mode (can a disconnected cluster upgrade RHDH?)
- **Multi-arch** (the amd64 shim is there, but is it exercised in CI?)

Not a blocker for this PR, just noting for awareness:)

Fortune-Ndlovu (commented):
Thanks for pushing this PR, looks good to me. I've only added a few comments worth considering; I will approve after. Thanks!

.ci/pipelines/jobs/ocp-disconnected-helm.sh
zdrapela: The follow-ups are covered in https://redhat.atlassian.net/browse/RHIDP-13967, but thank you for noting it :)

.ci/pipelines/jobs/ocp-disconnected-operator.sh
zdrapela: Good catch, applied

.ci/pipelines/lib/disconnected.sh
zdrapela: Cleanup is in `.ci/pipelines/cleanup.sh` (`rm -rf "${DISCONNECTED_TMPDIR}"`). `openshift-ci-tests.sh` already sources that file and `trap cleanup EXIT`. `local-run.sh` uses the same entrypoint inside the e2e-runner, so local disconnected runs are covered. A second `EXIT` trap in `disconnected.sh` would replace `cleanup()` and skip the rest of CI teardown.

.ci/pipelines/jobs/ocp-disconnected-operator.sh
zdrapela: Pushed. Content lives in `resources/disconnected/policy.json`. `rhdh-plugin-mirror-policy` stays as a second ConfigMap because Operator extraFiles names one volume per ConfigMap.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
zdrapela: Extracted to `common::normalize_chart_image_ref` in `lib/common.sh`; disconnected Helm is the only chart field that encodes digest as `repository@sha256`.

.ci/pipelines/jobs/ocp-disconnected-helm.sh
zdrapela: Same helper as the other separator comment: `common::normalize_chart_image_ref`.

zdrapela:
Qodo follow-up on the latest review items:

**Implemented**
- Catalog index extraction now uses `tar -xf` (GNU tar auto-detects gzip; `tar -xzf` failed on uncompressed layers).
- `disconnected::resolve_homepage_plugin_package` no longer installs a sourced `RETURN` trap. Temp dir cleanup is `rm -rf` on the explicit return path (same class of bug as #4679).
- Proxy comment in `e2e-tests/playwright/utils/proxy.ts` no longer uses a credential-shaped URL example.

**Not changing — extraFiles mountPath**
Per-item `mountPath` in rhdh-operator v1alpha5 is the **directory**. `pkg/model/deployment.go` `mountPath()` + `mountFilesFrom()` join `mountPath` and `key` when a key is set (`filepath.Join(mountPath, file)`). Current values already land at `/etc/containers/policy.json`, `/etc/containers/registries.conf.d/rhdh-registries.conf`, etc. Using the full file path as `mountPath` would produce `/etc/containers/policy.json/policy.json`. Operator disconnected CI already passed with the directory form.

**Not changing — GA chart name**
`helm pull openshift-helm-charts/redhat-developer-hub` is the charts.openshift.io published name. Upstream Chart.yaml `name: backstage` is the git-repo chart name, not the marketplace chart. Helm disconnected CI already pulled this chart successfully.

**Not changing — post-renderer selector**
mikefarah/yq treats `*` as a glob in string `==`, which is documented in `helm-post-renderer.sh`. The selector is `*-developer-hub` on purpose. Helm disconnected CI already injected the disconnected volumes.

zdrapela:
Significant deltas vs the current PR description (behavior that landed after that text was written):

**Smoke plugin set.** We no longer enumerate the catalog with `--plugin-index`. Smoke mirrors only the homepage plugin (`--plugin-list`) plus the catalog index image. Full-catalog mirroring stays RHIDP-13967.

**Homepage package.** `includes: []` so unpublished `:tag` refs from `dynamic-plugins.default.yaml` are not pulled. `ref://` cannot resolve with an empty include set, so Helm/Operator `envsubst` a digest-pinned OCI package (`HOMEPAGE_PLUGIN_PACKAGE`, `oci://…@sha256:…!red-hat-developer-hub-backstage-plugin-homepage`) instead of catalog `ref://`. The description’s “re-pin homepage/catalog-index digests” / `ref://` framing is stale.

**Operator `--filter-versions`.** On `next`, both CI and local pin to chart major.minor from `CHART_VERSION` (e.g. `2.0-80-CI` → `2.0`), not `*`. Mirroring every historical operator version (~43 images) is unnecessary for smoke and aborted CI on quay CDN EOF for old 1.7/1.8 blobs. Release branches still pass their concrete version (e.g. `1.9`).

**Local-only (integrated registry).** Catalog index is tagged `:sha256-<digest>` because `:next` alone is `manifest unknown`. ImageStream tagging retries on registry Recreate EOF/503. Catalog extract uses `linux/amd64` skopeo overrides.

The Helm/Operator “passed” Prow links in the description are from earlier SHAs; later operator failures were the homepage CrashLoop (fixed by digest-pinned OCI) and the `*` catalog mirror (fixed by the filter pin).

Fortune-Ndlovu (approved):
lgtm

Fortune-Ndlovu (approved):
lgtm thanks!

PR #5271, opened 2026-08-18
docs(testing-matrix): the 209 are tests that exist — 8 of 10 lanes green, 1 persistently red

gustavolira:
Follow-up to #5270, which merged before these two commits landed. Docs-only, +15/-2.

#5270 left two things in the merged text that I had already corrected on the branch:

**1. It asserts 209 cluster-free tests as if they were available coverage.** They are tests that exist. Swept 2026-08-18 over the last 300 `rhdh-plugins` CI runs:

| | |
|---|---|
| **8 of 10 green** | `theme` 1m28s · `global-header` 1m34s · `homepage` 2m24s · `quickstart` 2m52s · `extensions` 4m44s · `bulk-import` 4m47s · `scorecard` 13m30s · `intelligent-assistant` 25m35s |
| **`orchestrator` — persistently red** | three consecutive runs failed after **195m, 216m and 191m** against the job timeout. ~3.5h of runner time per run, and it never reports a real result |
| **`adoption-insights` — no signal** | no PR touched the workspace in the window |

That last row is a property of the gate rather than the lane, and it qualifies every number in the table: the CI matrix runs only workspaces a PR changed, so **"green" means green the last time that workspace changed, not green today**. Anyone planning to move assertions into these lanes should know that before treating them as a standing safety net.

`orchestrator` is worth calling out separately — it is not an outlier, and a step that burns three and a half hours to reach a timeout is a cost in its own right, independent of this epic.

**2. The column said "Layer 4a upstream".** In RHDH "upstream" means `backstage/community-plugins`, and 11 of these 24 workspaces come from `rhdh-plugins`, which is not upstream of anything. Renamed to "in plugin repo" — the same objection was raised on the Jira comment this table came from, and fixed there too.

No threshold, support-level requirement or enforcement rule changes.

### DIFF
--- a/docs/testing-requirements-matrix.md
+++ b/docs/testing-requirements-matrix.md
@@ -120,7 +120,7 @@ Recorded because the rule above — *"where the tables say E2E, Layer 4a is the 
 as aspirational, and it is not. Measured 2026-08-18 for the 10 `rhdh-plugins` workspaces that
 also carry an e2e suite in `rhdh-plugin-export-overlays`:
 
-| Workspace | Layer 4a upstream (no cluster) | Layer 4b in overlays (cluster) | Identical test names |
+| Workspace | Layer 4a in plugin repo (no cluster) | Layer 4b in overlays (cluster) | Identical test names |
 |---|---|---|---|
 | `intelligent-assistant` | 92 | 34 | 7 |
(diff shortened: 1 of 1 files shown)


hopehadfield (approved):
/lgtm

PR #5283, opened 2026-08-20
fix(lint): shellcheck excessive memory usage on .ci pipeline scripts

zdrapela:
## Description

ShellCheck 0.11.0 was using excessive memory (10GB+, with lingering processes) when linting `.ci/pipelines/**` scripts, because its extended dataflow analysis re-parses the entire cross-sourced dependency graph for every file checked (`utils.sh` sources 8 libs; each `jobs/*.sh` sources `utils.sh` plus several libs directly).

Fix:

1. **`.shellcheckrc`**: sets `extended-analysis=false`, which disables the expensive dataflow pass while keeping source-following and function/variable resolution intact. Cuts peak memory ~9x (2.3 GB → ~260 MB on the worst file).
2. **Source guards**: extends the existing re-sourcing guard pattern (already used in `.ci/pipelines/lib/*.sh`) to the remaining sourced `.sh` files, so redundant runtime re-sourcing is a no-op. Guard variable names reflect their containing directory (e.g. `_LIB_`, `_JOBS_`, `_AKS_`).

Jira: https://redhat.atlassian.net/browse/RHDHBUGS-3685

- [RHDHBUGS-3685](https://redhat.atlassian.net/browse/RHDHBUGS-3685)

## PR acceptance criteria

- [ ] Unit Tests are updated and passing (N/A)
- [ ] E2E Tests are updated and passing (N/A — no runtime behavior change)
- [ ] Documentation is updated if necessary (N/A)
- [ ] Add a screenshot if the change is UX/UI related (N/A)

- `cd .ci && yarn shellcheck` and `cd e2e-tests && yarn shellcheck` both pass clean.
- Memory: `shellcheck -x --severity=warning .ci/pipelines/jobs/aks-operator.sh` went from 2311 MB → 268 MB peak RSS.

### DIFF
--- a/.ci/pipelines/env_variables.sh
+++ b/.ci/pipelines/env_variables.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 # shellcheck disable=SC2034
-set -a # Automatically export all variables
-
 if [[ -n "${ENV_VARIABLES_SOURCED:-}" ]]; then
   return 0
 fi
 readonly ENV_VARIABLES_SOURCED=1
+
+set -a # Automatically export all variables
 
 # Define log file names and directories.
 LOGFILE="test-log"
--- a/.ci/pipelines/env_variables.sh
+++ b/.ci/pipelines/env_variables.sh
@@ -2,6 +2,11 @@
 # shellcheck disable=SC2034
 set -a # Automatically export all variables
 
+if [[ -n "${ENV_VARIABLES_SOURCED:-}" ]]; then
+  return 0
+fi
+readonly ENV_VARIABLES_SOURCED=1
+
 # Define log file names and directories.
 LOGFILE="test-log"
 
--- a/.ci/pipelines/playwright-projects.sh
+++ b/.ci/pipelines/playwright-projects.sh
@@ -9,11 +9,6 @@
 #   source "${DIR}/playwright-projects.sh"
 #
 
-if [[ -n "${PLAYWRIGHT_PROJECTS_SOURCED:-}" ]]; then
-  return 0
-fi
-readonly PLAYWRIGHT_PROJECTS_SOURCED=1
-
 # Source logging library
 # shellcheck source=.ci/pipelines/lib/log.sh
 source "${DIR}/lib/log.sh"
--- a/.ci/pipelines/playwright-projects.sh
+++ b/.ci/pipelines/playwright-projects.sh
@@ -9,6 +9,11 @@
 #   source "${DIR}/playwright-projects.sh"
 #
 
+if [[ -n "${PLAYWRIGHT_PROJECTS_SOURCED:-}" ]]; then
+  return 0
+fi
+readonly PLAYWRIGHT_PROJECTS_SOURCED=1
+
 # Source logging library
 # shellcheck source=.ci/pipelines/lib/log.sh
 source "${DIR}/lib/log.sh"
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,16 @@
+# ShellCheck configuration for this repository.
+#
+# The .ci/ and e2e-tests/ pipeline scripts form a heavily cross-sourced
+# dependency graph (see .ci/pipelines/utils.sh and .ci/pipelines/jobs/*.sh).
+# ShellCheck's extended dataflow analysis re-parses the fully resolved
+# transitive `source` graph for every file it checks, which can balloon
+# memory usage into the multiple-gigabyte range for a single job script
+# (e.g. .ci/pipelines/jobs/aks-operator.sh with -x can exceed 2 GB RSS).
+# This is especially costly with IDE extensions that check on every save
+# and follow `# shellcheck source=` directives by default.
+#
+# Disabling extended-analysis keeps source-following, function/variable
+# resolution, and the vast majority of checks intact, while cutting
+# resource usage roughly 9x on the worst offenders. See:
+# https://github.com/koalaman/shellcheck/blob/master/shellcheck.1.md#rc-files
+extended-analysis=false
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,16 +1,9 @@
-# ShellCheck configuration for this repository.
+# Repo-wide ShellCheck defaults.
 #
-# The .ci/ and e2e-tests/ pipeline scripts form a heavily cross-sourced
-# dependency graph (see .ci/pipelines/utils.sh and .ci/pipelines/jobs/*.sh).
-# ShellCheck's extended dataflow analysis re-parses the fully resolved
-# transitive `source` graph for every file it checks, which can balloon
-# memory usage into the multiple-gigabyte range for a single job script
-# (e.g. .ci/pipelines/jobs/aks-operator.sh with -x can exceed 2 GB RSS).
-# This is especially costly with IDE extensions that check on every save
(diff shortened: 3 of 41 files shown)


.ci/pipelines/env_variables.sh
```diff
@@ -2,6 +2,11 @@
 # shellcheck disable=SC2034
 set -a # Automatically export all variables
 
+if [[ -n "${ENV_VARIABLES_SOURCED:-}" ]]; then
```
gustavolira, line 3: This guard sits below `set -a` (line 3), and this file's matching `set +a` is the last line (346). That combination bites twice — both verified on bash 5.3:

**1. The guard variable gets exported.** `readonly ENV_VARIABLES_SOURCED=1` runs under `allexport`, so it lands as `declare -rx` — it's the only one of the ~25 guards in this PR that escapes the shell. A child process that inherits it and sources this file returns at line 6 and defines nothing, silently, with status 0:

```
$ env -i PATH=/bin:/usr/bin ENV_VARIABLES_SOURCED=1 bash -c 'source env_variables.sh; echo ${SOME_SECRET:-UNSET}'
UNSET
$ env -i PATH=/bin:/usr/bin bash -c 'source env_variables.sh; echo ${SOME_SECRET:-UNSET}'
s3cr3t
```

Nothing in the tree does that today — `openshift-ci-tests.sh:27` is the only source site and `container-init.sh` doesn't pre-source it — but it's exactly the trap a local-run wrapper would fall into.

**2. Re-sourcing leaves `allexport` on.** The shell runs `set -a` at line 3, hits the guard, and returns, so line 346's `set +a` never executes. Every later top-level assignment in `utils.sh`, `lib/*.sh` and the job handlers then gets exported into the environment of `helm`, `oc` and `yarn playwright`. That's worse than the pre-PR behaviour, where a redundant re-source was wasteful but ended with `allexport` correctly off.

Moving the guard block above `set -a` fixes both at once.

e2e-tests/local-test-setup.sh
```diff
@@ -21,6 +21,11 @@
 #   yarn install
 #   yarn playwright test --headed
 
+if [[ -n "${LOCAL_TEST_SETUP_SOURCED:-}" ]]; then
```
gustavolira: Qodo already flagged this one, but I want to add the evidence for it, because I think it's worth blocking on rather than filing away.

This script is argument-driven and side-effecting, so it's not idempotent by design: the arg loop at 57-63 parses `showcase|rbac|--env`, lines 68/72 set `BASE_URL` accordingly, lines 115/123 mint a fresh 8-hour `K8S_CLUSTER_TOKEN`, and line 151 regenerates `e2e-tests/.env`.

Re-sourcing it in the same shell isn't an accident — it's what the repo tells people to do:

- `local-run.sh:504-505` prints both `source local-test-setup.sh` and `source local-test-setup.sh rbac`
- `e2e-tests/README.md:180-181` — same pair
- `.claude/skills/e2e-reproduce-failure/SKILL.md:28,31` — `showcase` then `rbac`
- `.claude/skills/e2e-verify-fix/SKILL.md:37` — "To regenerate (e.g. after token expiry), re-run the same command."
- the generated `.env` header itself, line 171: "Regenerate by running: `source local-test-setup.sh <showcase|rbac> --env`"

With the guard, every one of those second invocations returns 0 and prints nothing. RBAC tests then run against the non-RBAC instance, an expired token is never refreshed, and `--env` writes no file.

There's no recovery inside the shell either — `LOCAL_TEST_SETUP_SOURCED` is `readonly`, so `unset` fails and the user has to open a new terminal.

And the guard buys nothing for this PR's goal: this file sources only `lib/log.sh`, so it isn't part of the cross-sourced graph that makes ShellCheck expensive. Dropping it from the change looks like the cleanest fix.

.ci/pipelines/playwright-projects.sh
```diff
@@ -9,6 +9,11 @@
 #   source "${DIR}/playwright-projects.sh"
 #
 
+if [[ -n "${PLAYWRIGHT_PROJECTS_SOURCED:-}" ]]; then
```
gustavolira: The guard latches at line 15, but the file can still fail at line 27 when `projects.json` is missing. If the first source fails, the marker is already set, so any later source returns 0 without exporting a single `PW_PROJECT_*` — a hard failure quietly becomes a silent one.

`set -o errexit` in `openshift-ci-tests.sh` means the first `return 1` aborts the run, so this isn't reachable in CI. It is reachable in a local shell, or anywhere the source gets wrapped in `|| true`.

Same shape as the `local-test-setup.sh` issue: the marker is set before the file has actually done its work. Moving lines 12-15 below the `projects.json` check keeps the guard and drops the trap.

.shellcheckrc
```diff
@@ -0,0 +1,16 @@
...
+# memory usage into the multiple-gigabyte range for a single job script
+# (e.g. .ci/pipelines/jobs/aks-operator.sh with -x can exceed 2 GB RSS).
+# This is especially costly with IDE extensions that check on every save
+# and follow `# shellcheck source=` directives by default.
+#
+# Disabling extended-analysis keeps source-following, function/variable
+# resolution, and the vast majority of checks intact, while cutting
+# resource usage roughly 9x on the worst offenders. See:
+# https://github.com/koalaman/shellcheck/blob/master/shellcheck.1.md#rc-files
+extended-analysis=false
```
gustavolira, line 9: Two notes on this living at the repo root.

**It falls outside both lint gates' path filters.** `bash-e2e-lint.yaml` triggers on `.ci/**` and `e2e-tests-lint.yaml` on `e2e-tests/**`, so a future PR that changes only `.shellcheckrc` runs neither ShellCheck job — the file configuring both linters isn't covered by either. This PR is covered incidentally because it also touches `.ci/**`. Worth adding `.shellcheckrc` to both `paths:` lists.

**The comment frames the scope narrower than it is.** It describes `.ci/` and `e2e-tests/`, but the setting applies repo-wide — `scripts/*.sh` (7 files) and `.ci/manual-tests/postgres/install.sh` are affected too, and none of them are in the cross-sourced graph. Root is still the right location, since `local-test-setup.sh:36` sources `../.ci/pipelines/lib/log.sh` and a `.ci/`-scoped file would split the graph. Just saying "repo-wide default" would stop the next reader assuming `scripts/` is untouched.

For what it's worth, I measured the tradeoff and it's narrower than the comment suggests: `extended-analysis=false` only drops SC2317 and SC2329 — SC2034, SC2154, SC2030 and SC2031 all still fire. On `jobs/aks-operator.sh -x` I got 2421 MB / 58.9s down to 283 MB / 3.5s. Worth noting the source guards contribute almost none of that (2421 to 2170 MB with extended analysis still on), so the `.shellcheckrc` is doing essentially all the work here.

e2e-tests/local-test-setup.sh
zdrapela: Ah, that's a very good point. Thank you for noting this! I'll fix it.

.ci/pipelines/playwright-projects.sh
zdrapela: Agreed

.ci/pipelines/env_variables.sh
zdrapela: Good catch, thanks!

gustavolira (approved):

PR #5291, opened 2026-08-21
docs(testing-matrix): correct the Layer 4a counts and say what they count

gustavolira:
The Layer 4a table added on 2026-08-18 carried two numbers its source document disagreed with, and never said what it was counting.

## What was wrong

**Two cells contradict the source.** `scorecard` is **16** overlay tests and `orchestrator` **26**, not 15 and 24 — so the total is **138**, not 135. The companion analysis in `rhdh-plugin-export-overlays` had the correct figures in its per-workspace table all along; only the summary table here was off.

**Neither figure is a run count.** `npx playwright test --list` reports **19** for `scorecard` and **32** for `orchestrator`, because parametrised tests expand at collection time:

```
$ cd workspaces/scorecard/e2e-tests && npx playwright test --list
Total: 19 tests in 2 files
```

Both columns count statically declared `test()` blocks. That makes the two sides comparable to each other, but makes them a floor rather than a run count — which now appears under the table instead of being left for a reader to discover.

**"the 10 `rhdh-plugins` workspaces that also carry an e2e suite" is 11.** The eleventh is `app-defaults`, and it is the one with no upstream lane — which is precisely why it is absent from the table. Naming it stops a reader concluding it already has the coverage the table describes.

## Scope

Docs only, one file. No layer definitions, thresholds, or requirements change.

Found while reviewing redhat-developer/rhdh-plugin-export-overlays#3291, which carries the same table and is corrected the same way in that PR.

### DIFF
--- a/docs/testing-requirements-matrix.md
+++ b/docs/testing-requirements-matrix.md
@@ -117,24 +117,30 @@ apiRef{...}`.
 ## Layer 4a already exists, and overlaps Layer 4b
 
 Recorded because the rule above — *"where the tables say E2E, Layer 4a is the default"* — reads
-as aspirational, and it is not. Measured 2026-08-18 for the 10 `rhdh-plugins` workspaces that
-also carry an e2e suite in `rhdh-plugin-export-overlays`:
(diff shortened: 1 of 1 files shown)


hopehadfield (approved):
/lgtm

PR #5341, opened 2026-09-03
fix(ci): pin the chart to the image tag and sort chart tags explicitly

gustavolira:
## Problem

Two issues in how `CHART_VERSION` is resolved. Both silently produce a run that does not test what the operator asked for, which matters most for RC and GA verification triggered through Gangway.

### A pinned image tag does not pin the chart

`--tag 1.10-170` alone leaves `CHART_VERSION` empty, so `helm::get_chart_version` resolves the newest chart on the branch:

```bash
if [[ -z "${CHART_VERSION:-}" ]]; then
  CHART_VERSION=$(helm::get_chart_version)   # newest, regardless of TAG_NAME
fi
```

An RC verification with the `1.10-170` image ran against chart `1.10-170-CI` on Sep 1 and against `1.10-171-CI` from Sep 3 onwards, with nothing in the log saying the pair had drifted. The image and the chart are published together under the same build number — `1.10-170` and `1.10-170-CI` were pushed three minutes apart — so a pinned `TAG_NAME` already determines the chart.

```bash
| jq -r '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'
```

## Fix

Derive `CHART_VERSION` from `TAG_NAME` when the tag is pinned to a build number, and log that it was derived. Select the CI chart tags explicitly and take `max_by(.start_ts)`.

Behaviour preserved:

| Input | Before | After |
|---|---|---|
| `--chart-version 1.10-170-CI` | used as given | unchanged |
| `--tag 1.10-170`, no chart | newest chart | `1.10-170-CI` |
| `--tag 1.10` or `next` | newest chart | unchanged |

## Verification

```
1.10-170 -> 1.10-170-CI
2.0-84   -> 2.0-84-CI
1.10     -> resolves from quay (not pinned)
next     -> resolves from quay (not pinned)
1.10.4   -> resolves from quay (not pinned)
```

The new jq returns `1.10-171-CI` against the live API, matching the previous behaviour on today's data while no longer relying on response order. `shellcheck --severity=warning` and `yarn prettier:check` pass.

## Other branches

`release-1.10` carries this code verbatim and needs the same fix — happy to open the backport, or it can be cherry-picked.

`release-1.9` has a different and worse variant: no `-z` guard at all, so `CHART_VERSION` is assigned unconditionally and an explicit `--chart-version` is silently discarded. Out of scope here, but worth a separate look if 1.9 still gets RC runs.

### DIFF
--- a/.ci/pipelines/lib/helm.sh
+++ b/.ci/pipelines/lib/helm.sh
@@ -177,7 +177,7 @@ helm::get_chart_version() {
   local version
   version=$(curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" \
     -H "Content-Type: application/json" \
-    | jq -r '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI') || {
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name') || {
     log::error "Failed to resolve chart version for ${chart_major_version}"
     return 1
   }
--- a/.ci/pipelines/lib/helm.sh
+++ b/.ci/pipelines/lib/helm.sh
@@ -137,8 +137,7 @@ helm::_highest_published_chart_major() {
     -H "Content-Type: application/json" \
     | jq -r '.tags[].name' \
     | grep -oE '^[0-9]+\.[0-9]+' \
-    | sort -t. -k1,1n -k2,2n \
-    | uniq | tail -1
+    | sort -uV | tail -1
 }
 
 # Get the chart major.minor version based on RELEASE_BRANCH_NAME or an optional override.
--- a/.ci/pipelines/lib/helm.sh
+++ b/.ci/pipelines/lib/helm.sh
@@ -124,10 +124,8 @@ helm::get_previous_release_values() {
 # Chart Operations
 # ==============================================================================
 
-# Fetch a page of chart tags from Quay.
-# Piping curl straight into jq loses curl's status, which makes a 5xx or a
-# timeout indistinguishable from an empty result. Capture the body first so a
-# transport failure is reported rather than read as "no tags published".
+# Fetch a page of chart tags from Quay. Captures the body first: piping curl
+# into jq loses curl's status, so a 5xx would read as "no tags published".
 # Args:
 #   $1 - query: extra query string, e.g. "filter_tag_name=like:1.10-"
 helm::_fetch_chart_tags() {
@@ -154,9 +152,8 @@ helm::_latest_chart_tag() {
     | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name // empty'
 }
 
-# Print the highest major.minor stream that has any chart published.
-# Reads only the first page, which Quay returns newest-first, so the newest
-# stream is present even though the listing is not exhaustive.
+# Print the highest stream with any chart published. Reads the first page only,
+# which Quay returns newest-first.
 helm::_highest_published_chart_stream() {
   helm::_fetch_chart_tags "limit=100" \
     | jq -r '.tags[].name' \
@@ -164,13 +161,9 @@ helm::_highest_published_chart_stream() {
     | sort -uV | tail -1
 }
 
-# Resolve the chart version for this run, in precedence order:
-#   1. an explicit CHART_VERSION (--chart-version, or the environment)
-#   2. a pinned TAG_NAME, since the image and the chart are published together
-#   3. the newest chart on the stream this checkout belongs to
-#
-# Owning all three here keeps the tag/chart naming convention in one place
-# instead of spreading it between the dispatcher and this library.
+# Resolve the chart version: an explicit CHART_VERSION wins, then a pinned
+# TAG_NAME (image and chart ship together), then the newest chart on this
+# checkout's stream.
 # Uses globals: CHART_VERSION, TAG_NAME
 helm::resolve_chart_version() {
   if [[ -n "${CHART_VERSION:-}" ]]; then
@@ -179,9 +172,8 @@ helm::resolve_chart_version() {
     return 0
   fi
 
-  # An RC image is tagged x.y-N and its chart x.y-N-CI; a GA image is tagged
-  # x.y.z and its chart carries the same x.y.z. Both shapes appear in the
-  # documented verification flows, so both have to map.
+  # RC images are tagged x.y-N with chart x.y-N-CI; GA images x.y.z with chart
+  # x.y.z. Both shapes appear in the verification flows, so both have to map.
   if [[ "${TAG_NAME:-}" =~ ^[0-9]+\.[0-9]+-[0-9]+$ ]]; then
     log::info "Derived CHART_VERSION from pinned TAG_NAME: ${TAG_NAME}-CI" >&2
     echo "${TAG_NAME}-CI"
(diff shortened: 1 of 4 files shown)


.ci/pipelines/lib/helm.sh
```diff
@@ -177,7 +177,7 @@ helm::get_chart_version() {
   local version
   version=$(curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" \
```
nickboldt:
```suggestion
  version=$(curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}." \
```

nickboldt:
```
curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:2." -H "Content-Type: application/json" | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name'
2.0-84-CI
```

but

```
curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:2-" -H "Content-Type: application/json" | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name'
2.0-82-CI
```

So... you want to look for `2.*` not `2-*`

Even better would be to search for `2.0-*` -- can we get the RHDH version from package.json and use that instead of `chart_major_version` ?

nickboldt (changes requested):

.ci/pipelines/lib/helm.sh
```diff
@@ -122,6 +122,25 @@ helm::get_previous_release_values() {
...
+#   $1 - chart_major_version: e.g. "1.10"
+helm::_latest_chart_tag() {
+  curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${1}-" \
+    -H "Content-Type: application/json" \
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name // empty'
+}
+
+# Print the highest major.minor stream that has any chart published.
+helm::_highest_published_chart_major() {
+  curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&limit=100" \
```
nickboldt: simpler - use sort with unique and version-sorting, then pick the last one:

```
curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&limit=100"     -H "Content-Type: application/json"     | jq -r '.tags[].name'     | grep -oE '^[0-9]+\.[0-9]+' | sort -uV | tail -1
```

gustavolira: Done, thanks!

.ci/pipelines/lib/helm.sh
```diff
@@ -142,15 +160,15 @@ helm::get_chart_major_version() {
...
+    # pushed, so "the highest version on quay" silently follows whichever stream
+    # happens to be ahead rather than the one this checkout belongs to.
     local chart_major_version
-    chart_major_version=$(curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&limit=100" \
-      -H "Content-Type: application/json" \
-      | jq -r '.tags[].name' \
-      | grep -oE '^[0-9]+\.[0-9]+' \
-      | sort -t. -k1,1n -k2,2n \
-      | uniq | tail -1)
+    chart_major_version=$(jq -r '.version // empty' "${DIR}/../../package.json" 2> /dev/null \
```
nickboldt: this should be `chart_major_minor` or `chart_xy` because  = 2.0, not 2.

Also if there's no version set in package.json we have a serious problem and this should fail... not just return an empty string.

.ci/pipelines/lib/helm.sh
```diff
@@ -122,6 +122,24 @@ helm::get_previous_release_values() {
...
+# Args:
+#   $1 - chart_major_version: e.g. "1.10"
+helm::_latest_chart_tag() {
+  curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${1}-" \
+    -H "Content-Type: application/json" \
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name // empty'
+}
+
+# Print the highest major.minor stream that has any chart published.
+helm::_highest_published_chart_major() {
```
nickboldt: should be `highest_published_chart_major_minor` or if that's too verbose then `highest_published_chart_xy`

.ci/pipelines/lib/helm.sh
```diff
@@ -175,12 +193,27 @@ helm::get_chart_version() {
...
+  version=$(helm::_latest_chart_tag "${chart_major_version}") || {
     log::error "Failed to resolve chart version for ${chart_major_version}"
     return 1
   }
+
+  # Between a version bump and the first chart built from it, no tag exists for
+  # the version package.json declares. Fall back to the newest published stream
+  # so the run continues rather than failing on a gap that closes by itself.
+  if [[ -z "$version" || "$version" == "null" ]]; then
+    local fallback_major
```
nickboldt: this should be `fallback_xy` or `fallback_major_minor` because it's two digits ;)

.ci/pipelines/lib/helm.sh
gustavolira: Held off on this one — `chart_major_version` is `X.Y`, so the suggestion becomes `like:1.10.` rather than `like:1.10`, and that matches the GA tags instead of the CI ones:

```
like:2.0-   -> 32 tags -> 2.0-84-CI
like:2.0.   ->  0 tags -> nothing
like:1.10-  -> 50 tags -> 1.10-171-CI
like:1.10.  ->  5 tags (the 1.10.x GA charts) -> nothing
```

Your earlier `like:2.` worked because it was major-only. With the minor included the trailing dot excludes every `-CI` tag.

Your GA point landed elsewhere though — I'd missed that a GA run pins `--tag 1.10.4` with no `--chart-version`, so it was still resolving a mismatched chart. Added `x.y.z -> x.y.z` alongside `x.y-N -> x.y-N-CI`, and moved both into `helm::resolve_chart_version` so the convention lives in one place.

.ci/pipelines/lib/helm.sh
gustavolira: Both fixed in the latest push — you were on the older revision.

Went with `chart_stream` over `chart_xy`; the tags are grouped by stream, so it names the thing rather than the shape. Same for `get_chart_major_version` -> `get_chart_stream`. Can switch to `_xy` if you prefer.

And yeah, empty version fails now — separate errors for a missing file vs a missing `.version`, both return 1, so errexit kills the run.

.ci/pipelines/lib/helm.sh
gustavolira: Renamed to `helm::_highest_published_chart_stream` in the latest push. Went with `stream` over `_xy` since it says what the value is rather than how many digits it has, but happy to switch if you prefer the explicit form.

.ci/pipelines/lib/helm.sh
gustavolira: That block is gone — it is `fallback_stream` now.

The `|| { }` above it was dead code anyway: the helper ends in a pipe, so the status is jq's and a failed curl still returns 0. Branches on empty output now.

.ci/pipelines/lib/helm.sh
```diff
@@ -168,19 +240,34 @@ helm::get_chart_major_version() {
...
+  # returns 0. Empty output is the only reliable signal.
   local version
-  version=$(curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" \
-    -H "Content-Type: application/json" \
-    | jq -r '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI') || {
-    log::error "Failed to resolve chart version for ${chart_major_version}"
+  version=$(helm::_latest_chart_tag "${chart_stream}")
+
+  # Right after a version bump, package.json can name a stream with no chart
+  # built yet. Fall back so the run survives that one-build gap - but only on
```
nickboldt, line 255: ^^ nice!

nickboldt (approved):

PR #5342, opened 2026-09-03
fix(ci): pin the chart to the image tag and sort chart tags explicitly

gustavolira:
Same fix as #5341, opened directly against `release-1.10` rather than as a backport — `main` has refactored the surrounding code since, so a cherry-pick would not apply cleanly. The two hunks below are identical in both branches.

## Problem

Two issues in how `CHART_VERSION` is resolved. Both silently produce a run that does not test what the operator asked for, which matters most on this branch: the release nightlies are triggered on demand for RC and GA verification, where the entire point is to test one specific set of artifacts.

### A pinned image tag does not pin the chart

`--tag 1.10-170` alone leaves `CHART_VERSION` empty, so `helm::get_chart_version` resolves the newest chart on the branch:

```bash
if [[ -z "${CHART_VERSION:-}" ]]; then
  CHART_VERSION=$(helm::get_chart_version)   # newest, regardless of TAG_NAME
fi
```

An RC verification with the `1.10-170` image ran against chart `1.10-170-CI` on Sep 1 and against `1.10-171-CI` from Sep 3 onwards, with nothing in the log saying the pair had drifted. The image and the chart are published together under the same build number — `1.10-170` and `1.10-170-CI` were pushed three minutes apart — so a pinned `TAG_NAME` already determines the chart.

```bash
| jq -r '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'
```

## Fix

Derive `CHART_VERSION` from `TAG_NAME` when the tag is pinned to a build number, and log that it was derived. Select the CI chart tags explicitly and take `max_by(.start_ts)`.

Behaviour preserved:

| Input | Before | After |
|---|---|---|
| `--chart-version 1.10-170-CI` | used as given | unchanged |
| `--tag 1.10-170`, no chart | newest chart | `1.10-170-CI` |
| `--tag 1.10` | newest chart | unchanged |

## Verification

```
1.10-170 -> 1.10-170-CI
2.0-84   -> 2.0-84-CI
1.10     -> resolves from quay (not pinned)
1.10.4   -> resolves from quay (not pinned)
```

The new jq returns `1.10-171-CI` against the live API, matching the previous behaviour on today's data while no longer relying on response order. `shellcheck --severity=warning` and `yarn prettier:check` pass.

## Related

`release-1.9` has a different and worse variant: no `-z` guard at all, so `CHART_VERSION` is assigned unconditionally and an explicit `--chart-version` is silently discarded. Out of scope here, but worth a look if 1.9 still gets RC runs.

### DIFF
--- a/.ci/pipelines/lib/helm.sh
+++ b/.ci/pipelines/lib/helm.sh
@@ -9,6 +9,8 @@ if [[ -n "${HELM_LIB_SOURCED:-}" ]]; then
 fi
 readonly HELM_LIB_SOURCED=1
 
+readonly HELM_CHART_TAGS_API="https://quay.io/api/v1/repository/rhdh/chart/tag/"
+
 # shellcheck source=.ci/pipelines/lib/log.sh
 source "${DIR}/lib/log.sh"
 # shellcheck source=.ci/pipelines/lib/common.sh
@@ -85,7 +87,7 @@ helm::get_previous_release_values() {
   local value_file_type=${1:-"showcase"}
 
   local current_release_version
-  current_release_version=$(helm::get_chart_major_version)
+  current_release_version=$(helm::get_chart_stream)
   if [[ -z "$current_release_version" ]]; then
     return 1
   fi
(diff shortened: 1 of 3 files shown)


.ci/pipelines/lib/helm.sh
```diff
@@ -178,7 +178,7 @@ helm::get_chart_version() {
   local version
   version=$(curl -sSfX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" \
     -H "Content-Type: application/json" \
-    | jq -r '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI') || {
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name') || {
```
nickboldt: why the `\\.` instead of `\.` ? do jq regexed need the extra escape?

gustavolira: @nickboldt  jq needs it. test() takes the pattern as a jq string literal, so jq's string parser eats the backslash first. If I try it I will receive `jq: error: Invalid escape at line 1, column 4 (while parsing '"\."')`

nickboldt (changes requested):
See https://github.com/redhat-developer/rhdh/pull/5341#discussion_r3934329707

PR #5343, opened 2026-09-03
fix(ci): honour a preset CHART_VERSION and sort chart tags explicitly

gustavolira:
Companion to #5341 (`main`) and #5342 (`release-1.10`). This branch carries a different and worse variant of the same problem, so this is not a straight cherry-pick.

### `--chart-version` is silently discarded on this branch

```bash
CHART_VERSION=$(get_chart_version "$CHART_MAJOR_VERSION")
export CHART_VERSION
```

No `-z` guard. The Gangway trigger passes an override through `MULTISTAGE_PARAM_OVERRIDE_CHART_VERSION`, which reaches the script as `CHART_VERSION` and is overwritten on this line before anything reads it. Every `release-1.9` run has resolved the newest chart on the branch regardless of what the operator asked for. `main` and `release-1.10` already guard this; this branch never did.

### A pinned image tag does not pin the chart either

`--tag 1.9-246` alone leaves `CHART_VERSION` empty, so the newest chart is resolved. The image and the chart are published together under the same build number, so a pinned `TAG_NAME` already determines the chart.

```bash
| jq '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'
```

## Fix

Add the `-n` guard so a preset `CHART_VERSION` wins. Derive it from `TAG_NAME` when the tag is pinned to a build number, and log that it was derived. Select the CI chart tags explicitly and take `max_by(.start_ts)`.

| Input | Before | After |
|---|---|---|
| `--chart-version 1.9-246-CI` | **discarded**, newest chart used | used as given |
| `--tag 1.9-246`, no chart | newest chart | `1.9-246-CI` |
| `--tag 1.9` | newest chart | unchanged |

## Verification

Both the old and the new jq expression return `1.9-247-CI` against the live API, so today's behaviour is unchanged while the dependency on response order is gone.

```
1.9-246 -> 1.9-246-CI
1.9     -> resolves from quay (not pinned)
```

## Note on structure

`get_chart_version` lives in `.ci/pipelines/utils.sh` on this branch rather than `.ci/pipelines/lib/helm.sh`, and takes `$CHART_MAJOR_VERSION` as an argument. The change is equivalent to the other two PRs, adapted to that shape.

### DIFF
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -1510,7 +1510,7 @@ get_previous_release_version() {
 get_chart_version() {
   local chart_major_version=$1
   curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" -H "Content-Type: application/json" \
-    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name'
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name // empty'
 }
 
 # Helper function to get dynamic value file path based on previous release version
--- a/.ci/pipelines/utils.sh
+++ b/.ci/pipelines/utils.sh
@@ -1510,7 +1510,7 @@ get_previous_release_version() {
 get_chart_version() {
   local chart_major_version=$1
   curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" -H "Content-Type: application/json" \
-    | jq '.tags[0].name' | grep -oE '[0-9]+\.[0-9]+-[0-9]+-CI'
+    | jq -r '[.tags[] | select(.name | test("^[0-9]+\\.[0-9]+-[0-9]+-CI$"))] | max_by(.start_ts) | .name'
 }
 
 # Helper function to get dynamic value file path based on previous release version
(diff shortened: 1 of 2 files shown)


.ci/pipelines/utils.sh
```diff
@@ -1510,7 +1510,7 @@ get_previous_release_version() {
 get_chart_version() {
   local chart_major_version=$1
   curl -sSX GET "https://quay.io/api/v1/repository/rhdh/chart/tag/?onlyActiveTags=true&filter_tag_name=like:${chart_major_version}-" -H "Content-Type: application/json" \
```
nickboldt, line 1512: can you add  comment explaining that `chart_major_version` is actually in the form x.y (major.minor or stream) ?

gustavolira: Added:

```bash
# Get the newest CI chart tag for a release stream.
# Args:
#   $1 - chart_major_version: despite the name, this is a major.minor pair -
#        the release stream, e.g. "1.9" - not a bare major. CHART_MAJOR_VERSION
#        in env_variables.sh carries the same shape.
# Returns:
#   Prints the tag (e.g. "1.9-247-CI"), or nothing when the stream has none.
```

Documented rather than renamed

nickboldt: AI-generated and therefore overly verbose, but at least it's documented.

nickboldt:
failing for unrelated reason:
```
  [ERROR] Base image includes nodejs v22.23.2 but could not find releases/node-v22.23.2-headers.tar.gz to install!
  [ERROR] To fix, upload the node-v22.23.2-headers.tar.gz file into https://github.com/redhat-developer/rhdh/tree/main/.nvm/releases (or related branch)!
```

nickboldt (approved):

PR #5351, opened 2026-09-08
[release-1.10] fix(e2e): let local-run.sh pull the runner image for the host arch

gustavolira:
Backport of fixes already on `main`. Nothing new is invented here — the three hunks are `main`'s current code.

## Problem

`local-run.sh` could not run on an arm64 host from this branch. Three separate causes, each costing a debugging round:

**The runner image was pulled as amd64.** The image is multi-arch and publishes `linux/arm64`, but the pull hardcoded `--platform=linux/amd64`. Under emulation every Go binary in it crashed:

```
SIGSEGV: segmentation violation
  gopsutil/v3/cpu.init.0()                     <- vault

runtime: lfstack.push invalid packing
  fatal error: lfstack.push                    <- helm
```

**The vault binary was downloaded as amd64**, so it could not run on the arm64 variant either.

**A failed vault read was invisible.** `vault kv get | jq` takes jq's exit status, so a crashed vault yielded an empty `/tmp/secrets/`, the script logged `✓ Secrets written to /tmp/secrets/`, and the run died fifteen seconds later on `cat: /tmp/secrets/GH_USER_ID: No such file or directory`. That masking is architecture-independent.

## Fix

Taken from `main` verbatim:

- Drop `--platform`, let podman match the host, fall back to a local copy if the pull fails.
- `VAULT_ARCH=$(dpkg --print-architecture)` for the download.
- Guard `SECRETS` before use.

## Verification

Found while running the full `handle_ocp_nightly` pipeline from an Apple Silicon machine. With these three applied, the run completed end to end on OCP 4.22:

```
showcase                 35 passed
showcase-rbac            17 passed
showcase-runtime         19 passed
showcase-sanity-plugins   9 passed
```

### DIFF
--- a/e2e-tests/container-init.sh
+++ b/e2e-tests/container-init.sh
@@ -23,14 +23,21 @@ set -e
 if ! command -v vault &> /dev/null; then
   VAULT_VERSION="${VAULT_VERSION:-1.15.4}"
   log::info "Installing vault ${VAULT_VERSION}..."
-  curl -fsSL "https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip" -o /tmp/vault.zip
+  VAULT_ARCH=$(dpkg --print-architecture)
+  curl -fsSL "https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_${VAULT_ARCH}.zip" -o /tmp/vault.zip
   unzip -q /tmp/vault.zip -d /usr/local/bin/
   rm /tmp/vault.zip
 fi
 
 # Fetch and write secrets to /tmp/secrets/
 log::section "Fetching Vault Secrets"
+set -o pipefail
 SECRETS=$(vault kv get -format=json -mount="kv" "selfservice/rhdh-qe/rhdh" | jq -r ".data.data")
+set +o pipefail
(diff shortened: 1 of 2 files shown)


e2e-tests/container-init.sh
```diff
@@ -23,14 +23,21 @@ set -e
 if ! command -v vault &> /dev/null; then
   VAULT_VERSION="${VAULT_VERSION:-1.15.4}"
   log::info "Installing vault ${VAULT_VERSION}..."
-  curl -fsSL "https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip" -o /tmp/vault.zip
+  VAULT_ARCH=$(dpkg --print-architecture)
```
nickboldt: rather than depending on a script that's specific to debian, why not use this? 
```
uname -m
```

gustavolira: `uname -m` alone breaks it — the names do not match what HashiCorp publishes:

```
vault_1.15.4_linux_amd64.zip    200
vault_1.15.4_linux_x86_64.zip   404
```

That is why dpkg was there. Switched to uname with a case mapping, plus an explicit failure on anything without a build.

gustavolira: Opened #5352 for main with the same change, so the two branches stay in step.

nickboldt (approved):
/lgtm

PR #5275, opened 2026-08-18
fix(e2e): align Playwright npm and workflow image with v1.62.1

kim-tsao:
Renovate bumped the CI Dockerfile to playwright v1.62.1-noble but left e2e-tests and the cluster-free workflow on 1.61.1, causing browser version mismatch failures at test runtime.

## Description

Please explain the changes you made here.

Fixes e2e breakage caused by Renovate [PR](https://github.com/redhat-developer/rhdh/pull/5176)
## Which issue(s) does this PR fix

- Fixes #?

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

PR #5362, opened 2026-09-09
chore(renovate): group Playwright npm and Docker updates together

kim-tsao:
## Summary
- Group `@playwright/test`, `playwright`, `playwright-core`, and `mcr.microsoft.com/playwright` into a single Renovate PR so npm and Docker browser images stay in lockstep
- Exclude the Playwright Docker image from the GitHub Actions group (it was landing there and desyncing from `@playwright/test`)
- Apply the same 3-day `minimumReleaseAge` to the Playwright Docker image so it cannot race ahead of the npm age gate

Fixes https://redhat.atlassian.net/browse/RHIDP-16879

## Test plan
- [ ] Confirm Renovate config JSON is valid
- [ ] After merge, verify the next Playwright bump opens one `playwright-` PR that includes both the npm package and Docker image references (workflow + `.ci/images/Dockerfile`)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -179,9 +179,11 @@
       "automerge": false
     },
     {
-      "description": "Align Playwright Docker release age with npm minimumReleaseAge",
+      "description": "Age Playwright Docker via GitHub Releases (MCR has no releaseTimestamp)",
       "matchPackageNames": ["mcr.microsoft.com/playwright"],
       "matchDatasources": ["docker"],
+      "overrideDatasource": "github-releases",
+      "overridePackageName": "microsoft/playwright",
       "minimumReleaseAge": "3 day"
     },
     {
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -160,9 +160,29 @@
       "description": "Keep GitHub Actions updates in a separate PR",
       "matchManagers": ["github-actions"],
       "matchUpdateTypes": ["patch", "minor"],
+      "matchPackageNames": ["!mcr.microsoft.com/playwright"],
       "groupName": "GitHub Actions (non-major)",
       "additionalBranchPrefix": "github-actions-",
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -160,10 +160,30 @@
...
+      "matchUpdateTypes": ["patch", "minor"],
+      "groupName": "Playwright (non-major)",
+      "additionalBranchPrefix": "playwright-",
+      "automerge": false
+    },
+    {
+      "description": "Align Playwright Docker release age with npm minimumReleaseAge",
+      "matchPackageNames": ["mcr.microsoft.com/playwright"],
+      "matchDatasources": ["docker"],
+      "minimumReleaseAge": "3 day"
```
gustavolira, line 187: This blocks the image updates instead of gating them. `minimumReleaseAge` needs a `releaseTimestamp`, and the docker datasource only produces one for Docker Hub, so MCR tags never get one. `minimumReleaseAgeBehaviour` defaults to `timestamp-required`, which leaves every tag pending, and pending upgrades get dropped from a mixed branch.

So the `playwright-` PR would carry `e2e-tests/package.json` alone and leave `.ci/images/Dockerfile` and both `container.image` lines on the old tag for good. Same break as #5275, minus the PR that would tell us about it.

Dropping the npm gate for Playwright gets both sides moving together instead. MCR published `v1.63.0-noble` 58 min after npm published `1.63.0`.

```json
{ "matchPackageNames": ["@playwright/test", "playwright", "playwright-core"], "minimumReleaseAge": "0 days" }
```

kim-tsao: I updated the config so it ages based on the gh release date so we can still get the 3 day "aged-in" minimum to guard against supply chain attacks

gustavolira (approved):

PR #5009, opened 2026-06-24
chore(docs): add frontend migration guides

gashcrumb:
## Description

This change adds a couple migration guides to help plugin developers and RHDH operators better understand the changes required to adopt the new frontend system once it's made the default in RHDH.

Assisted-By: Cursor Desktop

## Which issue(s) does this PR fix

- Fixes [RHIDP-15092](https://redhat.atlassian.net/browse/RHIDP-15092)

## PR acceptance criteria

Please make sure that the following steps are complete:

- [ ] GitHub Actions are completed and successful
- [ ] Unit Tests are updated and passing
- [ ] E2E Tests are updated and passing
- [ ] Documentation is updated if necessary (requirement for new features)
- [ ] Add a screenshot if the change is UX/UI related

PR #5383, opened 2026-09-11
fix(docs): repair the dead links across docs and README

gustavolira:
Repairs every dead link in `docs/` and the root markdown. Most have been broken for months; `markdown-link-check` only started reporting them when #5360 bumped the action from v1.1.2 to v1.1.3 on Sep 9 and the newer version began checking in-file anchors.

The check is required, so it has blocked every PR since — but the value here is the links, not the unblock.

## Anchors that never worked

Three links point at:

```markdown
### Adding application drawers (`application/internal/drawer-*`)
```

GitHub strips the backticks, parens, slash and asterisk but keeps the hyphen before the asterisk, so the slug ends in one: `#adding-application-drawers-applicationinternaldrawer-`. The links omit it, so they have gone nowhere since the guides landed in July (#5009). Verified with `github-slugger`; it is the only heading in `docs/` with this shape, out of 35 internal anchors.

## Pages that moved upstream

| Was | Now | Why |
| --- | --- | --- |
| `backstage.io/docs/reference/filter-predicates` | `backstage.io/api/stable/modules/_backstage_filter-predicates.html` | moved to the generated API reference — thanks @hopehadfield for finding it |
| `backstage.io/docs/reference/core-plugin-api.apptheme/` | `.../modules/_backstage_core-plugin-api.html` | same move; the hand-written page is gone |
| `docs.ci.openshift.org/docs/architecture/prow/` | `/docs/architecture/` | subpage removed |
| community-plugins `redhat-argocd` | `argocd` | renamed in their #6820 |
| community-plugins `github-actions`, `github-issues`, `github-pull-requests-board` | all under `github` | merged in their #7507 |
| `gajus/global-agent` test files | the repo root | neither path exists on `master` or `main` |

## Things that no longer exist here

- `values_showcase-auth-providers.yaml` was deleted in #4458. The bullet above it already says that job deploys with the **Operator** rather than Helm, so a Helm values file does not apply either way — the whole bullet goes.
- `github-discovery.spec.ts` is gone. The `auth-providers` suites are what use fixtures today, so the example points there.

## Rate-limited hosts, not dead links

`docs.redhat.com`, `developers.redhat.com` and `devservices.dpp.openshift.com` answer `200` (or resolve) from a workstation and `403`/`0` from CI. The config's retry settings only cover `429`. They join `pkgs.devel.redhat.com` and `gitlab.cee.redhat.com`, already in `ignorePatterns` for the same reason.

Worth being explicit about the cost: this stops checking every link behind those three hosts, not just the ones that failed. An `aliveStatusCodes` entry for `403` would be narrower in intent but global in effect, weakening every link in the repo instead of three domains.

## Verification

Ran `markdown-link-check` with the repo's own config over both sets the workflow checks — `docs/` and the root markdown. Both clean.

Note the root step had never run: the workflow has no `if: always()`, so `docs/` failing meant the README's six dead links stayed hidden until the first set was fixed.

### DIFF
--- a/docs/dynamic-plugins/migrating-plugins-to-new-frontend-system.md
+++ b/docs/dynamic-plugins/migrating-plugins-to-new-frontend-system.md
@@ -440,7 +440,7 @@ app:
 
 - `EntityCardBlueprint` replaces `createEntityCardExtension` / mount-point card wiring.
 - Layout grid positioning from RHDH `config.layout` is not a standard blueprint config — implement layout inside your card component or use extension overrides for advanced cases.
-- Filter predicates use the filter predicate schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
+- Filter predicates use the [filter predicate](https://backstage.io/api/stable/modules/_backstage_filter-predicates.html) schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
 
 ---
 
--- a/docs/dynamic-plugins/migrating-plugins-to-new-frontend-system.md
+++ b/docs/dynamic-plugins/migrating-plugins-to-new-frontend-system.md
@@ -440,7 +440,7 @@ app:
 
 - `EntityCardBlueprint` replaces `createEntityCardExtension` / mount-point card wiring.
 - Layout grid positioning from RHDH `config.layout` is not a standard blueprint config — implement layout inside your card component or use extension overrides for advanced cases.
-- Filter predicates use the [filter predicate](https://backstage.io/docs/reference/filter-predicates) schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
+- Filter predicates use the filter predicate schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
 
 ---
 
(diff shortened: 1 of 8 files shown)


docs/dynamic-plugins/migrating-plugins-to-new-frontend-system.md
```diff
@@ -440,7 +440,7 @@ app:
 
 - `EntityCardBlueprint` replaces `createEntityCardExtension` / mount-point card wiring.
 - Layout grid positioning from RHDH `config.layout` is not a standard blueprint config — implement layout inside your card component or use extension overrides for advanced cases.
-- Filter predicates use the [filter predicate](https://backstage.io/docs/reference/filter-predicates) schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
+- Filter predicates use the filter predicate schema in config (not the RHDH `isKind`/`isType` shorthand, though similar concepts apply).
```
hopehadfield: I think this may be the correct link https://backstage.io/api/stable/modules/_backstage_filter-predicates.html instead

hopehadfield (approved):
/lgtm

PR #5359, opened 2026-09-09
fix(deps): bump better-sqlite3 to 13.0.3 for Node 24

NiallTwomey2:
Node 24 aborts in better-sqlite3 12.x during Statement teardown (`RemoveEnvironmentCleanupHook`, `env != nullptr`). 13.x uses N-API and avoids that crash. This unblocks rhdh-local `dev` CI, which runs `quay.io/rhdh-community/rhdh:next` with in-memory SQLite.

## Description

Bump `better-sqlite3` from `12.11.1` to `13.0.3` in `packages/backend`.

rhdh-local `dev` does not contain this dependency. It pulls `:next`, which is built from rhdh `main`. After this merges and `:next` rebuilds, rhdh-local default compose jobs should stop aborting.

## Which issue(s) does this PR fix

- https://redhat.atlassian.net/browse/RHDHBUGS-3725
- Symptom (rhdh-local `dev` CI): https://github.com/redhat-developer/rhdh-local/pull/307
- Example job: https://github.com/redhat-developer/rhdh-local/actions/runs/33883263689/job/101057223181

## PR acceptance criteria

- [ ] rhdh CI checks are passed
- [ ] Local `yarn start` on Node 24 stays up (`Listening on :7007`, no `sqlite native abort`)
- [ ] After merge and `:next` rebuild, rhdh-local default / corporate-proxy / dynamic-plugins-root jobs return HTTP 200

PR #5403, opened 2026-09-15
fix(ci): use better-sqlite3 v13 prebuilds under Yarn 4

jonkoops:
After #5359, cluster-free E2E fails on `yarn install` because Yarn 4 runs `node-gyp` whenever a dependency ships `binding.gyp`, even though better-sqlite3 13 bundles platform prebuilds and sets `gypfile: false` for npm ([WiseLibs/better-sqlite3#1503](https://github.com/WiseLibs/better-sqlite3/issues/1503)).

This adds `dependenciesMeta.better-sqlite3.built: false` (same pattern as `keytar`) so Yarn skips the native build and loads the shipped prebuild at runtime.


jonkoops:
Closing per author request.

PR #5226, opened 2026-08-06
chore(deps): update yarn to v4.17.1

nickboldt:
## Summary
- Bump Yarn from 4.12.0 to 4.17.1 (packageManager, yarnPath / release binary, Containerfile pin) to align with rhdh-plugins.

## Test plan
- [ ] Confirm `yarn -v` reports 4.17.1 via yarnPath / Corepack
- [ ] Spot-check install/build for affected packages

Generated-by: cursor

Ref: https://redhat.atlassian.net/browse/RHIDP-16074

PR #5405, opened 2026-09-15
fix: disable yarn install scripts now that better-sqlite3 uses prebuilds

jonkoops:
better-sqlite3 v13 ships prebuilt binaries, but our Yarn config still ran install scripts on every dependency. That triggered `node-gyp` compiles during `yarn install`, which failed in cluster-free E2E on Playwright images that do not have `make` or gcc, even though no compile was needed.

[#5226](https://github.com/redhat-developer/rhdh/pull/5226) added `enableScripts: true` when upgrading to Yarn 4.17, because better-sqlite3 v12 still needed a local compile. [#5359](https://github.com/redhat-developer/rhdh/pull/5359) bumped better-sqlite3 to v13 with prebuilds, which left that setting forcing unnecessary compiles.

This PR turns install scripts off again (Yarn 4’s default), removes the root `dependenciesMeta` and the `node-gyp` devDependency from [#546](https://github.com/redhat-developer/rhdh/pull/546), drops the e2e-tests `postinstall` Playwright hook (removing the need for `--mode=skip-build` on CI), and updates docs to describe the manual `yarn playwright install chromium` step and drop the node-gyp toolchain prerequisites. Stale comments in CI and SonarCloud config are trimmed to match.


docs/e2e-tests/README.md
```diff
@@ -24,13 +24,6 @@ From the root of the project directory, navigate to the `e2e-tests` directory:
 ```bash
 cd e2e-tests
 yarn install
-```
-
-## Install Playwright Browsers
-
-The Playwright browsers should be installed automatically via the `postinstall` script in `package.json`. If not, you can manually install them:
-
-```bash
```
polasudo, line 33: just a nitpick here, this can be left there

polasudo (approved):

docs/e2e-tests/README.md
jonkoops: I'd rather not, as it complicates the CI, where we then have to ignore these scripts with a flag, or it will kick off a build of native deps and install browsers (which are already installed in the image it uses).

Using postinstall scripts is bad practice in general, so people can just run this one step manually. If they don't, then Playwright will prompt them to install it anyway, so I am willing to sacrifice a little DX for sanity here.

polasudo (approved):

PR #5412, opened 2026-09-15
fix: drop node-gyp workarounds now that better-sqlite3 uses prebuilds

jonkoops:
Backport of https://github.com/redhat-developer/rhdh/pull/5405 to `release-1.10`.

#5365 already bumped better-sqlite3 to v13 with prebuilds on this branch. This drops the leftover root `node-gyp` devDependency and `dependenciesMeta`, removes the e2e-tests `postinstall` Playwright hook (and `--mode=skip-build` in the e2e lint workflow), and updates docs to match.

`release-1.10` does not have `enableScripts: true` in `.yarnrc.yml` or the cluster-free harness from `main`, so those parts of #5405 do not apply here.


jonkoops:
Closing: release-1.10 does not have enableScripts or cluster-free E2E, so #5405 does not fix a broken scenario here. The partial backport also regressed hermetic builds by removing keytar dependenciesMeta without an equivalent on main.

PR #5413, opened 2026-09-15
fix(e2e): stabilize bulk import Keycloak login and catalog verification (release-1.9)

gustavolira:
## Description

Fixes two recurring bulk-import E2E failures in the release-1.9 nightlies:

**[RHDHBUGS-3754](https://redhat.atlassian.net/browse/RHDHBUGS-3754) / [RHDHBUGS-3773](https://redhat.atlassian.net/browse/RHDHBUGS-3773) — "Verify Added Repositories Appear in the Catalog as Expected"**
The test waited a single 15s locator timeout for the imported component to show up in the catalog, but the catalog processing loop can take longer than that to materialize the entity after the import is registered. The verification now polls with page reloads (`toPass`, up to 2 minutes), the same pattern already used by the other steps in this spec.

**[RHDHBUGS-3756](https://redhat.atlassian.net/browse/RHDHBUGS-3756) — "unexpected sign-in method page" login failure**
`loginAsKeycloakUser` registered its `popup` listener only *after* clicking "Sign In", so when the popup opened early the listener missed it and the login never proceeded past the "Select a sign-in method" page. Changes:
- attach the popup listener before clicking (`Promise.all` with `waitForEvent("popup")`)
- wait for the OIDC provider card to render before clicking Sign In
- treat a popup that closes immediately (active Keycloak SSO session) as already logged in
- stop depending on the exact `"Target closed"` error-message substring, which newer Playwright versions no longer produce, and check `popup.isClosed()` instead

- [RHDHBUGS-3754](https://redhat.atlassian.net/browse/RHDHBUGS-3754)
- [RHDHBUGS-3773](https://redhat.atlassian.net/browse/RHDHBUGS-3773)
- [RHDHBUGS-3756](https://redhat.atlassian.net/browse/RHDHBUGS-3756)

## PR acceptance criteria

- [x] Existing E2E flow preserved; changes are test-stability only
- [x] ESLint: 0 errors; Prettier clean


.ci/pipelines/utils.sh
```diff
@@ -1743,34 +1795,32 @@ EOF
...
-      if [[ $elapsed -ge $timeout_secs ]]; then
-        log::warn "Timeout waiting for operator reconciliation of '$workflow' after ${timeout_secs}s"
-        break
-      fi
-      local ready
-      ready=$(oc get deployment "$workflow" -n "$namespace" -o jsonpath='{.status.conditions[?(@.type=="Progressing")].status}' 2> /dev/null || echo "")
-      if [[ "$ready" == "True" ]]; then
-        log::info "SonataFlow operator reconciled '$workflow' deployment"
-        break
+    while ! oc get deployment "$workflow" -n "$namespace" &> /dev/null; do
```
rostalan: this and lines 1691-1695 is duplicated, maybe extract and call from here?

e2e-tests/playwright/utils/common.ts
```diff
@@ -96,24 +96,66 @@ export class Common {
...
+        .waitForEvent("close", { timeout: 3_000 })
+        .then(
+          () => true,
+          () => false,
+        );
+      if (closedLate) {
+        return;
+      }
+      throw error;
+    }
```
rostalan:
```suggestion
      await popup
        .waitForEvent("close", { timeout: 3_000 })
        .catch(() => {
          throw error;
        });

      return;
    }
```

PR #5435, opened 2026-09-17
fix(ci): override stale RELATED_IMAGE_backstage digest in the operator CSV

gustavolira:
## Description

All operator-based nightlies (`e2e-osd-gcp-operator-nightly`, `e2e-gke-operator-nightly`, `e2e-aks-operator-nightly`) have been red for weeks with **zero tests executed**: the `install-dynamic-plugins` init container crash-loops with

```
InstallException: Cannot resolve ref:// reference: no plugin named
'backstage-community-plugin-catalog-backend-module-keycloak' found in included plugins
```

### Root cause

- The operator bundle CSV pins `RELATED_IMAGE_backstage` to the hub digest captured at bundle build time. The current `next` bundle (built 2026-08-31) still pins `rhdh-hub-rhel9@sha256:a89235...` — the **2026-08-04** build, which bundles `install-dynamic-plugins` **0.4.0**. That version resolves `ref://` references only *after* the pre-merge disabled pass, so refs pointing at plugins that the catalog index ships disabled (keycloak, scaffolder-relation-processor, rbac…) can never resolve. Fixed in 0.4.1 (rhdh-plugins#4410, 2026-08-20), which today's `:next` tag already carries — the same config installs cleanly against `quay.io/rhdh/rhdh-hub-rhel9:next` (verified locally with the exact CI ConfigMap).
- The Backstage CR `deployment.patch` already overrides both containers to `$IMAGE_REGISTRY/$IMAGE_REPO:$TAG_NAME`, but the installed operator applies the `RELATED_IMAGE_backstage` env to the init container **after** the patch, so the stale digest wins there (pod events confirm the init container pulls the digest while `backstage-backend` gets `:next`).

### Fix

After installing the operator, patch the CSV's `RELATED_IMAGE_backstage` to the image under test and wait for OLM to roll the operator deployment. This keeps every operator job testing the same image the helm jobs test, on all branches (`next` and `-v X.Y` installs alike).

The jq patch and the poll condition were validated against the real `rhdh-operator-bundle:next` CSV.

### Follow-ups (out of scope here)

- The `next` operator bundle pinning a 4-week-old hub digest is a productization/midstream issue worth its own report.
- The operator ignoring `deployment.patch` for init containers while honoring it for containers looks like an rhdh-operator bug.


gustavolira:
The osd-gcp validation run ([build 2100932305752166400](https://prow.ci.openshift.org/view/gs/test-platform-results-public/pr-logs/pull/redhat-developer_rhdh/5435/pull-ci-redhat-developer-rhdh-main-e2e-osd-gcp-operator-nightly/2100932305752166400)) revealed a gap: on OCP/OSD ≥ 4.18 the installer auto-detects OLM v1 and installs via ClusterExtension, so there is no CSV and the override takes its skip guard — the stale digest stayed in effect and the `ref://` crash-loop persisted (`RHDH installed via OLM v1 ClusterExtension; skipping CSV backstage image override` in the step log).

Pushed a follow-up commit that passes `--olm-version v0` to `install-rhdh-catalog-source.sh`, keeping the standard operator jobs on the CSV-based path the override patches. OLM v1 coverage keeps its dedicated flows (disconnected jobs / [RHIDP-14647](https://redhat.atlassian.net/browse/RHIDP-14647)). The osd-gcp job needs a re-run to validate end to end.

gustavolira:
End-to-end validation on `e2e-osd-gcp-operator-nightly` ([build 2101025969811755008](https://prow.ci.openshift.org/view/gs/test-platform-results-public/pr-logs/pull/redhat-developer_rhdh/5435/pull-ci-redhat-developer-rhdh-main-e2e-osd-gcp-operator-nightly/2101025969811755008)) — the full chain worked:

- `Image registry default-route is ready` → operator installed via OLM v0 on attempt 1
- `RHDH CSV present` → `Overriding RELATED_IMAGE_backstage in ...rhdh-operator.v2.0.0 with 'quay.io/rhdh-community/rhdh:next'` → `Operator deployment picked up backstage image override` → rolled out
- The `Cannot resolve ref://` InstallException is **gone**; `install-dynamic-plugins` completed (Exit 0) against `rhdh-community/rhdh:next` (the stale Aug-4 digest appears nowhere)
- **showcase-operator: 0 failures · showcase-operator-rbac: 0 failures**

The job's red comes solely from `showcase-runtime`'s deploy setup (backend never passed readiness, pre-existing on this env and unrelated to this PR — no pod logs were collected on that path, worth a follow-up). Ready for review.

gustavolira:
Post-rebase run ([build 2101050308804218880](https://prow.ci.openshift.org/view/gs/test-platform-results-public/pr-logs/pull/redhat-developer_rhdh/5435/pull-ci-redhat-developer-rhdh-main-e2e-osd-gcp-operator-nightly/2101050308804218880)) revalidated the full chain — registry probe (`Image registry is serving through the default route`), CSV override applied, no `ref://` errors, **showcase-operator 22/22 and showcase-operator-rbac 16/16 green**. The only red is the pre-existing `showcase-runtime` setup stall (backend Running but never Ready; root cause unobservable because the timeout diagnostics crash before printing pod events and never collect backend logs — fixed separately in #5462). Retesting.

/test e2e-osd-gcp-operator-nightly

hopehadfield (approved):
/lgtm

zdrapela:
Hi, since this is a temporary override, can you please create a tracker JIra issue so we can revert this when Operator images are rebuilt?

gustavolira:
@zdrapela Tracker created: [RHIDP-17170](https://redhat.atlassian.net/browse/RHIDP-17170).

One nuance captured there: the part that's genuinely temporary is the forced `--olm-version v0` (until the OLM v1 lanes have an equivalent image-override mechanism, [RHIDP-14647](https://redhat.atlassian.net/browse/RHIDP-14647)). The CSV `RELATED_IMAGE_backstage` patch itself is arguably permanent-by-design: even with freshly rebuilt bundles, the CSV pins the digest captured at bundle-build time, so without the override the operator lanes would test that snapshot instead of the same floating tag the Helm lanes test. The tracker frames both decisions for when automatic bundle rebuilds land.

zdrapela:
Thanks. I see a future risk in using OLM v0, because for OpenShift 5.0 it shouldn't be supported, I believe.
There are also advancements in Operator and how it handles the `RELATED_IMAGE_*`. I have some knowledge because it was discussed at Install meetings. 

Please read the AI-generated comment as an outcome of an AI research:

After checking the current operator implementation, I agree that forcing OLM v0 is temporary, but I would not treat the CSV patch as permanent. It is only needed while the legacy operator uses RELATED_IMAGE_backstage for both the backend and plugin installer.

The replacement architecture is already implemented:

- rhdh-operator#3289 (https://github.com/redhat-developer/rhdh-operator/pull/3289), tracked by [RHIDP-15025](https://redhat.atlassian.net/browse/RHIDP-15025) (https://redhat.atlassian.net/browse/RHIDP-15025) and [RHIDP-15028](https://redhat.atlassian.net/browse/RHIDP-15028) (https://redhat.atlassian.net/browse/RHIDP-15028), introduces the dedicated installer selected through OPERATOR_DP_PROCESSING and RELATED_IMAGE_plugin_installer.
- rhdh-operator#3422 (https://github.com/redhat-developer/rhdh-operator/pull/3422) adds DevHubPluginCatalog, which the operator needs to resolve our existing ref:// configuration.
- rhdh-operator#3466 (https://github.com/redhat-developer/rhdh-operator/pull/3466), tracked by [RHIDP-16681](https://redhat.atlassian.net/browse/RHIDP-16681) (https://redhat.atlassian.net/browse/RHIDP-16681), replaces the installer with the current Go implementation.
- rhdh-operator#3436 (https://github.com/redhat-developer/rhdh-operator/pull/3436) / [RHIDP-16024](https://redhat.atlassian.net/browse/RHIDP-16024) (https://redhat.atlassian.net/browse/RHIDP-16024) publishes quay.io/rhdh-community/rhdh-plugin-installer:next.
- Productization is still pending in [RHIDP-16025](https://redhat.atlassian.net/browse/RHIDP-16025) (https://redhat.atlassian.net/browse/RHIDP-16025).

In that model, the CI contract should be:

- backstage-backend uses $IMAGE_REGISTRY/$IMAGE_REPO:$TAG_NAME.
- install-dynamic-plugins uses the bundle’s dedicated RELATED_IMAGE_plugin_installer.
- CI configures DevHubPluginCatalog from $CATALOG_INDEX_IMAGE and waits for it to become ready.

Therefore, once the new operator artifacts are available, the CSV RELATED_IMAGE_backstage patch becomes unnecessary for the installer and should be removed or retained only as a legacy fallback. We must also remove the current CR patches that force the init container back to the hub image.

The inspected downstream quay.io/rhdh/iib:next-v4.20-x86_64 still contains the August 4 operator and lacks the new environment variables, related image, and catalog support. Adoption therefore requires rebuilding the operator image, bundle, and IIB, not only refreshing the bundle.

I suggest expanding [RHIDP-17170](https://redhat.atlassian.net/browse/RHIDP-17170) (https://redhat.atlassian.net/browse/RHIDP-17170) to track:

1. Publishing the updated operator image, bundle, and IIB.
2. Migrating CI to the dedicated installer and DevHubPluginCatalog.
3. Removing the init-container hub-image overrides.
4. Removing or legacy-gating the CSV patch.
5. Removing forced OLM v0 and validating OLM v1 under [RHIDP-14647](https://redhat.atlassian.net/browse/RHIDP-14647) (https://redhat.atlassian.net/browse/RHIDP-14647).
6. Updating disconnected handling through [RHIDP-15003](https://redhat.atlassian.net/browse/RHIDP-15003) (https://redhat.atlassian.net/browse/RHIDP-15003) / rhdh-operator#3601 (https://github.com/redhat-developer/rhdh-operator/pull/3601).

PR #5436, opened 2026-09-17
fix(e2e): surface uncaught page errors when the app bootstrap hangs

gustavolira:
Jira: [RHIDP-17093](https://redhat.atlassian.net/browse/RHIDP-17093)

## Description

Nightly OCP helm runs (`e2e-ocp-helm-nightly`, `e2e-ocp-v4-22-helm-nightly`) intermittently fail across many unrelated test titles with the same signature: the page stuck on the initial `MuiCircularProgress` spinner until the 120s `toBeHidden` timeout in `waitForLoadingToSettle` (`app-shell.ts`), through 3 retries each (~6 min per affected test).

Playwright traces show the real cause is an **uncaught bootstrap error** — currently:

```
ReferenceError: Cannot access 'y' before initialization
    at Module.GlobalHeaderComponentBlueprint (…global-header/…exposed-PluginRoot.….chunk.js)
```

after which React never mounts, so the spinner never clears. The timeout hides that error entirely: the junit shows N different test titles, none pointing at the module-federation failure.

### Fix

- `watchPageErrors(page)`: collects `pageerror` events, attached **before** `page.goto` in `signInAsGuest` and the auth-provider landing flow so bootstrap-time errors are captured; errors reset on main-frame navigation so stale ones can't leak into later waits.
- `waitForLoadingToSettle` now fails fast: if an uncaught page error was seen and the spinner is still visible after a 10s grace period, throw immediately with the actual JS error(s) instead of burning the full timeout and retries. Behavior is unchanged when no page error occurs (same `toBeHidden` assertion at the deadline) and when the app recovers within the grace period.

This does not fix the global-header/module-federation bug itself (tracked separately — the fix belongs in the dynamic-plugin share-scope handling/exporter), but it turns a 6-minute misleading timeout into an immediate failure that names the real error.


e2e-tests/playwright/support/auth/app-shell.ts
```diff
@@ -8,12 +9,108 @@ const LOADING_INDICATOR_SELECTORS = [
...
+ */
+export function watchPageErrors(page: Page): Error[] {
+  const existing = capturedPageErrors.get(page);
+  if (existing) {
+    return existing;
+  }
+  const errors: Error[] = [];
+  capturedPageErrors.set(page, errors);
+  page.on("pageerror", (error) => errors.push(error));
+  page.on("domcontentloaded", () => {
```
hopehadfield: Resetting errors on `domcontentloaded` can erase the bootstrap errors this PR is intended to expose. Because the watcher is registered before `page.goto()`, an exception raised during document loading may be captured and then cleared before `waitForLoadingToSettle()` starts, causing the original misleading timeout to remain. Client-side navigations also do not emit `domcontentloaded`, so errors can become stale across later waits. Please reset the state on main-frame navigation instead, and add a test covering an error emitted before `domcontentloaded`.

hopehadfield (commented):
Added one comment, but otherwise should be good to go.

e2e-tests/playwright/support/auth/app-shell.ts
gustavolira: Good catch. Went with an explicit `resetPageErrors(page)` right before the `page.goto` calls instead of any event listener — `framenavigated` also fires for client-side route changes, which would wipe a just-captured bootstrap crash mid-wait. Added the requested test: an error emitted before `domcontentloaded` is now retained (plus one covering client-side navigations).

hopehadfield (approved):
/lgtm

PR #5441, opened 2026-09-18
fix(ci): mount real conditional policies in operator RBAC deployments [release-1.9]

gustavolira:
## Description

The `release-1.9` `showcase-rbac` project has been failing on operator-based nightlies (gke-operator et al.) with RBAC tests seeing 0 permissions and an empty catalog ([RHDHBUGS-3773](https://redhat.atlassian.net/browse/RHDHBUGS-3773), [RHDHBUGS-3538](https://redhat.atlassian.net/browse/RHDHBUGS-3538)).

### Root cause

`create_conditional_policies_operator` extracts the conditional policies from `.upstream.backstage.initContainers[0].command[2]` in `values_showcase-rbac.yaml` — but that section was removed in February by #4162, which moved the policies to the static `resources/config_map/conditional-policies.yaml`. Since then the `yq` path resolves to `null` and every **operator** RBAC deployment mounts an **empty** `conditional-policies.yaml` (Helm deployments mount the real file, which is why helm RBAC passes).

Without conditional policies, the test users lose their conditional `catalog.entity` read/update/delete grants: role pages show 0 permissions and the catalog renders empty — exactly the `$currentUser` / `$ownerRefs` / `policyDecisionPrecedence` failures in the nightly (verified against the 2026-09-18 gke-operator run artifacts).

### Fix

Backport main's behavior: always create the `rbac-policy` ConfigMap from the static `conditional-policies.yaml`, and remove the dead helper plus its five call sites.

Also fixes the default RBAC GitHub App env block in `env_variables.sh`: `GITHUB_APP_PRIVATE_KEY_RBAC` read the **client secret** (JWT minting can never succeed) and `GITHUB_APP_CLIENT_SECRET_RBAC` was missing. These defaults kick in whenever the per-run prefix rotation finds an incomplete secret set, then GitHub discovery/bulk-import fail with 401s.

### Verification

`bash -n` on all touched scripts, `yarn shellcheck` and `yarn prettier:check` green; no references to the removed helper remain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5442, opened 2026-09-18
fix(ci): correct default RBAC GitHub App env vars

gustavolira:
## Description

In `.ci/pipelines/env_variables.sh`, the default RBAC GitHub App block (used whenever the per-run prefix rotation finds an incomplete `_RBAC_<n>` secret set — see `override_github_app_env_with_prefix`) has two bugs:

- `GITHUB_APP_PRIVATE_KEY_RBAC` reads `GITHUB_APP_CLIENT_SECRET_OPERATOR` — an app JWT can never be minted from a client secret, so every GitHub App call from the RBAC namespace fails with `401 Unauthorized` (breaking catalog discovery and bulk-import) whenever the defaults kick in.
- `GITHUB_APP_CLIENT_SECRET_RBAC` is missing from the block entirely.

Same fix as the release-1.9 counterpart in #5441 (where the 2026-09-18 gke-operator nightly showed the resulting `Fetching repositories with token from token failed 401 Unauthorized`).


hopehadfield (approved):
/lgtm

PR #5447, opened 2026-09-18
fix(ci): merge the AWS RDS global CA bundle into the RDS trust store

gustavolira:
## Description

The RDS TLS e2e tests in `showcase-runtime` fail with `self-signed certificate in certificate chain` on all four RDS instances (latest .. latest-3) — seen on `e2e-osd-gcp-helm-nightly`, `e2e-ocp-helm-nightly` and the operator variants ([RHDHBUGS-3411](https://redhat.atlassian.net/browse/RHDHBUGS-3411), [RHDHBUGS-3691](https://redhat.atlassian.net/browse/RHDHBUGS-3691)).

### Root cause

`RDS_DB_CERTIFICATES_PATH` points at the Vault copy of the RDS CA bundle. AWS rotates RDS CAs, and GSM cannot store the full ~170KB global bundle ([RHDHBUGS-3744](https://redhat.atlassian.net/browse/RHDHBUGS-3744)), so the stored copy is a trimmed snapshot that goes stale — the harness pg client (`clearDatabase`) and RHDH itself (via the `postgres-crt` secret) then reject the servers' chains.

### Fix

At env-setup time, download the official [AWS RDS global trust bundle](https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem) and merge the Vault copy into it, pointing `RDS_DB_CERTIFICATES_PATH` at the merged file. Every current and future RDS CA generation/region validates, custom CAs from Vault are preserved, and if the download fails the behavior falls back to today's (Vault copy only, with a warning).


hopehadfield (approved):
/lgtm

zdrapela:
Hi, since we have a size constraint for storing it in GSM (https://redhat.atlassian.net/browse/RHDHBUGS-3744), we can switch to using the downloaded `global-bundle.pem` without a fallback.
The fallback would require having the certificates stored into multiple secrets and will be overly complicated.
So I mean that we can remove the fallback and close https://redhat.atlassian.net/browse/RHDHBUGS-3744

gustavolira:
@zdrapela Agreed — follow-ups opened dropping the Vault copy entirely: #5486 (main) and #5487 (release-1.10). The downloaded official bundle becomes the only RDS trust store; on a failed download the file is simply absent and the RDS suite reports/skips on the missing certificate. Once they merge, the `rds-db-certificates.pem` secret can be deleted and [RHDHBUGS-3744](https://redhat.atlassian.net/browse/RHDHBUGS-3744) closed.

PR #5462, opened 2026-09-21
fix(e2e): make setup-timeout diagnostics actually diagnose

gustavolira:
## Description

Two consecutive `e2e-osd-gcp-operator-nightly` runs (on #5435) failed `showcase-runtime`'s operator deploy with the backend Running but never Ready — and the root cause is unknowable from the artifacts, because the diagnostics pipeline destroys its own evidence. Three fixes:

1. **Pod-event listing crashed before printing anything**: `getEventSortTimestamp`/`formatEventTimestamp` guard `!== undefined`, but the API returns `firstTimestamp: null`; `typeof null === "object"` routes into `.getTime()` → `Cannot read properties of null (reading 'getTime')` (seen verbatim in both runs). The pod events are exactly where the failing readiness-probe lines appear. Now guards `null` explicitly.
2. **Backend container logs were never collected on setup timeout**: `collectTimeoutDiagnostics` calls events/conditions but not the already-existing `logPodContainerLogs`. Now it does.
3. **Misleading timeout message**: `pollUntil` throws the bare label, so junit shows `Error: Deployment backstage-rhdh ready (1 replicas)` — reads like a success. Now: `Timed out after 600000ms waiting for: Deployment backstage-rhdh ready (1 replicas)`.

With this merged, the next runtime setup failure will carry probe status codes and backend logs in the step output.

### Verification
`yarn tsc` 0, `yarn lint` 0 errors, `yarn fmt:check` clean, `yarn test:unit` 86 passed.


hopehadfield (approved):
/lgtm

gustavolira:
The `e2e-ocp-helm` failures here are the chart-level CI breaker, not this PR: chart `2.0-90-CI`'s bumped Bitnami PostgreSQL subchart rejects our Red Hat postgresql image at template time, killing every Helm deploy on main. Fix: #5490 (`global.security.allowInsecureImages`). Will retest once it merges.

PR #5464, opened 2026-09-21
chore: remove OFS plugin-utils and static dynamic-plugins config

jonkoops:
Follow-up to #5452. Removes `@red-hat-developer-hub/plugin-utils` and `app-config.dynamic-plugins.yaml`, neither of which the NFS app uses.

[RHIDP-17024](https://redhat.atlassian.net/browse/RHIDP-17024)


hopehadfield (approved):
LGTM, just a couple items to note.

We probably want to update the docs to reflect this change as well (I see at least one reference to the static dynamic-plugins config [here](https://github.com/redhat-developer/rhdh/blob/main/docs/e2e-tests/layer-migration-matrix.md?plain=1#L42)). Additionally, the [EOL dependency inventory report](https://github.com/redhat-developer/rhdh/blob/main/scripts/eol-inventory/eol-dependency-inventory-report.md) should be regenerated now that this package is removed.

PR #5478, opened 2026-09-22
test(e2e): use PostgreSQL 18

zdrapela:
## Summary

- use PostgreSQL 18 for integrated Helm E2E deployments
- allow downstream jobs to override the PostgreSQL image
- update the Crunchy external database and OpenShift setup images

## Testing

- `yarn shellcheck && yarn prettier:check` in `.ci`
- `yarn lint && yarn fmt:check && yarn tsc --noEmit && yarn test:unit` in `e2e-tests`
- Bash syntax, YAML parsing, and `git diff --check`

## Jira

- https://redhat.atlassian.net/browse/RHIDP-13956
- https://redhat.atlassian.net/browse/RHIDP-13961


.ci/pipelines/lib/postgres.sh
```diff
@@ -0,0 +1,277 @@
...
+  # A fresh cluster already contains the postgres role and database. Continue
+  # past those expected duplicate-object errors, but reject any other SQL error.
+  oc exec -i -n "${namespace}" "${pod}" -- \
+    env LC_ALL=C psql -U postgres -v ON_ERROR_STOP=0 < "${dump_file}" > "${restore_log}" 2>&1 \
+    || restore_status=$?
+
+  if [[ ${restore_status} -ne 0 ]]; then
+    log::error "PostgreSQL restore command failed with status ${restore_status}"
+    grep -E 'ERROR:|FATAL:' "${restore_log}" || true
+    rm -f "${restore_log}"
```
gustavolira: On a failed restore this deletes the full psql log before anything saves it, so all we keep is the grepped `ERROR:` lines. The statements around them and any connection/`FATAL` output are gone. Could we `common::save_artifact` the log before returning 1 (same at line 242)? `cleanup.sh` already removes `${POSTGRES_UPGRADE_DUMP_FILE}.restore.log`, so the early `rm` isn't needed to clean up.

.ci/pipelines/jobs/upgrade.sh
```diff
@@ -59,8 +137,93 @@ handle_ocp_helm_upgrade() {
...
   local url="https://${RELEASE_NAME}-developer-hub-${NAME_SPACE}.${K8S_CLUSTER_ROUTER_BASE}"
-  initiate_upgrade_base_deployments "${RELEASE_NAME}" "${NAME_SPACE}" "${url}"
-  initiate_upgrade_deployments "${RELEASE_NAME}" "${NAME_SPACE}" "${url}"
+  UPGRADE_PHASE="baseline-deploy"
+  initiate_upgrade_base_deployments \
+    "${RELEASE_NAME}" "${NAME_SPACE}" "${url}" "${upgrade_artifacts_subdir}"
+
+  UPGRADE_PHASE="baseline-readiness"
+  postgres::wait_ready "${NAME_SPACE}" "${RELEASE_NAME}"
+  if ! testing::check_backstage_running "${RELEASE_NAME}" "${NAME_SPACE}" "${url}" "showcase-upgrade-base"; then
```
gustavolira: nit: this hard-codes `showcase-upgrade-base` as the artifacts dir, but the rest of the baseline phase now writes to `${upgrade_artifacts_subdir}/baseline` (the values file in `utils.sh`, `upgrade::save_phase_artifacts "baseline"`). A baseline readiness failure ends up split across two folders. Pass `"${upgrade_artifacts_subdir}/baseline"` here?

.ci/pipelines/lib/common.sh
```diff
@@ -88,7 +88,8 @@ common::highest_release_stream_for_major() {
   # Capture before filtering: piping straight into sed would discard git's
   # status, making an unreachable remote look like "no such branch".
   local refs
-  if ! refs=$(git ls-remote --heads "https://github.com/${REPO_OWNER:-redhat-developer}/${REPO_NAME:-rhdh}" \
+  if ! refs=$(cd "${TMPDIR:-/tmp}" && git ls-remote --heads \
```
gustavolira: `common::highest_release_stream_for_major` has no callers left after this PR. Its only user was `get_previous_release_version`, which now runs its own `git ls-remote` loop below. Should we drop it rather than keep editing it?

.ci/pipelines/jobs/upgrade.sh
```diff
@@ -59,8 +137,93 @@ handle_ocp_helm_upgrade() {
...
+    # Persist the normal replica count in Helm and reconnect RHDH to PostgreSQL.
+    UPGRADE_PHASE="target-deploy"
+    initiate_upgrade_deployments "${RELEASE_NAME}" "${NAME_SPACE}" "${url}"
+  else
+    log::info "PostgreSQL major versions match; skipping database migration"
+    UPGRADE_PHASE="target-deploy"
+    initiate_upgrade_deployments "${RELEASE_NAME}" "${NAME_SPACE}" "${url}"
+  fi
+
+  UPGRADE_PHASE="target-rollout"
```
gustavolira: With the rollout, readiness and test steps inlined here, `testing::check_upgrade_and_test` (`lib/testing.sh:497`) is dead code. `.ci/pipelines/lib/README.md:61` still lists it and doesn't mention the new `lib/postgres.sh`. Worth removing it and updating the README in this PR?

.ci/pipelines/lib/helm.sh
```diff
@@ -77,47 +77,27 @@ helm::merge_values() {
...
-#   Prints the path to the downloaded value file
+#   $1 - value_file_type: Type of value file (default: "showcase")
 helm::get_previous_release_values() {
-  local value_file_type=${1:-"showcase"}
-
-  local current_release_version
-  current_release_version=$(helm::get_chart_stream)
-  if [[ -z "$current_release_version" ]]; then
+  local value_file_type=${1:-showcase}
+  if [[ -z "${previous_release_version:-}" ]]; then
```
gustavolira: This now reads `previous_release_version` from the caller's scope. It works only because `handle_ocp_helm_upgrade` assigns it as a global, and the doc comment only mentions `$1`. Could we pass the version in as `$2` instead? Then the dependency is visible and a second caller can't hit "Previous release version is not set" by surprise.

gustavolira (approved):

PR #5469, opened 2026-09-21
fix(ci): accept rhdh-pr tags in Podman PR Push publisher

nickboldt:
## Summary
- Transitional update to the default-branch `workflow_run` publisher so it accepts both `rhdh` and `rhdh-pr` preview tags
- Use `RHDH_PR_BOT_*` when pushing `rhdh-pr` tags (keep `QUAY_*` for legacy `rhdh` tags)
- Unblocks https://github.com/redhat-developer/rhdh/pull/5437 while that PR moves preview images to `rhdh-pr`

## Test plan
- [ ] Existing PRs that still build `rhdh:pr-*` continue to publish
- [ ] PR #5437 (and other `rhdh-pr` builders) no longer fail Podman PR Push on disallowed tags

https://redhat.atlassian.net/browse/PSIRTSUPT-22719

Generated-by: cursor

Made with [Cursor](https://cursor.com)

PR #5437, opened 2026-09-17
fix(ci): push PR images to rhdh-pr repo and consolidate workflows

nickboldt:
## Summary
- Push PR preview images to `quay.io/rhdh-community/rhdh-pr` using `RHDH_PR_BOT_USER` / `RHDH_PR_BOT_TOKEN` instead of the production `rhdh` repo
- Build and push directly in `pr-build-image.yaml` (`skipArtifactUpload: true`); remove artifact bridge
- Fold PR comments (skip / success / failure) into the build workflow via `GITHUB_TOKEN`
- Delete `pr-podman-push.yaml`
- Update `check-image-and-changes` skip logic and consumer docs/scripts for `rhdh-pr`
- Add fork guard via `QUAY_PR_IMAGE_REPO` (mirrors e2e-runner pattern)

## Test plan
- [ ] Open or update a PR and confirm `PR Build Image (Hermetic)` builds, pushes to `rhdh-pr`, and posts the image link comment
- [ ] Confirm skip path still comments when no relevant changes and image exists
- [ ] Confirm production `rhdh` repo is not written by PR CI

https://redhat.atlassian.net/browse/PSIRTSUPT-22719

Generated-by: cursor

Made with [Cursor](https://cursor.com)

PR #5479, opened 2026-09-22
fix(ci): harden PR image publish with trusted retags

nickboldt:
## Summary
- Harden `pr-podman-push` like [rhdh-operator #3557](https://github.com/redhat-developer/rhdh-operator/pull/3557): trusted retag from PR number + `workflow_run.head_sha`, no fork checkout
- Keep destination `quay.io/rhdh-community/rhdh` (no `rhdh-pr` migration)
- Pre-pull UBI bases with retries; fix skip-build false positives on merge checkouts
- Document fork PR publish via `workflow_run`

Supersedes the Quay-repo move in #5437 / #5469 and openshift/release #85611.

## Test plan
- [ ] Same-repo PR: hermetic build uploads artifact; publisher pushes `rhdh:pr-<N>` + `rhdh:pr-<N>-<sha>` and comments
- [ ] Fork PR into upstream: same publish path (secrets only on publisher)
- [ ] Skip path: `[skip-build]` / docs-only still skip without false positives from main commits
- [ ] Confirm e2e still waits on `rhdh-community/rhdh` (no openshift/release change)

Generated-by: cursor

Ref: https://redhat.atlassian.net/browse/PSIRTSUPT-22719


.github/workflows/pr-podman-push.yaml
rm3l, line 1: A concurrency group might make sense to handle rapid pushes to the same PR. We added this to the operator: https://github.com/redhat-developer/rhdh-operator/blob/main/.github/workflows/pr-container-push.yaml#L22-L25

.github/workflows/pr-podman-push.yaml
```diff
@@ -31,28 +56,29 @@ jobs:
                 || github.event.workflow_run.head_branch
             }}
           FULL_SHA: ${{ github.event.workflow_run.head_sha }}
-          
+
         run: |
-          SHORT_SHA=$(echo "${{ env.FULL_SHA }}" | cut -c1-8)
+          SHORT_SHA=$(echo "${FULL_SHA}" | cut -c1-8)
```
rm3l, line 68:
```suggestion
          set -euo pipefail

          SHORT_SHA=$(echo "${FULL_SHA}" | cut -c1-8)
```

Suggested by Claude to avoid silent/confusing failures

rm3l (commented):
Looks good. A few minor comments suggested by Claude.

.github/workflows/pr-podman-push.yaml
rm3l, line 1: Looking more deeply at https://redhat.atlassian.net/servicedesk/customer/portal/68/PSIRTSUPT-22719, it seems like the issue reported on the rhdh repo was more about a possible ZIP path traversal attack. And they recommended to upgrade both occurrences of `actions/download-artifact` to v7 or later (v8 is available), which reportedly fixes this known path traversal.

nickboldt: Added — same concurrency group shape as rhdh-operator `pr-container-push` (cancel-in-progress on rapid PR updates).

.github/workflows/pr-podman-push.yaml
nickboldt: Applied `set -euo pipefail` at the start of Get PR number.

.github/workflows/pr-podman-push.yaml
nickboldt: Bumped both `download-artifact` uses in `pr-podman-push.yaml` to v8 (`3e5f45b2…`), and the remaining v4 pins in `next-build-image.yaml` / `push-e2e-runner.yaml` as well (ZIP path-traversal fix for PSIRTSUPT-22719).

rm3l (approved):

.github/workflows/pr-podman-push.yaml
```diff
@@ -7,6 +27,16 @@ on:
     types:
       - completed
 
+concurrency:
+  group: >-
+    ${{ github.workflow }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}-${{ github.event.workflow_run.pull_requests[0].number || github.event.workflow_run.id }}
```
gustavolira, line 32: `workflow_run.pull_requests` is empty for fork PRs — same thing the comment at 69-70 says, which is why the step below has to go through `gh pr view` at all. So this `||` lands on `workflow_run.id`, which is unique per run, and `cancel-in-progress` never matches anything on exactly the fork PRs this workflow exists for.

Two quick pushes on a fork PR then run publishers in parallel, and if the older one finishes last it wins the moving `:pr-N` tag while the PR comment claims otherwise. `head_repository.full_name` + `head_branch` already identify the PR head on their own — can we just drop the last term?

.github/workflows/pr-podman-push.yaml
```diff
@@ -31,28 +61,31 @@ jobs:
                 || github.event.workflow_run.head_branch
             }}
           FULL_SHA: ${{ github.event.workflow_run.head_sha }}
-          
+
         run: |
-          SHORT_SHA=$(echo "${{ env.FULL_SHA }}" | cut -c1-8)
+          set -euo pipefail
```
gustavolira, line 66: With `set -e` on, `PR_NUMBER=$(gh pr view ...)` on line 71 exits the step the moment gh fails, so the `else` at 79-87 never runs and the `::error::` plus the step summary it writes are dead code. Concrete case: a PR whose head branch was deleted before the `workflow_run` fires.

It cascades too, since `get-pr-info` has no `if:` guard — "Comment build failure" loses `pr_number`, and "Comment publish failure" still runs with an empty one, so `parseInt('')` gives `issue_number: NaN`. Keeping both the strict mode and the handler is easy enough:

```bash
if ! PR_NUMBER=$(gh pr view --repo "$PR_TARGET_REPO" "$PR_BRANCH" --json number --jq '.number'); then
  PR_NUMBER=""
fi
```

.github/workflows/pr-podman-push.yaml
```diff
@@ -142,49 +172,72 @@ jobs:
...
-          TAGS_LIST=$(cat ./rhdh-podman-artifacts/tags.txt)
+          # Load the image from tar (may carry untrusted destination tags).
+          LOAD_OUT=$(podman load -i ./rhdh-podman-artifacts/image.tar)
+          echo "$LOAD_OUT"
+
+          # Resolve a local reference to retag from. Prefer first tags.txt line
+          # if present; otherwise parse podman load output.
+          SOURCE_REF=""
+          if [ -f "./rhdh-podman-artifacts/tags.txt" ]; then
+            SOURCE_REF=$(head -n1 ./rhdh-podman-artifacts/tags.txt | tr -d '\r')
```
gustavolira, line 198: `tags.txt` comes from the artifact the `pull_request` job uploads, and that job runs `./.github/actions/docker-build` from the PR checkout — so a fork still controls this value, and it now reaches `podman tag` with nothing checking it.

The destination is genuinely safe, I traced it: the pushed tags come only from `REGISTRY_IMAGE` + PR number + `head_sha`, and nothing from the artifact influences them. But that also means the allowlist at 224-231 now only validates strings this same step just built, so the check that used to guard the untrusted value no longer guards anything. Should we validate `SOURCE_REF` the way `tags.txt` used to be, and reword the `# SECURITY:` comment at 212 so it reads as an invariant assertion rather than input validation?

.github/actions/docker-build/action.yaml
```diff
@@ -165,6 +165,42 @@ runs:
...
+    - name: Pre-pull base images with retries
+      shell: bash
+      env:
+        CONTAINERFILE_PATH: ${{ inputs.containerfilePath }}
+      run: |
+        set -euo pipefail
+        MAX_ATTEMPTS=3
+        mapfile -t BASE_IMAGES < <(
+          grep -E '^FROM[[:space:]]+[^[:space:]]+' "$CONTAINERFILE_PATH" \
+            | awk '{print $2}' \
```
gustavolira, line 180: `awk '{print $2}'` grabs the flag when a FROM line carries one. `FROM --platform=linux/amd64 registry.example.com/foo/bar:1 AS x` gives `--platform=linux/amd64`, which passes the `/` filter on the next line and ends up as `podman pull "--platform=linux/amd64"` — three failed attempts, 45s of backoff, then `exit 1` takes the whole build down.

The Containerfile we ship today has no `--platform`, but it's generated by `scripts/update-Dockerfile.sh` and `containerfilePath` is an action input, so it seems worth skipping flag tokens:

```bash
| awk '{for(i=2;i<=NF;i++) if($i !~ /^-/){print $i; exit}}' \
```

gustavolira: Correction to my own snippet above: `awk '{... exit}'` is wrong — `exit` ends the whole program after the first `FROM`, so only one base image gets pulled. It needs `next`:

```bash
| awk '{for(i=2;i<=NF;i++) if($i !~ /^-/){print $i; next}}' \
```

Checked against the current `build/containerfiles/Containerfile`, which still yields exactly the two digest-pinned UBI bases. Fixed along with the other three in #5485.

PR #5471, opened 2026-09-22
chore(translations): update s3297 translations (de, es, fr, it, ja)

lokanandaprabhu:
Merges s3297 TMS translations into `translations/`.

**Repos:** backstage, community-plugins
**Locales:** de, es, fr, it, ja
**Files updated:** 10

The `en` language key in TMS output has been fixed to the actual target locale.

PR #5481, opened 2026-09-22
fix(ci): pin the cluster-free E2E catalog index

gustavolira:
## Summary

The cluster-free E2E job resolved `quay.io/rhdh/plugin-catalog-index:next` at run time, with no pin. That index is built outside this repository and its floating tag is rebuilt several times a day, so one bad rebuild turns every open PR red at once — in a job whose failures have nothing to do with the PRs under test.

This pins the digest for the PR/push `e2e` job and lets Renovate move it.

## Evidence

On 2026-09-22 three tests failed across unrelated PRs: `plugins/frontend/sidebar.spec.ts`, `plugins/user-settings-info-card.spec.ts` and `settings.spec.ts`. The outcome tracks the index digest the run happened to resolve, not the PR content:

| Index digest | Runs | Result |
|---|---|---|
| `sha256:06487c4a…` | main push, Sep 21 18:35 | 2 failures |
| `sha256:70544f6a…` | #5471 at 10:03, #5464 at 10:28 | 3 failures |
| `sha256:8cb76f42…` | every run from 11:25 | 13 passed |
| `sha256:8ad2b49b…` | this PR's first run | 3 failures again |

Diffing the installed plugin sets between a bad and the good index shows `red-hat-developer-hub-backstage-plugin-global-header` moving from `sha256:03269175…` to `sha256:4f1a704e…` — exactly the surface those three tests assert on.

Note the last row. The index went bad, good, then bad again inside one day, so the upstream regression was not a one-off that has since been fixed. The first commit here pinned whatever `:next` resolved to at the time, and this PR's own run showed that digest was one of the bad ones. The pin now sits on `sha256:8cb76f42`, the digest a full run proves green.

That failure is the mechanism working: one PR went red over a bad index instead of the whole queue, and the pin only moves once a run proves the next digest.

## Changes

- `e2e-tests/local-harness/catalog-index.lock` — the pinned reference for this branch. It records `<repo>:<tag>@<digest>` because Renovate needs the tag to know which one to follow; the resolver emits the digest-only form, since skopeo rejects a reference carrying both.
- `resolve-catalog-index-image.sh` — a `--pinned` mode reading that lock. Precedence is otherwise unchanged: an explicit override still wins, and without `--pinned` the behaviour is what it was. It fails when the pinned tag does not match the branch, so a release-branch cut cannot silently keep testing the previous stream's index.
- `e2e-cluster-free.yaml` — the `e2e` job passes `pinned: "true"`, and no longer forwards `catalog_index_image`: that dispatch input is for `plugin-sanity`, is always empty on this job's triggers, and an override would silently beat the pin. The scheduled `plugin-sanity` job deliberately keeps following the floating tag — validating the moving index is what that job is for.
- `renovate.json` — a regex custom manager for the lock, grouped and scheduled daily after the 05:00 `plugin-sanity` run, so the pin is bumped once that job has vetted the index. An index regression then lands as one red Renovate PR, which runs this same job against the new digest, instead of reddening the whole queue.
- `e2e-tests/unit/resolve-catalog-index-image.test.ts` — 13 cases over the resolver, following `installed-plugins.test.ts`, which already covers the sibling parser in this subsystem.
- `e2e-tests-lint.yaml` — runs `yarn test:unit`, which no workflow ran. The 73 existing tests were not a gate either.
- `docs/e2e-tests/local-e2e-harness.md` — which index CI uses, how to reproduce a run locally, and what cutting a release branch has to do.

## Verification

- Every new assertion was watched failing first: dropping the digest-only emission, the multi-pin guard, the branch/tag check, the newline normalisation, or the unknown-mode guard turns its own case red and only its own case.
- `renovate --platform=local --dry-run=lookup` against the real repo config with a deliberately stale pin produces `updateType: digest`, `newDigest` pointing at the live tag, on branch `renovate/cluster-free-e2e-catalog-index`. A control run confirms the harness itself detects updates.
- `shellcheck`, `oxlint`, `oxfmt --check` and `renovate-config-validator --strict` all clean. The new lint gate caught two oxlint findings in the new test file, which is the argument for having it.

## Known limits

Renovate runs on the default branch only (no `baseBranches`), so a release branch's pin stays where the cut left it until someone moves it. Documented, not solved here.

Fixes: https://redhat.atlassian.net/browse/RHDHBUGS-3807


gustavolira:
The `e2e-ocp-helm` failures here are the chart-level CI breaker, not this PR: chart `2.0-90-CI`'s bumped Bitnami PostgreSQL subchart rejects our Red Hat postgresql image at template time, killing every Helm deploy on main. Fix: #5490 (`global.security.allowInsecureImages`). Will retest once it merges.

jonkoops:
Closing this as there is no need to pin.

PR #5489, opened 2026-09-22
feat: migrate CI values files and scripts to new RHDH chart 2.y format

rm3l:
## Description

Migrate all CI values files and shell scripts from the old bitnami-wrapper RHDH Helm chart structure (1.y) to the new standalone chart format (2.y).

**Key changes:**
- `global.dynamic` → `dynamicPlugins`
- `upstream.backstage.*` → root-level keys (`appConfig`, `image`, `extraEnv`, `extraEnvFrom`, etc.)
- `upstream.postgresql` → `postgresql`, `upstream.volumePermissions` → `postgresql.volumePermissions`
- `global.clusterRouterBase` → `openshift.clusterRouterBase`
- `global.host` → `host`
- `global.lightspeed` → `intelligentAssistant`
- `global.catalogIndex.image` → `catalogIndex.image`
- `route` → `openshift.route`
- `extraEnvVarsSecrets` → `extraEnvFrom` with `secretRef`
- Remove system volumes/mounts now auto-managed by the chart
- Remove auto-injected env vars (`BACKEND_SECRET`, `POSTGRESQL_ADMIN_PASSWORD`)
- Ensure `image.digest=` is set when overriding `image.tag` via `--set` flags

- Fixes [RHIDP-14274](https://redhat.atlassian.net/browse/RHIDP-14274)

## PR acceptance criteria

This PR updates CI infrastructure only. To verify:
1. Review YAML values files for correct path renames and removal of chart-managed fields
2. Review shell scripts for updated `--set` flags and `yq` paths
3. Run CI pipelines against a cluster with the new RHDH chart 2.y installed

Fixes https://redhat.atlassian.net/browse/RHIDP-14274


.ci/pipelines/value_files/values_showcase.yaml
```diff
@@ -1,126 +1,124 @@
...
+  successThreshold: 1
+  timeoutSeconds: 4
+
+extraEnvFrom:
+  - secretRef:
+      name: rhdh-secrets
+  - secretRef:
+      name: redis-secret
+
+ingress:
```
zdrapela, line 117: CI Failing `- ingress: Additional property host is not allowed
`

zdrapela (commented):
I basically agree with Qodo on the findings. https://github.com/redhat-developer/rhdh/pull/5489#issuecomment-5790936244

It would be good to trigger all the test suites affected with `/test`.
It may be easier to ensure each test suite is passing the tests, when the PR is split. I think OCP can be one PR, another for adjusting AKS, EKS, GKE, and the next for `e2e-tests/playwright/utils/runtime-config.ts` (runtime tests, mainly for the external db). But leave it up to you.

.ci/pipelines/value_files/values_showcase.yaml
rm3l: > CI Failing - ingress: Additional property host is not allowed 

Fixed in [`922766e` (this PR)](https://github.com/redhat-developer/rhdh/pull/5489/commits/922766e5a7bf6e66826b31855607d585a49d2deb).

> It may be easier to ensure each test suite is passing the tests, when the PR is split. I think OCP can be one PR, another for adjusting AKS, EKS, GKE, and the next for e2e-tests/playwright/utils/runtime-config.ts (runtime tests, mainly for the external db). But leave it up to you.

Well, I wanted to avoid too many review cycles on this one to get CI passing again (or at least not failing because of the new chart). So I'd leave the OCP, AKS, EKS and GKE changes in the same PR if that mases sense.
That said, regarding the upgrade test issue (Qodo finding #4: `initiate_upgrade_base_deployments passing chart 2.y --set flags to a 1.y baseline chart`), this is a valid concern. When upgrading from 1.y to 2.y, the base deployment uses CHART_VERSION_BASE which resolves to the previous 1.x chart, so the new root-level `image.*` and `openshift.clusterRouterBase` flags will be silently ignored by that chart.
This will be handled in a separate follow-up PR, since it needs version-conditional logic to choose the correct `--set` paths based on the major version of `CHART_VERSION_BASE`.

rm3l:
Closing as it looks like both https://github.com/redhat-developer/rhdh/pull/5503 and https://github.com/redhat-developer/rhdh/pull/5504 overlapped with this PR and have been merged.

/close

